PR #96, opened 2022-05-24
Add CODE_OF_CONDUCT and CONTRIBUTING files

buildbreaker:
Add files!


doriable (approved):
Stamping in advance of CLA issues being resolved.

timostamm:
We already have both files in [docs](https://github.com/bufbuild/protobuf-es/tree/main/docs) - seemed like a good place to collect them. 

According to the [GH docs](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#about-contributing-guidelines), it works too, but in case it causes trouble, let's just put them in `.github`, but we shouldn't have duplicates.

buildbreaker:
@timostamm great looking out. let me make those changes!

PR #100, opened 2022-05-26
Remove cycles

paul-sachs:
- Add lint rule
- Fix linting imports


timostamm:
Thank you, Paul. Looks like we need to change the way we generate the well-known types for `@bufbuild/protobuf` a bit more to avoid the cycle.

paul-sachs:
@timostamm yeah I figured this would fail. Just didn't know enough about the generator to know how to modify it.

PR #119, opened 2022-06-15
Remove circular dependencies

paul-sachs:
- Modified generator to not create cycles
- Add lint rule
- Fix linting imports
- Remove lint rule to restrict duplicate imports

### DIFF
--- a/private/protoplugin/runtime.go
+++ b/private/protoplugin/runtime.go
@@ -14,6 +14,8 @@
 
 package protoplugin
 
+import "fmt"
+
 // Runtime provides convenient access to exports from the runtime library.
 type Runtime struct {
 	ProtoN             *Symbol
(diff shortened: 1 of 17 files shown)


private/protoplugin/runtime.go
```diff
@@ -34,7 +36,27 @@ type Runtime struct {
...
+			JsonReadOptions:    symbolPool.new("JsonReadOptions", "./json-format.js").ToTypeOnly(),
+			JsonWriteOptions:   symbolPool.new("JsonWriteOptions", "./json-format.js").ToTypeOnly(),
+			JsonValue:          symbolPool.new("JsonValue", "./json-format.js").ToTypeOnly(),
+			JsonObject:         symbolPool.new("JsonObject", "./json-format.js").ToTypeOnly(),
+			ProtoInt64:         symbolPool.new("protoInt64", "./proto-int64.js"),
+			ScalarType:         symbolPool.new("ScalarType", "./field.js"),
+			MethodKind:         symbolPool.new("MethodKind", "./service-type.js"),
+			MethodIdempotency:  symbolPool.new("MethodIdempotency", "./service-type.js"),
+		}
+	}
```
paul-sachs, line 59: Couldn't think of a better way to handle this without doing some significant rearchitecting. I think it's sufficient, if a little frail.

timostamm (approved):
I can't think of a simple alternative either. The tight coupling to the internal file layout of the runtime library is not ideal, but as long as we keep releasing generator and library in sync, any possible mismatch we introduce will be caught be CI - the tests deliberately run in a separate package to catch things like that.

Thanks for the fix, this is great! We can remove index-runtime.ts and some other bits now, I'll open a follow-up.

PR #120, opened 2022-06-15
Remove index-runtime.ts

timostamm:
After bufbuild/protobuf-es#119, we can revert the introduction of index-runtime.ts, and don't need runtimeImportPathBootstrapWKT anymore.


private/protoplugin/runtime.go
```diff
@@ -36,7 +36,7 @@ type Runtime struct {
 	MethodIdempotency  *Symbol
 }
 
-func newRuntime(symbolPool *symbolPool, runtimeImportPath string, syntax ProtoSyntax, bootstrapWKT bool) *Runtime {
+func newRuntime(symbolPool *symbolPool, syntax ProtoSyntax, bootstrapWKT bool) *Runtime {
```
paul-sachs, line 39: This threw me a little, since the function internally accessing runtimeInputPath... I guess there's a scope it's accessing that I'm not familiar with... I guess it's because it shares the same package as `names.go`?

paul-sachs (approved):

private/protoplugin/runtime.go
timostamm: Exactly. In JavaScript, the file is the boundary. In Go, the directory is the boundary. So the removed argument shadowed the variable that was already there.

PR #129, opened 2022-06-21
Adding automated release process

paul-sachs:
An automated release worklow. Basically adds an action that can be triggered whenever we want to do release.

1. Manually trigger action
2. Approve resulting PR with release notes
3. Github will automatically create the associated release

Todo:
- Add npm auth token so we can do the actual publish.


paul-sachs:
@timostamm added secret references. Not much of a way to test this action except to actually run it.

timostamm:
Closing this as it has become a bit stale with a concurrent effort in connect-web. We are still tracking this internally in TCN-273.

PR #166, opened 2022-07-26
Fix doc links in README

tp:
The links in the main README were missing the branch to point to, and thus would result in a 404.


timostamm (approved):
Thank you for catching that, Timm - much appreciated.

PR #167, opened 2022-07-26
Fix example link in Readme

seankimdev:
Similar to the docs link fix in the recently merged pr for the docs links, fixes the broken example link in the README.


timostamm (approved):
Thanks for catching this, Sean 🙏

PR #170, opened 2022-07-27
Update dependencies

timostamm:


packages/protobuf-bench/README.md
```diff
@@ -11,4 +11,4 @@ server would usually do.
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
 | protobuf-es         | 73,756 b      | 36,834 b | 9,562 b |
-| protobuf-javascript | 368,088 b  | 270,868 b | 43,782 b |
+| protobuf-javascript | 369,934 b  | 270,869 b | 43,728 b |
```
paul-sachs, line 14: Weird that a lint dependency increased the size 🤷

paul-sachs (approved):

packages/protobuf-bench/README.md
timostamm: Pretty sure it happened [here](https://github.com/bufbuild/protobuf-es/pull/165), where we updated esbuild. Need to work on the discipline to keep it up to date. (Can't automate, sizes fluctuate between mac os and linux).

paul-sachs: Ah, that makes sense I suppose we could use a macOS runner on github if we really wanted to.

PR #171, opened 2022-07-27
Respect json_name in messages created from a descriptor set

timostamm:
We have a feature that allows creating a message type at run time from a set of file descriptors. This feature has a bug - it does not respect the json_name field option.

This adds test coverage for this situation, and fixes the issue.


paul-sachs (approved):

doriable (approved):

akshayjshah (approved):
😱 Nice catch!

PR #172, opened 2022-07-27
call process.exit only after write completes

pkwarren:
During integration testing locally with larger BSR modules, we're seeing
some error messages that appear like the new TS based plugins aren't
fully flushing output.

Update the call to `process.write` to use a callback and invoke
`process.exit` from there to give everything a chance to fully write.

### DIFF
--- a/packages/protoplugin/src/run-node.ts
+++ b/packages/protoplugin/src/run-node.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import type { Plugin } from "./plugin.js";
-import type { ReadStream } from "tty";
+import type { ReadStream, WriteStream } from "tty";
 import { CodeGeneratorRequest } from "@bufbuild/protobuf";
 import { PluginOptionError, reasonToString } from "./error.js";
 
@@ -45,13 +45,9 @@ export function runNodeJs(plugin: Plugin): void {
     .then((data) => {
       const req = CodeGeneratorRequest.fromBinary(data);
       const res = plugin.run(req);
-      process.stdout.write(res.toBinary(), (err) => {
-        if (err) {
-          process.exit(1);
-        }
-        process.exit(0);
-      });
+      return writeBytes(process.stdout, res.toBinary());
     })
+    .then(() => process.exit(0))
     .catch((reason) => {
       const message =
         reason instanceof PluginOptionError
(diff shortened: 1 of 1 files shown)


packages/protoplugin/src/run-node.ts
```diff
@@ -75,18 +78,3 @@ function readBytes(stream: ReadStream): Promise<Uint8Array> {
...
   });
 }
-
-/**
- * Node.js buffers stdout, and process.exit() will truncate output.
- * As a workaround, we set the stream to blocking via a private API.
- * See https://github.com/timostamm/protobuf-ts/issues/134
- * See https://github.com/nodejs/node/issues/6456
- */
-function setBlockingStdout(): void {
```
pkwarren, line 85: Tested both with and without this workaround and it works without it so we can probably give this a shot to see if we can simplify the code.

paul-sachs (approved):

packages/protoplugin/src/run-node.ts
```diff
@@ -46,8 +45,12 @@ export function runNodeJs(plugin: Plugin): void {
...
-      process.stdout.write(res.toBinary());
-      process.exit(0);
+      process.stdout.write(res.toBinary(), (err) => {
+        if (err) {
+          process.exit(1);
+        }
+        process.exit(0);
+      });
     })
     .catch((reason) => {
```
timostamm, line 55: I think there is a potential issue with this. Node.js automatically quits when the script has executed. For example, consider this script:

```js
console.log("hello world");
```

ECMAScript promises are the cleanest solution to keep the event loop running for async operations. Consider:

```js
someOp().then(() => console.log("hello world"));
```

Node.js knows there's a pending promise, and won't quit before it resolves. Unfortunately, Node.js is older than Promises, and the entire network/stream API is based on event listeners and keeping the event loop running as long as listeners are attached.

The thing is, I'm not sure if the callback to stdout.write _consistently_ queues. Clearly it works in the cases we tested, but does it work on every platform/situation? I think we should err on the safe side, and use the promise chain we already maintain:

```suggestion
      return writeBytes(process.stdout, res.toBinary());
    })
    .then(() => process.exit(0))
    .catch((reason) => {
```

With the addition of:

```ts
/**
 * Write a chunk of bytes to a stream.
 */
function writeBytes(stream: WriteStream, data: Uint8Array): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    stream.write(data, (err) => {
      if (err) {
        reject(err);
      } else {
        resolve();
      }
    });
  });
}
```

pkwarren: Updated and verified locally.

timostamm (approved):
Thanks for the catch, and for the fix!

PR #174, opened 2022-07-28
Fix links in Markdown

timostamm:
Fixes two broken links in markdown files, a few links to source code where line numbers became incorrect, and consistently link to the registry when linking @bufbuild/*.

Updates TCN-259.

### DIFF
--- a/docs/generated_code.md
+++ b/docs/generated_code.md
@@ -94,7 +94,7 @@ For the following message declaration:
 message Example {}
 ```
 
-we generate a class called `Example`, which extends the base class [Message](https://github.com/bufbuild/protobuf-es/blob/tstamm/add-docs/packages/protobuf/src/message.ts#L40)
+we generate a class called `Example`, which extends the base class [Message](../packages/protobuf/src/message.ts#L40)
 provided by [@bufbuild/protobuf](../packages/protobuf). See the [runtime API documentation](#runtime-api) for details.
 
 Note that some names cannot be used as class names and will be escaped by adding the suffix `$`.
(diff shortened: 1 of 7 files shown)


docs/generated_code.md
```diff
@@ -94,8 +95,9 @@ For the following message declaration:
 message Example {}
 ```
 
-we generate a class called `Example`, which extends the base class [Message](https://github.com/bufbuild/protobuf-es/blob/tstamm/add-docs/packages/protobuf/src/message.ts#L40)
-provided by [@bufbuild/protobuf](../packages/protobuf). See the [runtime API documentation](#runtime-api) for details.
+we generate a class called `Example`, which extends the base class [Message](../packages/protobuf/src/message.ts#L40)
```
pkwarren, line 98: Do you want to use a permalink here to a specific revision? Otherwise the line number may not be accurate if changes are made to this file in the future.

docs/runtime_api.md
```diff
@@ -541,7 +541,7 @@ for examples how to include them. The JSON and binary serialization mechanisms u
 ### PartialMessage
 
 The object initializers accepted by message constructors are defined by the type 
-[`PartialMessage<T>`](../packages/protobuf/src/message.ts#L145-L153).
+[`PartialMessage<T>`](../packages/protobuf/src/message.ts#L143).
```
pkwarren, line 544: Same comment and below on lines 553/577 here re: using an exact revision.

pkwarren (approved):

docs/generated_code.md
timostamm: There are many other ways docs and code can run out of sync, but yeah, let's use permalinks for those. Done.

PR #190, opened 2022-08-08
Increase dependabot frequency

smaye81:
Increases the dependabot frequency to monthly.


timostamm (approved):
Similar to connect-web, this will reduce the noise without any disadvantages (we don't have any non-dev dependencies). Thank you!

PR #198, opened 2022-08-17
Add ability to test with a preconfigured list of TypeScript versions

smaye81:
This adds a new command to the Makefile for testing backwards-compatibility of TypeScript:

## test-ts-compat

Running this target will loop over all versions specified in `TS_VERSIONS` and use that version to compile `protobuf-es`.  

NOTE: the Make target will exit immediately when the first error is encountered.  

The versions list is compiled of:
- the earliest TypeScript version we support (`4.1.2`)
- the earliest version we support with `skipLibChecks` = `false` (`4.5.2`)
- the current release candidate (`4.8.1-rc`)
- the latest patch release of all minor versions in between

To reiterate, this illustrates `4.1.2` as the current lowest-supported TypeScript version for `protobuf-es` using default compiler options generated by `tsc`.  In addition, we support as low as `4.5.2` with `skipLibCheck` set to false.


packages/protobuf-test/typescript/ts3_9_4/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "include": ["../../src/**/*.ts"],
+  "exclude": ["../../src/**/*.test.ts", "../../src/gen/**/*.ts"],
```
timostamm, line 3: As discussed, we can't really exclude `gen`. It is totally fair if we limit the tests to extra/ though. Those are pretty exemplary for what a user would generate. It will also speed up the tests considerably.

packages/protobuf-test/typescript/ts4_0_2/tsconfig.json
```diff
@@ -0,0 +1,72 @@
...
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true,                       /* Allow javascript files to be compiled. */
+    // "checkJs": true,                       /* Report errors in .js files. */
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
```
timostamm, line 15: Those config files should make it clear to the user what we test, and with what options. 

I suggest that we remove all options that are commented out to reduce the file size to a fraction, and add a comment that states explicitly that those are the default options for TypeScript v4.0.2 created with `tsc --init`.

packages/protobuf-test/typescript/ts4_5_5/tsconfig.json
```diff
@@ -0,0 +1,103 @@
...
+    // "noFallthroughCasesInSwitch": true,               /* Enable error reporting for fallthrough cases in switch statements. */
+    // "noUncheckedIndexedAccess": true,                 /* Include 'undefined' in index signature results */
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true /* Skip type checking all .d.ts files. */
```
timostamm, line 101: It looks like this is the first version that doesn't require `skipLibCheck: true`. This is an option many advanced users will set to `false` deliberately for better type safety. We should guard against regression here. The simplest way would be to set it to `false`, and add an explanatory comment.

That means we can't use default `tsconfig.json` files the compiler generates with `--init`, but we can't use them as is anyways because we need to specify our sources.

smaye81: I was able to set `skipLibCheck` to `false` all the way back to `4.5.2` with no issues.

Makefile
```diff
@@ -154,6 +154,64 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
...
+	@for number in $(TS_VERSIONS) ; do \
+		formatted=$$(echo "$${number}" | sed -r 's/[\.]/_/g'); \
+		dirname=packages/protobuf-test/typescript/ts$${formatted} ; \
+		echo "Testing TypeScript `node_modules/ts$$formatted/bin/tsc --version`" ; \
+		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.json --noEmit > $$dirname/errors.log 2>&1; \
+		if [ "$$?" -eq 0 ]; then \
+	  		echo "Success" ; \
+			rm $$dirname/errors.log ; \
+  		else \
+	  		echo "Failed compilation.  Errors saved to $${dirname}/errors.log" ; \
```
timostamm, line 211: I see your point in redirecting the error output, but I'd prefer if we can keep this as simple as possible, especially in the Makefile. Failing early seems totally fine to me.

Makefile
```diff
@@ -154,6 +154,64 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
...
+			  4.5.3 \
+			  4.5.4 \
+			  4.5.5 \
+			  4.6.2 \
+			  4.6.3 \
+			  4.6.4 \
+			  4.7.2 \
+			  4.7.3 \
+			  4.7.4 \
+			  4.8.1-rc \
```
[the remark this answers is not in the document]
smaye81: OK.  Do you think it's worth testing the upcoming release candidate also?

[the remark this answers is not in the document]
smaye81: I kept in the release candidate.  I figure being proactive and testing with the latest RC can't hurt?

Makefile
```diff
@@ -154,6 +154,64 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
...
+
+.PHONY: install-ts
+install-ts:  node_modules
```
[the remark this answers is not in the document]
smaye81: Sounds good.  Removed.  We can implement this script later if we need it.  In the meantime, I'll document the process of adding a new TS version in a Notion doc or something.

Makefile
smaye81: Sounds good.  I kinda like redirecting the error output for two reasons:

1.  It's easier to debug errors that may occur rather than spewing them to the terminal (esp. if there's a lot)
2. Failing early means that any TS versions not tested could also be failing (and for different reasons) but you won't know until you fix the first error.

My preference would be to test all versions, write to a log, and then exit if failure.  But def not a huge deal to me.  I fixed this so that it just exits on the first failure and doesn't write to any log files.

Makefile
```diff
@@ -154,6 +154,26 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
 		--text_format_failure_list packages/protobuf-conformance/conformance_failing_tests_text_format.txt \
 		packages/protobuf-conformance/bin/conformance_esm.js
 
+TS_VERSIONS = 4.1.2 \
+			  4.2.4 \
+			  4.3.5 \
+			  4.4.4 \
+			  4.5.2 \
+			  4.6.4 \
+			  4.8.1-rc
```
timostamm, line 163: Can you move those to a single line, and add the line at the end of the other variable declarations at the top of the file?

Makefile
```diff
@@ -154,6 +154,26 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
...
+.PHONY: test-ts-compat
+test-ts-compat: $(GEN)/protobuf-test node_modules 
+	@for number in $(TS_VERSIONS) ; do \
+		formatted=$$(echo "$${number}" | sed -r 's/[\.]/_/g'); \
+		dirname=packages/protobuf-test/typescript ; \
+		echo "Testing TypeScript `node_modules/ts$$formatted/bin/tsc --version`" ; \
+		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --noEmit ; \
+		if [ "$$?" -ne 0 ]; then \
+			exit 1 ; \
+	  	fi ; \
```
timostamm, line 174: I would have expected a non-zero exit code from `tsc` to exit the loop. Turns out `-e` has some [surprising behavior](http://mywiki.wooledge.org/BashFAQ/105).

What do you think about simply using `||`?

```suggestion
		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --noEmit || exit; \
```

Makefile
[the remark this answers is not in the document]
timostamm: Can you add a short doc to packages/protobuf-test/README.md?

packages/protobuf-test/README.md
```diff
@@ -10,6 +10,69 @@ Many test cases are run several times, once with the generated TypeScript code,
...
+For example, assuming we are adding version `4.5.3`.  
+
+- Run the following command from the project root.  This will install the 
+specified version of TypeScript as a package alias to `protobuf-test`.  Note 
+that the version before the `@` sign is the version number with all decimals 
+replaced with underscores:
+
+```bash
+  npm i -w packages/protobuf-test ts4_5_3@npm:typescript@4.5.3 --save-exact
+```
```
timostamm, line 45: You can indent code block, but any empty lines in between need to indented too:

```suggestion
  
  ```bash
  npm i -w packages/protobuf-test ts4_5_3@npm:typescript@4.5.3 --save-exact
  ```
```

Note that line 42 has two spaces.

timostamm (approved):

PR #201, opened 2022-08-19
Point to commonjs in `main` property

smaye81:
Changing the `main` property to point to the Common JS module entrypoint.  Using ESM forces some workarounds to be needed for libraries like Jest and this seems to cause less friction by default.

### DIFF
--- a/packages/protobuf/package.json
+++ b/packages/protobuf/package.json
@@ -15,13 +15,13 @@
     "build:cjs": "npx tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm+types": "npx tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types"
   },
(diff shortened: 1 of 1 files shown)


packages/protobuf/package.json
```diff
@@ -15,13 +15,13 @@
...
   },
-  "main": "./dist/esm/index.js",
+  "main": "./dist/cjs/index.js",
   "type": "module",
   "types": "./dist/types/index.d.ts",
   "exports": {
     "import": "./dist/esm/index.js",
     "require": "./dist/cjs/index.js",
-    "default": "./dist/esm/index.js"
+    "default": "./dist/cjs/index.js"
```
timostamm, line 24: The [webpack documentation](https://webpack.js.org/guides/package-exports/#guidelines) recommends:

> Avoid the default export. It's handled differently between tooling. Only use named exports.

So let's remove it here, and the two instances in [protoplugin](https://github.com/bufbuild/protobuf-es/blob/9b4b79df20d660c5787631587d8ac1341ca9e62f/packages/protoplugin/package.json#L23) too.

timostamm (approved):

PR #202, opened 2022-08-22
Fix string-based fallback for 64-bit integral values

timostamm:
We use BigInt for 64-bit integral values. They parse rather quickly from the binary format with the help of DataView. In case BigInt (or the DataView methods we need) is unavailable, we fall back on a string-based representation. Unfortunately, this representation has a bug. This fixes the bug and adds test coverage.


packages/protobuf/src/google/varint.ts
```diff
@@ -161,19 +161,47 @@ export function int64fromString(dec: string): [boolean, number, number] {
...
+ * Losslessly converts a 64-bit signed integer in 32:32 split representation
+ * into a decimal string.
  *
- * Copyright 2008 Google Inc.
+ * Copyright 2008 Google Inc.  All rights reserved.
+ *
+ * See https://github.com/protocolbuffers/protobuf-javascript/blob/a428c58273abad07c66071d9753bc4d1289de426/experimental/runtime/int64.js#L10
  */
-export function int64toString(bitsLow: number, bitsHigh: number): string {
+export function int64ToString(lo: number, hi: number): string {
```
timostamm, line 175: This could have been a much smaller patch, but it seemed more reasonable to keep this in sync.

packages/protobuf-test/src/msg-scalar.test.ts
```diff
@@ -33,38 +32,30 @@ describeMT(
     > = {
       doubleField: 0,
       floatField: 0,
-      // @ts-expect-error TS2737
-      int64Field: 0n,
-      // @ts-expect-error TS2737
-      uint64Field: 0n,
+      int64Field: protoInt64.parse(0),
```
timostamm, line 35: This is required because we now run the tests with BigInt support disabled.

packages/protobuf-conformance/failing_tests_without_bigint.txt
```diff
@@ -0,0 +1,30 @@
+# proto2 extensions are not implemented
+# WARNING, test=Recommended.Proto2.JsonInput.FieldNameExtension.Validator: Expected JSON payload but got type 1. request=json_payload: "{\n        \"[protobuf_test_messages.proto2.extension_int32]\": 1\n      }" requested_output_format: JSON message_type: "protobuf_test_messages.proto2.TestAllTypesProto2" test_category: JSON_TEST, response=parse_error: "Error: cannot decode message protobuf_test_messages.proto2.TestAllTypesProto2 from JSON: key \"[protobuf_test_messages.proto2.extension_int32]\" is unknown"
+Recommended.Proto2.JsonInput.FieldNameExtension.Validator
+
+# Without BigInt support, a JavaScript runtime can only safely represent values
+# in the range from Number.MIN_SAFE_INTEGER to Number.MAX_SAFE_INTEGER.
+# That means number values in JSON will be unsafe as well. If you plan to use
+# protobuf-es on runtimes without BigInt support, make sure that the JSON data
+# you consume does contains 64-bit values as strings instead of numbers.
```
timostamm, line 9: It's worth noting that the [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) specifies a string for all 64-bit integrals. It also specifies that parsers should be lenient and accept numbers too, but we cannot reasonable conform to that without a bespoke JSON parser.

This is not an issue with BigInt support, because we convert numbers to BigInt right away, and modern runtimes don't lose precision on that path.

smaye81 (approved):

PR #207, opened 2022-08-25
Add code generation examples to protoc-gen-es README.md

timostamm:
Preview in GitHub: [README.md](https://github.com/bufbuild/protobuf-es/blob/tstamm/update-plugin-readme/packages/protoc-gen-es/README.md)


packages/protoc-gen-es/README.md
```diff
@@ -31,45 +24,92 @@ pete = Person.fromBinary(bytes);
...
+like [buf](https://github.com/bufbuild/buf) and [protoc](https://github.com/protocolbuffers/protobuf/releases).
+It generates base types - messages and enumerations - from your Protocol Buffer 
+schema. The generated requires the runtime library [@bufbuild/protobuf](https://www.npmjs.com/package/@bufbuild/protobuf).
+
+To install the plugin and the runtime library, run:
 
 ```shell
-npm install @bufbuild/protoc-gen-es
+npm install --save-dev @bufbuild/protoc-gen-es
+npm install @bufbuild/protobuf
```
timostamm, line 41: With npm, it isn't necessary to install the runtime library explicitly. It is a peer dependency of the plugin, so npm will install it automatically. But yarn and pnpm don't, so it seems a bit clearer to add it in this example anyways.

packages/protoc-gen-es/README.md
```diff
@@ -31,45 +24,92 @@ pete = Person.fromBinary(bytes);
...
 ```
 
-This will install the code generator plugin in `node_modules/.bin/protoc-gen-es`. 
-Note that npm does not add the executable to your `$PATH`. You can do so with:
+We use peer dependencies to ensure that code generator and runtime library are
+compatible with each other. Note that yarn and pnpm only emit a warning in this case.
```
timostamm, line 45: Hoping this might help a bit with the situation regarding https://github.com/bufbuild/connect-web/issues/239.

packages/protoc-gen-es/README.md
```diff
@@ -31,45 +24,92 @@ pete = Person.fromBinary(bytes);
...
+Add the following script to your `package.json`:
 
-```shell
-yarn add @bufbuild/protoc-gen-es
+```json
+{
+  "name": "your-package",
+  "version": "1.0.0",
+  "scripts": {
+    "generate": "buf generate"
```
timostamm, line 72: WDYT, @smaye81?

timostamm: Tested with pnpm, yarn, npm. They all add npm binaries to the $PATH.

packages/protoc-gen-es/README.md
```diff
@@ -31,45 +24,92 @@ pete = Person.fromBinary(bytes);
...
+Learn more about the project at [github.com/bufbuild/protobuf-es](https://github.com/bufbuild/protobuf-es).
 
 
 ## Installation
 
-### With npm
+`protoc-gen-es` is a code generator plugin for Protocol Buffer compilers,
+like [buf](https://github.com/bufbuild/buf) and [protoc](https://github.com/protocolbuffers/protobuf/releases).
+It generates base types - messages and enumerations - from your Protocol Buffer 
+schema. The generated requires the runtime library [@bufbuild/protobuf](https://www.npmjs.com/package/@bufbuild/protobuf).
```
smaye81, line 35: Should this be:

The generated **code**?

packages/protoc-gen-es/README.md
```diff
@@ -31,45 +24,92 @@ pete = Person.fromBinary(bytes);
...
+  protoc -I . \
+  --es_out src/gen \
+  --os_opt target=ts \
+  a.proto b.proto c.proto
 ```
 
+Note that we are adding `node_modules/.bin` to the `$PATH`, so that the protocol
+buffer compiler can find them. This happens automatically with npm scripts.
+
+Since yarn does not use a `node_modules` directory, you need to change the variable
```
smaye81, line 101: May want to mention this is yarn 2 and above.

smaye81 (approved):

Issue #209, opened 2022-08-30
Is `moduleResolution=Node` a requirement?


faustbrian:
I'm currently migrating a project to ESM and noticed that all my `buf generate` commands fail when I use `"moduleResolution": "NodeNext"` in my `tsconfig.json` files. Is the use of `"moduleResolution": "Node"` an absolute requirement or is there some way to work around this so that `NodeNext` can be used as resolution mechanism?

```
source/index_pb.ts:42:38 - error TS2576: Property 'fromBinary' does not exist on type 'TransactionChecksum'. Did you mean to access the static member 'TransactionChecksum.fromBinary' instead?

42     return new TransactionChecksum().fromBinary(bytes, options);
                                        ~~~~~~~~~~

source/index_pb.ts:46:38 - error TS2576: Property 'fromJson' does not exist on type 'TransactionChecksum'. Did you mean to access the static member 'TransactionChecksum.fromJson' instead?

46     return new TransactionChecksum().fromJson(jsonValue, options);
                                        ~~~~~~~~

source/index_pb.ts:50:38 - error TS2576: Property 'fromJsonString' does not exist on type 'TransactionChecksum'. Did you mean to access the static member 'TransactionChecksum.fromJsonString' instead?

50     return new TransactionChecksum().fromJsonString(jsonString, options);
```

smaye81:
Hey @faustbrian.  `moduleResolution="Node"` is definitely not a requirement.  After some digging it turns out that this is due to a missing `types` property in the conditional exports in our `package.json` file.  I will put up a PR to fix this.  Thanks for pointing this out!

PR #210, opened 2022-08-30
Fix type prop issue in conditional exports

smaye81:
Fixes https://github.com/bufbuild/protobuf-es/issues/209

If consumers try to import `protobuf-es` and configure their `tsconfig.json` file to use `moduleResolution: "NodeNext"`, they experience issues resolving our declaration files.  This is because we specify conditional exports and the `NodeNext` module resolution will use this exports map to resolve modules.  Unfortunately, it also uses this to resolve `types`.  Since we didn't specify a `types` property there, declaration files couldn't be found.

For more context:  
- https://github.com/microsoft/TypeScript/issues/48235
- https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#esm-nodejs

In addition, this modifies our TypeScript tests to test against 4.7.4 and 4.8.2 using `NodeNext` module resolution (usage of this value with anything below 4.7 errors out because it is marked experimental in those versions).


timostamm (approved):
We should do the same for @bufbuild/protoplugin, but we can follow-up along with other improvements to this package. LGTM

smaye81:
Yep, I am planning to release a patch release today that fixes this.  Just want to test a few things before I do.  I will post here when I do.

smaye81:
@faustbrian We just published v0.1.1 of `protobuf-es`, so you should be able to install and get support for `NodeNext`.  Let me know if this fixes your issue.

smaye81:
Can you share the `compilerOptions` from your `tsconfig.json` file?  I am able to build with v0.1.1 and `NodeNext` in my config.

smaye81:
OK, I am able to recreate with your repo and in a sample repo I have here.  The issue seems to be with generating declaration files with a moduleResolution of "NodeNext".  A few different magic incantations of compilerOptions will get around the issue, but obviously that's just a band-aid I think.  For example:

- setting `noEmit` to false will compile successfully.
- setting `noEmit` to true, but `declaration` and `declarationMap` to  false will compile successfully.
- setting `moduleResolution` to `Node` will compile successfully.

I see what the issue is on our side and why this is happening.  What's curious to me is why it's only happening with `NodeNext` module resolution.  I will discuss the best way forward with our team and keep you posted.  In the meantime, you may have to use one of the above workarounds.  Apologies for the inconvenience.

smaye81:
Alternatively, you could set your target out to be `js+dts` in your `buf.gen.yaml`.  This way the TypeScript compiler won't compile the generated protobuf-es code.  This should also work and still allow you to emit declaration files with `NodeNext` module resolution.

PR #228, opened 2022-09-14
Add the option to transpile target files

smaye81:
This adds the ability to transpile generated `.ts` files into `.js` and/or `.d.ts` files.  Users can provide the following 3 functions to generate code in their plugin:

- `generateTs`
- `generateJs`
- `generateDts`

The `generateTs` function is required, but the latter two are optional.  If one or the other (or both) are not provided, then the framework will auto-transpile the generated TypeScript files to generate the desired code.  Note that this is all dependent on the provided target outs.  For example, if `target=js` and no `generateJs` or `generateDts` is provided, the framework will only transpile JavaScript files (ignoring declaration files).  In essence, the framework will only transpile files if needed and even then, will only transpile the ones being requested.

It is worth noting that transpilation is an expensive process so if performance is a concern, plugin authors should pass all 3 functions as the plugin code generation process is much faster (`protoc-gen-es` uses this approach).

When auto-transpiling, the framework uses an older, pinned version of TypeScript for stability (at press time `4.5.2`).  The compiler options used are intentionally lenient so that the odds of generating code under most circumstances are high.  If this is not desirable for plugin authors, they also have the option to provide their own `transpile` function as part of the plugin initialization configuration.  If they provide their own, then this `transpile` function will be invoked instead and the framework's transpilation process will be bypassed.

For an example of implementing a transpiler, see `packages/protoplugin/src/ecmascript/transpile.ts`.


packages/protoplugin/src/ecmascript/generated-file.ts
```diff
@@ -37,6 +37,27 @@ type Printable =
   | DescEnum
   | Printable[];
 
+export interface TSFile {
```
smaye81, line 40: Temporary name because I couldn't think of one better.  I considered `OutFile`, but not sure if that's best.

packages/protoc-gen-es/src/transpile.ts
```diff
@@ -0,0 +1,189 @@
...
+      }
+    }
+  }
+
+  lookupVirtualFile(fileName: string): TSFile | undefined {
+    let vf = this._files.get(fileName);
+    if (vf) return vf;
+    let cwd = process.cwd();
+    if (fileName.startsWith(cwd)) {
+      let relativePath = path.relative(cwd, fileName);
```
timostamm, line 50: I suspect that we can use `getCurrentDirectory()` and/or `getCanonicalFileName()` to avoid `process.cwd()`. Can you look into that? It seems wrong to need the cwd here.

timostamm (commented):
Very excited for this :)

packages/example/package.json
```diff
@@ -6,12 +6,11 @@
     "clean": "rm -rf ./dist/esm/* ./dist/types/*",
     "build": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "add-person": "node dist/esm/add-person.js addressbook.bin",
-    "list-people": "node dist/esm/list-people.js addressbook.bin"
+    "list-people": "node dist/esm/list-people.js addressbook.bin",
+    "buf:generate": "buf generate"
```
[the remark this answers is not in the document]
smaye81: Nice!  Done.

packages/example/package.json
```diff
@@ -6,12 +6,11 @@
...
   },
   "type": "module",
   "dependencies": {
-    "@bufbuild/protobuf": "0.1.1",
-    "@bufbuild/protoc-gen-es": "0.1.1",
```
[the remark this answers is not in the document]
smaye81: Ah yeah.  This is just temporary in my local setup I think because I have the libraries linked so I can run against my local changes.  I will revert this before it lands.

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -59,10 +71,53 @@ export function createEcmaScriptPlugin(
...
+        if (init.transpile) {
+          // Transpile the generated files and add to the master list of files
+          const transpiledFiles = init.transpile(
+            files,
+            transpileJs,
+            transpileDts
+          );
+          files.push(...transpiledFiles);
+        } else {
+          throw new PluginGenerationError(
```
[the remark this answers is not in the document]
smaye81: Added a TODO

packages/protoplugin/src/ecmascript/schema.ts
```diff
@@ -79,7 +76,7 @@ export interface Schema {
 
 interface SchemaController {
   schema: Schema;
-  toResponse: (res: CodeGeneratorResponse) => void;
+  toTSFile: () => TSFile[];
```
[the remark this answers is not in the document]
smaye81: Added a TODO on Line 137

packages/protoc-gen-es/src/transpile.ts
```diff
@@ -0,0 +1,189 @@
...
+    );
+  }
+
+  getDefaultLibFileName(options: ts.CompilerOptions): string {
+    return this.wrapped.getDefaultLibFileName(options);
+  }
+
+  getNewLine(): string {
+    return this.wrapped.getNewLine();
+  }
```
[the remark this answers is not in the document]
smaye81: Not sure if this is still relevant with the use of `@typescript/vfs` now.

packages/protoc-gen-es/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "files": ["src/protoc-gen-es-plugin.ts"],
+  "files": ["src/protoc-gen-es-plugin.ts", "src/transpile.ts"],
```
smaye81, line 2: Had to add this temporarily so that linting passes.  Once we decide where to put the `transpile.ts` file, we can fix this.

packages/protoc-gen-es/src/transpile.ts
```diff
@@ -0,0 +1,104 @@
...
+import * as ts from "typescript";
+import {
+  createDefaultMapFromNodeModules,
+  createSystem,
+  createVirtualCompilerHost,
+} from "@typescript/vfs";
+
+function createTranspiler(options: ts.CompilerOptions, files: TSFile[]) {
+  const fsMap = createDefaultMapFromNodeModules({
+    target: ts.ScriptTarget.ES2015,
```
smaye81, line 25: This `ScriptTarget` value may need to change (or we can just use the `options` passed in since this accepts `CompilerOptions`.  We just need to solidify what are default options are.

packages/protoplugin/src/ecmascript/schema.ts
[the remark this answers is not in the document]
smaye81: Tested this and verified that it works fine without having to evict the JS files (or the DTS files if a JS generator isn't provided).

packages/protoplugin/src/ecmascript/transpile.ts
```diff
@@ -0,0 +1,103 @@
...
+      const file = files.find((x) => sourceFiles[0].fileName === x.name);
+      if (!file) {
+        err = new Error(
+          `unable to map emitted file "${fileName}" to a source file: not found`
+        );
+        return;
+      }
+      results.push({
+        name: fileName,
+        preamble: file.preamble,
```
smaye81, line 94: Is it worth adding a line here about TypeScript?  Something like `// transpiled with TypeScript v4.5.2`?

packages/protoplugin/src/ecmascript/transpile.ts
```diff
@@ -0,0 +1,103 @@
...
+export function transpile(
+  files: FileInfo[],
+  transpileJs: boolean,
+  transpileDts: boolean
+): FileInfo[] {
+  // TODO - Need more options here by default
+  const options: ts.CompilerOptions = {
+    declaration: transpileDts,
+    emitDeclarationOnly: transpileDts && !transpileJs,
+  };
```
timostamm, line 53: We do. I think the general goal here should be to emit code, even if it is not completely correct.

packages/protoplugin/src/ecmascript/transpile.ts
timostamm: That would be pretty awesome, but I don't know how we can achieve that. 

If you look at the preamble, you see that it also contains package comments, and we can't really put this info next to our `@generated ...` annotations:

https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf-test/src/gen/ts/extra/comments_pb.ts#L17-L25

packages/protoplugin-test/src/initial.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  expect(resp.file.length).toEqual(targets.length);
+  targets.forEach((target) => {
+    const expectedFiles = fixture[target];
+    expectedFiles.forEach((file) => {
+      expect(resp.file.findIndex((e) => e.name === file)).toBeGreaterThan(-1);
+    });
+  });
+}
+
+describe("all generators with variant target outs", function () {
```
timostamm, line 61: These tests are great for a start, but we do need coverage for imports that are not generated.

Can you add a second proto file, and import it from test.proto, but _not_ add it to `file_to_generate` in the code generator request?

packages/protoplugin/src/ecmascript/transpile.ts
```diff
@@ -0,0 +1,103 @@
...
+import ts from "typescript";
+import {
+  createDefaultMapFromNodeModules,
+  createSystem,
+  createVirtualCompilerHost,
+} from "@typescript/vfs";
+
+/* eslint-disable import/no-named-as-default-member */
+
+function createTranspiler(options: ts.CompilerOptions, files: FileInfo[]) {
```
timostamm, line 25: I forgot: Can you add a comment somewhere that explains 1. that we intentionally transpile with a pinned older version of TypeScript for stability, and 2. that we put @typescript/vfs in the top-level package.json in a conflicting version to avoid npm hoisting it to the top-level node_modules directory, where it would pick up the unpinned TypeScript version?

packages/protoplugin/package.json
```diff
@@ -35,12 +35,15 @@
     }
   },
   "dependencies": {
-    "@bufbuild/protobuf": "0.1.1"
-  },
-  "devDependencies": {
-    "typescript": "^4.8.2"
+    "@bufbuild/protobuf": "0.1.1",
+    "@typescript/vfs": "^1.4.0",
+    "lz-string": "^1.4.4",
```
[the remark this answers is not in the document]
smaye81: We do unfortunately.  Without it, the build process complains with:

`node_modules/@typescript/vfs/dist/index.d.ts:57:140 - error TS2307: Cannot find module 'lz-string' or its corresponding type declarations.`

However, in going back to verify this, I realized we only need the `@types/lz-string` dep, so I removed the other one.

packages/protoplugin-test/src/initial.test.ts
smaye81: Yep.  Done.  Let me know what you think.

packages/protoplugin/src/ecmascript/schema.ts
```diff
@@ -132,17 +133,39 @@ export function createSchema(
...
+      // retrieve only the ts files here. If that turns out to be the case, we can change the method behavior
+      // to evict the generated files. As in "get the files generated so far, and remove them from
+      // the schema controller".
+      return generatedFiles.flatMap((file) => {
+        const fileInfo = file.getFileInfo();
+        // undefined is returned if the file has no content
+        if (!fileInfo) {
+          return [];
+        }
+        return [fileInfo];
```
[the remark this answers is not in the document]
smaye81: I basically wanted to `map` but only add elements conditionally.  The [official documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map) for `flatMap` recommends doing it this way.

I could also use a `reduce` here, which may be better.  I'm ambivalent so I'm open to changing it if you think it's worthwhile.

packages/protoplugin/src/ecmascript/transpile.ts
smaye81: What about putting it after the eslint/ts-nocheck directives?  That is actually pretty easy to do:

image

packages/protoplugin/src/ecmascript/transpile.ts
smaye81: I added the default compiler options generated by `tsc --init` on 4.5.2.  Let me know if you think there are others we should include.

packages/protoplugin/src/ecmascript/transpile.ts
timostamm: Package comments are rare, but there are valid use cases for them. I don't think we can append our own comments to the package comments.

packages/protoc-gen-es/src/transpile.ts
timostamm: Let's use ES2017 as the script target, following https://github.com/bufbuild/connect-web#supported-browsers

timostamm: Oh, I see we're setting module CommonJS in the line below. We definitely need to use ECMAScript module imports here, and need to make sure we include the `.js` extension. I think this is worth a test case.

packages/protoplugin/src/ecmascript/transpile.ts
timostamm: Synced offline - we need to do the opposite here, and be very lenient.

I think this should work:

```ts
    // Type checking
    strict: false,

    // modules
    module: ts.ModuleKind.ES2020,
    moduleResolution: ts.ModuleResolutionKind.NodeJs,
    noResolve: true,
    resolveJsonModule: false,

    // emit
    emitBOM: false,
    importsNotUsedAsValues: ts.ImportsNotUsedAsValues.Preserve,
    newLine: ts.NewLineKind.LineFeed,
    preserveValueImports: false,

    // JavaScript Support
    allowJs: true,
    checkJs: false,

    // Language and Environment
    lib: [],
    moduleDetection: "force",
    target: ts.ScriptTarget.ES2017,

    // Completeness
    skipLibCheck: true,
    skipDefaultLibCheck: false,
```

I haven't checked if this is the right combination we want for ESM imports. Please adjust as necessary.

packages/protoplugin/src/ecmascript/transpile.ts
```diff
@@ -0,0 +1,155 @@
...
+ * top-level package as well as this package.  This is to avoid npm hoisting
+ * @typescript/vfs to the top-level node_modules directory, which then causes
+ * type mismatches when trying to use it with this package's version of
+ * TypeScript.  Ideally we would use something like Yarn's nohoist here, but
+ * npm does not support that yet.
+ */
+function createTranspiler(options: ts.CompilerOptions, files: FileInfo[]) {
+  const fsMap = createDefaultMapFromNodeModules({
+    target: options.target,
+  });
```
timostamm, line 80: Looking at the [source of this function](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescript-vfs/src/index.ts#L194), I think we should be able to replace this with just `const fsMap = new Map<string, string>()`.

The behavior of `createDefaultMapFromNodeModules` is:
> Sets up a Map with lib contents by grabbing the necessary files from the local copy of typescript via the file system.

This behavior is actually undesirable in our case, because we want to avoid hitting the file system. We don't want to introduce any variance in the output depending on the local file system, and we intentionally ignore any resolution errors anyway.

I _think_ `createDefaultMapFromNodeModules` doesn't do anything because we pass an empty `lib` option, so this change will probably not affect the outcome. But not using the function in the first place would remove any doubt, and make the intent clear.

WDYT?

timostamm (commented):
This looks pretty great to me, just needs a little bit of tidying up, see the two comments below.

Can you remove `packages/protoplugin-test/descriptorset.bin` and then add it to `.gitignore`? I know we check in generated source files, but this binary file is not very useful in git.

packages/protoplugin/src/ecmascript/transpile.ts
smaye81: So, it's weird.  When I was first coding this, I used the empty `Map` example you have and was getting errors with transpilation.  After reading the docs, it seemed I needed to add this `createDefaultMap...` call.  Once I did that, everything started working.

But, when I take it out now, things still seem to be working, so I'm not sure if something I did also fixed that issue.  All tests are passing, so I'm guessing it's OK?  I'll remove and we can revisit if something blows up.

packages/protoplugin/src/ecmascript/schema.ts
```diff
@@ -132,17 +133,37 @@ export function createSchema(
...
-        CodeGeneratorResponse_Feature.PROTO3_OPTIONAL
-      );
-      for (const genFile of generatedFiles) {
-        genFile.toResponse(res);
-      }
+    getFileInfo() {
+      // TODO - If all targets are requested, but the plugin doesn't have a generateDts, we might need to
+      // retrieve only the ts files here. If that turns out to be the case, we can change the method behavior
+      // to evict the generated files. As in "get the files generated so far, and remove them from
+      // the schema controller".
```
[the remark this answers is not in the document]
smaye81: Well things work, but we still see those pre-emit diagnostics error messages, such as:

`"File 'proto/address_book_proto.js' is a JavaScript file. Did you mean to enable the 'allowJs' option?",`

and

`"Cannot redeclare block-scoped variable 'test'."`

which are both related to us passing the `.js` files to the transpiler.  I think ideally we'd like to only send `.ts` files here.  I can't think of a scenario where we'd ever want to pass `.js` files.  We require a `generateTs` function, so users would never be able to only specify a `generateJs` function and expect to transpile declarations.

What are your thoughts?  If we can't think of a reason why a user would want to send JS through the transpiler, then I think I should just prevent that from happening here.  In looking at it, it would be pretty easy.  I'll just save off the results of `getFileInfo` after the call to `generateTs` and only send that array to the transpiler if needed.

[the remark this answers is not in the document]
timostamm: > I'll just save off the results of getFileInfo after the call to generateTs and only send that array to the transpiler if needed.

Sounds like a good plan :+1:

packages/protoplugin/src/ecmascript/transpile.ts
timostamm: The tests are pretty lightweight, and don't use any imports or lib types. But I think this is fine as is for now. 

Once we add documentation for this feature, we can provide a simple example (maybe a simple Twirp client generator using fetch - it would use both imports and lib types then), and gain confidence that it works as expected.

timostamm (approved):

PR #229, opened 2022-09-23
Add a plugin option to keep empty files

timostamm:
By default, protoc-gen-es (and all other plugins based on @bufbuild/protoplugin) omit empty files from the plugin output.

This is reasonable behaviour, because a file without contents is meaningless, and plugin authors often run into a situation where they only find out that they don't have anything to generate deep into some loop, long after they have constructed a file to generate. Users may have protobuf files that only declare services, which does not generate any code with protoc-gen-es.

Unfortunately, some tooling like Bazel requires all output files to be declared ahead of time, which makes it impossible to generate code through Bazel without looking at the contents of the protobuf sources.

All standard base type generators do create files for empty sources files. Most contain somewhat useful code (a file descriptor), in case of the JavaScript generator the files only contain unexported initialization code. But all do create a file.

This adds the plugin option `keep_empty_files=true` to @bufbuild/protoplugin. Bazel users can use this option to ensure that every input file has an output file, even though it may be empty. For all other users, this option is irrelevant.

An alternative to this option would be to generate empty files in protoc-gen-es exclusively. On the upside, we would not need an additional option. On the downside, we would generate meaningless files in some situations, and RPC generators could still not be used in combination with Bazel.

### DIFF
--- a/packages/protoplugin/src/create-es-plugin.ts
+++ b/packages/protoplugin/src/create-es-plugin.ts
@@ -48,8 +48,13 @@ export function createEcmaScriptPlugin(
     name: init.name,
     version: init.version,
     run(req) {
-      const { targets, tsNocheck, bootstrapWkt, rewriteImports } =
-        parseParameter(req.parameter, init.parseOption);
+      const {
+        targets,
+        tsNocheck,
+        bootstrapWkt,
+        rewriteImports,
+        keepEmptyFiles,
+      } = parseParameter(req.parameter, init.parseOption);
       const { schema, toResponse } = createSchema(
         req,
         targets,
@@ -57,7 +62,8 @@ export function createEcmaScriptPlugin(
         init.version,
         tsNocheck,
         bootstrapWkt,
-        rewriteImports
+        rewriteImports,
+        keepEmptyFiles
       );
       generateFn(schema);
       const res = new CodeGeneratorResponse();
@@ -74,6 +80,7 @@ function parseParameter(
   let targets: Target[] = ["js", "dts"];
   let tsNocheck = true;
   let bootstrapWkt = false;
+  let keepEmptyFiles = false;
   const rewriteImports: RewriteImports = [];
   for (const { key, value } of splitParameter(parameter)) {
     switch (key) {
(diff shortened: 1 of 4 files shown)


jcready:
> Unfortunately, some tooling like Bazel requires all output files to be declared ahead of time

Yeah, Bazel is **extremely** strict about outputting files. This change will be great for us though! We previously just had to patch the dist generated-file.js file to comment out that `if (content.length === 0) {` block. Now if we could just get that `use_proto_field_name` option we could get rid of another patch 😬

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -242,6 +249,21 @@ function parseParameter(
         rewriteImports.push({ pattern, target });
         break;
       }
+      case "keep_empty_files": {
```
smaye81, line 252: Could we shorten this to:

```js
case "keep_empty_files": {
   keepEmptyFiles = value === "true" || "1"
}
```

smaye81:
I think we also need to fix this line:

https://github.com/bufbuild/protobuf-es/pull/229/files#diff-0a9300e9a6f8ffd00fe00474f9c662da864c48cf18be0d30260cde31a3fb40c9R146

So that we only return an empty array if `keepEmptyFiles` is false.

smaye81 (approved):

packages/protoplugin/src/create-es-plugin.ts
timostamm: It would need to be `value == "true" || value === "1"`. We _could_ do that, but I think it's better to be exact here. For example `keep_empty_files=no` should error.

timostamm:
> I think we also need to fix line 146 in schema.ts, so that we only return an empty array if `keepEmptyFiles` is false.

Just tested this with [empty-file.proto](https://github.com/bufbuild/protobuf-es/blob/tstamm/empty-files/packages/protobuf-test/extra/empty-file.proto) - the single change in generated-file.ts is enough.

timostamm:
> Yeah, Bazel is **extremely** strict about outputting files. This change will be great for us though! We previously just had to patch the dist generated-file.js file to comment out that `if (content.length === 0) {` block. Now if we could just get that `use_proto_field_name` option we could get rid of another patch 😬

A pleasure seeing you over here, @jcready. Glad this will solve a problem for you - proto field names are a different story, but maybe we can find a decent way to support that too, in the future.

PR #244, opened 2022-10-03
Add convenience functions for retrieving custom options

smaye81:
This adds convenience functions for retrieving custom options defined in protobuf files.  

Note that this does not involve the support of extensions yet (or custom options for that matter).  These convenience functions are intended as a stopgap until full support for proto2 extensions and custom options can be implemented.


packages/protoplugin/src/ecmascript/custom-options.ts
```diff
@@ -0,0 +1,367 @@
...
+      }
+      reader = new BinaryReader(field.data);
+    }
+  }
+  return reader;
+}
+
+type ScalarValue = number | bigint | string | undefined;
+
+export function getCustomScalarOption(
```
smaye81, line 52: Put this as an example for maybe having a single function for scalar values instead of one per type.

packages/protoplugin-test/src/custom-options.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  getCustomOptionInt32,
+  getCustomOptionInt64,
+  getCustomOptionString,
+  getCustomOptionMessage,
+  getCustomOptionEnum,
+} from "@bufbuild/protoplugin/ecmascript";
+import { proto3 } from "@bufbuild/protobuf";
+
+// Used for custom options of type FooMessage
+const foo = proto3.makeMessageType("FooMessage", [
```
timostamm, line 30: As the author of a plugin with custom options, you provide two things:
1. the code generator plugin
2. the proto file defining the custom options

Since you are in control, you can simply generate the code for the proto file, and use the generated `class FooMessage` instead of `makeMessageType`. That should be a lot easier to use for plugin authors?

Sorry if I lead you on the wrong path with `makeMessageType`. The idea was to get the field information from the schema, which is something we have to do anyways if we want to expose custom options in the generated code. We can get by without it.

packages/protoplugin-test/src/custom-options.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  });
+  describe("finds options correctly", function () {
+    test("file options", () => {
+      for (const file of files) {
+        expect(getCustomOptionString(file, 50000)).toEqual("Hello");
+      }
+    });
+    test("enum options", () => {
+      for (const file of files) {
+        for (const enumeration of file.enums) {
```
timostamm, line 89: Iterating over the files is going to be brittle: If you add another proto file to packages/protoplugin/test/proto, these tests will fail. 

What do you think about changing the approach a bit, and look up an enumeration by name? `descriptorSet.enums.get("foo")` This also works for messages, files, etc.

packages/protoplugin-test/src/custom-options.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  },
+]);
+
+describe("custom options", function () {
+  let files: DescFile[] = [];
+  beforeEach(() => {
+    files = getDescriptorSet().files.filter(
+      (file) => file.name === "proto/custom_options"
+    );
+  });
```
timostamm, line 80: `beforeEach` is good practice to make sure no state is shared across tests, but here it will read the descriptors from disk and parse them for every single test. We have to avoid that to keep the test suite running fast.

Can you switch to `beforeAll`? In my book, it is also an option to just define a `const` in the describe block. Same effect and reads nicer.

packages/protoplugin/src/ecmascript/custom-options.ts
timostamm: Makes a lot of sense to me! 

If we deliberately don't support repeated fields (which is reasonable), a function each for scalar, message, and enum fields seems like a decent API.

What do you think about letting it return a `string` for `ScalarType.STRING`, a `boolean` for `BOOL` etc?

packages/protoplugin-test/src/custom-options.test.ts
smaye81: I am actually filtering out only the `custom_options` proto in the `beforeEach` currently, so that _should_ protect against tests failing if another file is added right?

Regardless, I'm totally cool with changing them to access the internal maps, but just wanted to see if my assumption was true.

timostamm: Sorry, I missed the filter. Still makes sense to me to be very explicit about picking something from the descriptor set, it will make it much easier to add cases if needed.

For example:

```ts
      const enumeration = descriptorSet.enums.get("FooEnum");
      assert(enumeration);
      expect(getCustomOptionBoolean(enumeration, 50004)).toBeTruthy();

      const enumValue = descriptorSet.enums.get("FooEnum")?.values.find(v => v.name === "OFF");
      assert(enumValue);
      expect(getCustomOptionBoolean(enumValue, 50004)).toBeTruthy();
```

Where `assert()` could be copied from assert.ts into a test helper. If the assertion fails, test data is missing - the tests are broken. If the expectation fails, the tests are good, but the tested code failed the test.

packages/protoplugin-test/src/custom-options.test.ts
```diff
@@ -0,0 +1,242 @@
...
+      const service = descriptorSet.services.get("example.FooService");
+      assert(service);
+      for (const method of service.methods) {
+        const option = findCustomMessageOption(method, 50007, FooMessage);
+        expect(option?.foo).toEqual(567);
+        expect(option?.bar).toEqual("Some string");
+        expect(option?.qux?.case).toEqual("quux");
+        expect(option?.qux?.value).toEqual("Oneof string");
+        expect(option?.many).toEqual(["a", "b", "c"]);
+        expect(option?.mapping).toEqual({ testKey: "testVal" });
```
smaye81, line 102: @timostamm If it's OK with you, I'd like to leave these additional expectations here just as examples for how to read values from a message option.  If you feel strongly about it, I can remove though.

packages/protoplugin-test/src/custom-options.test.ts
```diff
@@ -0,0 +1,242 @@
...
+          expect(
+            findCustomScalarOption(message, 99999, ScalarType.INT32)
+          ).toBeUndefined();
+        }
+      }
+    });
+    test("field options", () => {
+      for (const file of descriptorSet.files) {
+        for (const message of file.messages) {
+          for (const member of message.members) {
```
timostamm, line 146: This should also look up by name. Otherwise, renaming a field in the proto source will skip the assertions in this test without us noticing. 

For example:

```ts
      const m = descriptorSet.messages.get("example.FooService")?.members.find(m => m.name === "qux");
      assert(m);
      assert(m.kind === "oneof");
      findCustomScalarOption(m, 99999, ScalarType.INT64)
```

.gitignore
```diff
@@ -2,6 +2,7 @@ node_modules
 /packages/*/dist
 /packages/protobuf-test/descriptorset.bin
 /packages/protoplugin-test/descriptorset.bin
+/packages/protoplugin-test/src/gen/
 /testdata/gen/**/*.js
 /testdata/gen/**/*.ts
 /testdata/gen/**/**.d.ts
```
timostamm, line 8: Huh, those have been irrelevant for a _long_ time. Can you delete them?

PR #245, opened 2022-10-03
Dependency bump

paul-sachs:
Dependabot batched updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,18 +17,18 @@
     "npm": ">=8"
   },
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^5.36.1",
-    "@typescript-eslint/parser": "^5.36.1",
-    "eslint": "^8.23.0",
-    "eslint-import-resolver-typescript": "^3.4.1",
+    "@typescript-eslint/eslint-plugin": "^5.38.1",
+    "@typescript-eslint/parser": "^5.38.1",
+    "eslint": "^8.24.0",
+    "eslint-import-resolver-typescript": "^3.5.1",
     "eslint-plugin-import": "^2.26.0",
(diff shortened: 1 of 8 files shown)


smaye81 (approved):

package.json
```diff
@@ -17,18 +17,18 @@
...
     "eslint-plugin-node": "^11.1.0",
     "prettier": "^2.7.1",
-    "typescript": "^4.8.2"
+    "typescript": "^4.8.4"
   },
   "dependencies": {
-    "@typescript/vfs": "1.0.0",
-    "@types/jest": "^29.0.3",
-    "jest": "^29.0.3"
+    "@typescript/vfs": "1.4.0",
```
timostamm, line 30: We have to keep this at 1.0.0. We use this pin to avoid npm from hoisting this dependency in packages/protoplugin. It's a terrible hack.

packages/protoplugin/package.json
```diff
@@ -37,7 +37,7 @@
   "dependencies": {
     "@bufbuild/protobuf": "0.1.1",
     "@typescript/vfs": "^1.4.0",
-    "typescript": "4.5.2"
+    "typescript": "4.8.4"
```
timostamm, line 40: We should keep this pinned at the older version. We deliberately don't want to keep this up to date, because updating the TypeScript compiler can change to transpiled code, and we want the generated code to be stable. 

https://github.com/bufbuild/protobuf-es/blob/5ae3246385fec75f80a9cb058f9c8abc48912611/packages/protoplugin/src/ecmascript/transpile.ts#L56-L58

timostamm (commented):
The recently added feature in @bufbuild/protoplugin means we have to retain some pins.

PR #246, opened 2022-10-05
Revert dependency updates

smaye81:
We explicitly pin `typescript` and `@typescript/vfs` to specific versions in our top-level package for usage with the plugin transpilation process.  So these should be excluded from Dependabot.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -14,4 +14,9 @@ updates:
       day: "monday"
       timezone: UTC
       time: "07:00"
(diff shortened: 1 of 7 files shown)


timostamm (commented):
This makes sense to me.

While we are tidying up - `protoplugin-test` and `protobuf-conformance` have `typescript` in their `dependencies`, but both use the compiler from the root of the workspace. Can you try removing those two dependencies?

timostamm (approved):

PR #247, opened 2022-10-06
Add documentation for working with plugin framework

smaye81:
This adds an extensive README to `protoplugin` which illustrates how to create plugins using the plugin framework.

For a more easy-to-read view:  https://github.com/bufbuild/protobuf-es/blob/sayers/plugin_docs/docs/writing_plugins.md


docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+See the [`PluginInit`](#plugininit) object definition for more information.
+
+### Create an executable file
+
+Then, you need to create an executable file that will execute the plugin returned from the above call:
+
+```js
+// bin/protoc-gen-foo
+
+#!/usr/bin/env node
```
timostamm, line 88: The hashbang (#!) needs to be the first line in the file. The comment might throw users off. Can you instead mention in the text that this file should to be put to `bin/protoc-gen-foo` _without_ a file extension ?

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+========================
+
+## Introduction
+
+This package helps to create your own code generator plugin.  At a high level, the process is:
+
+1.  Implement a TypeScript generator function, `generateTs`, which is used to generate TypeScript files
+2.  Pass this function along with some static configuration values (and optional functions) to `createEcmaScriptPlugin` from the `@bufbuild/protoplugin` package.
+3.  Create an executable file which runs your plugin.
+4.  Specify this in your `package.json` file.
```
timostamm, line 11: This is a great summary, but can we give an introduction to plugins first? For example:

> Code generator plugins are a unique feature of protocol buffer compilers like protoc and the buf CLI. With a plugin, you can generate RPC clients and server stubs, mappings from protobuf to SQL, validation code, and pretty much anything else you can think of.

We can also quickly summarize how plugins work:
> The contract between the protobuf compiler and a code generator plugin is defined in plugin.proto. Plugins are simple executables (typically on your $PATH) that are named `protoc-gen-x`, where `x` is the name of the language or feature that the plugin provides. The protobuf compiler parses the protobuf files, and invokes the plugin, sending a `CodeGeneratorRequest` on standard in, and expecting a `CodeGeneratorResponse` on standard out. The request contains a set of descriptors (see descriptor.proto) - an abstract version of the parsed protobuf files. The response contains a list of files, each having a name and text content.

This lets us explain what @bufbuild/protoplugin adds on top:
- A simplified view of the descriptors that resolves references, hides implementation details like synthetic map entries, and provides simple access to comments.
- Automatic import and export management for ECMAScript.
- First class support for referencing code generated by @bufbuild/protoc-gen-es.
- Automatic transpilation of TypeScript to JavaScript and TypeScript declaration files.

(Not necessarily in the form of this bullet list - it's fine if our documentation makes these points obvious)

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+  
+     for (const enumeration of file.enums) {
+        f.print(`// generating enums from ${file.name}`);
+        ...
+     }
+  
+     for (const message of file.messages) {
+        f.print(`// generating messages from ${file.name}`);
+        ...
+     }
```
timostamm, line 53: Messages can contain other message definitions, and enum definitions. We should call out that this example does not generate all elements.

This is actually something we could support better, by providing a complete list of all messages an a list of all enumerations in a file 🤔

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+}
+
+```
+
+For a complete working example of a plugin written with the framework, check out [protoc-gen-es](https://github.com/bufbuild/protobuf-es/tree/main/packages/protoc-gen-es).
+
+
+## API
+
+Following are the definitions of relevant objects in the plugin framework.
```
timostamm, line 119: I don't think we should try to provide an API doc here (in the sense of a document that aims to list all functions, parameters, etc), because it takes too much effort to maintain without tooling like [jsdoc](https://jsdoc.app/about-getting-started.html#generating-a-website) or [typedoc](https://typedoc.org/).

Instead, I think should use a conversational style, similar to a tutorial, and explain/give examples for common topics in separate sections. For reference, take a look at the [runtime API doc for protobuf-es](https://github.com/bufbuild/protobuf-es/blob/sayers/plugin_docs/docs/runtime_api.md).

For @bufbuild/protoplugin, the list of sections could be:

- Setting up a plugin
  - Why give `version` and `name`? (part of preamble in generated code, shown with `protoc-gen-foo -h`)
- Providing a generator function
- How to walk through the Schema
- Generating a file (naming convention! we want users to add `_foo` for `protoc-gen-foo`)
- Printing to a generated file (at this point, we need to explain how to run the plugin)
- Importing
  - from a npm package
  - from `protoc-gen-es` generated code
- Exporting symbols, and importing them in other generated files
- Parsing plugin options
- Accessing custom protobuf options
- Generating JavaScript or TypeScript declarations yourself (explaining why this can be worth the extra effort)

It would also be great if we could provide answers for the questions:

- How to test my plugin?
- How can I generator non-ECMAScript files?

smaye81: Tried to implement all of this except the last two (testing and non-ECMAScript).  I wasn't entirely sure of the best way to do this yet.  I still need to try out generating a non-ES file.  For testing, I was thinking of pointing users to our protoplugin-test package, but not sure if that helps them.

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+
+For more information on how plugins work, check out [our documentation](https://docs.buf.build/reference/images). 
+
+## Writing a plugin
+
+The following will describe the steps and things to know when writing your own code generator plugin.  The main step in the process is passing a plugin initialization object to the `createEcmaScriptPlugin` function exported by the plugin framework.  This plugin initalization object will contain various properties pertaining to different aspects of your plugin.
```
timostamm, line 37: Would be helpful to call out how to find the plugin framework with a link to the package on npmjs.com.

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+The natural instinct would be to simply print your own import statements as `f.print("import { Foo } from 'bar'")`, but this is not the recommended approach.  Using `f.import()` has many advantages such as:
+
+- **Conditional imports**
+   - Import statements belong at the top of a file, but you usually only find out later whether you need the import, such as further in your code in a nested if statement.  Conditionally printing the import symbol will only generate the import statement when it is actually used.
+   
+- **Preventing name collisions**
+    - For example if you `import { Foo } from "bar"`  and `import { Foo } from "baz"` , `f.import()` will automatically rename one of them `Foo$1`, preventing name collisions in your import statements and code.
+
+- **Compatibility across standards**
+    - Abstracting the generation of imports allows the library to offer a plugin option for all ECMAScript plugins that generates imports with CommonJS instead of ECMAScript imports, and all plugins work with it out of the box.
```
timostamm, line 284: Can we word this a bit more cautiously? For example:

> ... to potentially offer other import styles in the future ...

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+
+```ts
+function findCustomScalarOption<T extends ScalarType>(
+  desc: AnyDesc,
+  id: number,
+  scalarType: T
+): ScalarValue<T> | undefined;
+```
+
+`AnyDesc` represents any of the `DescXXX` objects such as `DescFile`, `DescEnum`, `DescMessage`, etc.  The `id` parameter represents the ID of the custom options field definition.
```
timostamm, line 345: "the ID" doesn't ring right 🤔  

Can we use "extension numbers" ([ref](https://developers.google.com/protocol-buffers/docs/proto#choosing)) ?

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+
+```ts
+export function findCustomMessageOption<T extends Message<T>>(
+  desc: AnyDesc,
+  id: number,
+  msgType: MessageType<T>
+): T | undefined {
+```
+
+The `msgType` parameter represents the type of the message you are searching for.  Note that this can be a message generated from a proto file or can be a type created at runtime via `makeMessageType`.
```
timostamm, line 386: Do we need to call out that you can also pass a type created at run time? 

Our recommendation should be to generate code for this message, because it is much more convenient.

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+ *     many: ["a", "b", "c"],
+ *     mapping: [{key: "testKey", value: "testVal"}]
+ * }
+ */
+ ```
+ 
+ Alternatively, you can use a type created at runtime:
+
+```ts
+const opts = proto3.makeMessageType("ServiceOptions", [
```
timostamm, line 448: Same here. 

There _are_ use cases for creating a type at run time, but not hard coded like this, but via `createRegistryFromDescriptors()`. This would allow you to parse custom options and - for example - turn them into JSON without knowing their concrete type. Not sure we need to go into those details in this doc?

docs/writing_plugins.md
```diff
@@ -0,0 +1,550 @@
...
+  option (foo_enum_option) = ACTIVE;
+
+  string name = 1;
+}
+```
+
+The value of this option can be retrieved as follows:
+
+```ts
+const enumVal = findCustomEnumOption(descMessage, 50001);  // 1
```
timostamm, line 536: If we recommend to generate enumerations used in custom options, like we do for messages, this could be:

```ts
const enumVal: ServiceStatus | undefined = findCustomEnumOption(service, 50006);
```

What do you think?

timostamm (commented):
This is really great, just a couple of comments below.

Also, can you add a short sentence above the ToC, similar to the [other](https://github.com/bufbuild/protobuf-es/blob/sayers/plugin_docs/docs/runtime_api.md) [docs](https://github.com/bufbuild/protobuf-es/blob/sayers/plugin_docs/docs/generated_code.md)?

Those other docs also prefix the headline with "Protobuf-ES". Can you do that here as well, or alternatively remove it from the other docs?

timostamm (approved):
This looks great to me.

PR #249, opened 2022-10-10
Scope out "traditional" approach for proto2 extensions

timostamm:
This is a starting for full support for proto2 extensions, generating a distinct symbol for each extension field.

Note that it is a very incomplete draft that only investigates the API shape. This may never land.

In contrast to https://github.com/bufbuild/protobuf-es/pull/231, this approach works well for consuming custom options in custom code generators, but requires an extension registry that needs to be populated by the user, and is likely to increase bundle size for users who do not need the feature.


timostamm:
proto2 extensions can be very useful, but we are not going to support the feature for v1. Closing this PR, but we are going to keep the branch.

PR #253, opened 2022-10-18
Fix TypeScript file generation bug

smaye81:
There are conditions where TypeScript files must be generated even if `ts` was not specified in the target out.  This is because the TypeScript files are needed for the transpilation process.

In these circumstances, the plugin should not generate actual `.ts` files.  This removes them from the `CodeGeneratorResponse`.


packages/protoplugin/src/ecmascript/schema.ts
```diff
@@ -136,10 +136,6 @@ export function createSchema(
   return {
     schema,
     getFileInfo() {
-      // TODO - If all targets are requested, but the plugin doesn't have a generateDts, we might need to
```
smaye81, line 139: Removing this comment as this has already been implemented.

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -163,8 +163,16 @@ export function createEcmaScriptPlugin(init: PluginInit): Plugin {
         }
       }
 
-      // Get all generated files
-      const files = getFileInfo();
+      // Get generated files.  If ts was specified as a target, then we want all generated files.  If ts was not
+      // specified, we still may have generated TypeScript files to assist in transpilation.  If they were
+      // generated but not specified in the target out, we shouldn't produce these files
+      // in the CodeGeneratorResponse.
```
timostamm, line 169: Very nitty, but can you wrap the comment at 80 columns?

packages/protoplugin/src/ecmascript/transpile.ts
```diff
@@ -146,5 +152,10 @@ export function transpile(
   if (err) {
     throw err;
   }
+  if (result.emitSkipped) {
+    throw Error(
+      "An problem occurred during transpilation and files were not generated.  Contact the plugin author for support."
+    );
+  }
```
timostamm, line 159: 👍

timostamm (approved):
LGTM, pending comment wrapping.

PR #254, opened 2022-10-20
fix: arbitrary imports

fubhy:
Simple fix of a typo that currently breaks code generation with arbitrary imports.


timostamm (approved):
Hah, that's a good one :laughing:

Thank you for the fix, Sebastian!

fubhy:
Oh wow that was fast. Thanks

PR #255, opened 2022-10-20
Export printable type

fubhy:


timostamm:
I felt tempted to make it public too. The rationale not to do so was to encourage keeping plugin code simple, and pass an `f: GeneratedFile` to helper functions. Can I ask about your use case?

fubhy:
I'm writing a coupler helper functions in my generator plugin for which it would be great to access this type (to type input & output in a composable manner). 

It's not a big deal if you don't expose this type because I can also derive / infer it from `f.print` parameters but it would be a bit less "hacky" to be able to access it directly.

fubhy:
I see your point though about passing around the file context everywhere. I'll have to think about that ...

It makes unit testing a bit less ergonomic tbh.

fubhy:
Exactly. I need to generate a potentially long line (schema validation), sometimes also pull in an import, etc. so I was trying to split up the logic for that a bit without adding any newlines.

timostamm:
Gotcha. Let's make it public!

Can I ask you to use the new export in [gencommon.ts](https://github.com/bufbuild/protobuf-es/blob/5490fef984969ea66e8ec6ced41b33097691b14a/packages/protoplugin/src/ecmascript/gencommon.ts#L33)?

fubhy:
Yep. Also found some other places. I guess you would want to use it there too.

timostamm (approved):
LGTM!

PR #263, opened 2022-10-31
Expose helper WKT methods

smaye81:


timostamm (approved):
I think we can integrate this better in the long run, but let's do it incrementally. LGTM!

PR #274, opened 2022-11-01
Update dependencies

seankimdev:
```
npm install --prefix packages/protobuf-bench esbuild@latest
npm install --prefix packages/protoplugin-example tsx@latest
npm install eslint@latest
npm install @typescript-eslint/eslint-plugin@latest
npm install --save-dev @typescript-eslint/parser@latest
npm install --save-dev eslint-import-resolver-typescript@latest
npm install jest@latest
npm install @types/jest@latest
```


packages/protobuf-bench/README.md
```diff
@@ -11,4 +11,4 @@ server would usually do.
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
 | protobuf-es         | 74,055 b      | 36,960 b | 9,651 b |
-| protobuf-javascript | 370,857 b  | 271,536 b | 43,759 b |
```
smaye81, line 14: Decent jump here.  v3.21.2 must have a bit more code than 3.20.1

smaye81 (approved):

packages/protobuf-bench/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "0.2.1",
-    "esbuild": "^0.15.10",
-    "google-protobuf": "^3.20.1"
+    "esbuild": "^0.15.12",
+    "google-protobuf": "^3.21.2"
```
timostamm, line 11: If we update the package google-protobuf to the latest version, we should also update the code generator to the latest version. The generated code depends on this package. 

The generator is pinned [here](https://github.com/bufbuild/protobuf-es/blob/33ce05a2fbe1351a271070228f3f4c447cbb5162/packages/protobuf-bench/buf.gen.yaml#L11). Can you update this?

seankimdev: Thanks Timo and Steve. How do you update it? curious

smaye81: You just have to change the value in that `buf.gen.yaml` that Timo linked.  However, there is no plugin with that version published in the BSR yet.  Once something is published there, I'll post an update.  Until then, since I'm not sure when that will be, let's just remove the update to `protobuf` in this PR and we'll update both in a separate PR.

seankimdev: Sounds good. Reverted the protobuf version update.

packages/protobuf-bench/package-lock.json
```diff
@@ -0,0 +1,564 @@
+{
+  "name": "@bufbuild/protobuf-bench",
```
timostamm, line 2: This lock file should not exists. In a project with npm workspaces, there must only be a single lock file.

seankimdev: I see. That makes sense. Thank you. Removed both lock files in protobuf-bench and protobuf-example.

smaye81 (approved):

packages/protobuf-bench/README.md
```diff
@@ -11,4 +11,4 @@ server would usually do.
...
+| protobuf-javascript | 389,691 b  | 288,639 b | 45,153 b |
```
timostamm, line 14: We aren't updating google-protobuf anymore, right? Where does the increase in bundle size come from?

timostamm: `make bench` should fix it.

seankimdev: Oddly no. It seems running `make` after removing those lock files changed them.

seankimdev: Is it possible for `es-build` to affect it?

timostamm: Yes (you can look at the changelog), but very unlikely. 

It is much more likely that this is still using the newer version of google-protobuf. There we multiple lock files at one point, and now we probably have the wrong version in the remaining lock file. Note that the lock file is the authority in the installed version, the package.json file just has version constraints.

seankimdev: Ah yes, you are right. The lock file was using a newer version since the version constraint was set at `^3.20.1`, so it generated the lock file with the latest version `3.21.2`. In this case, would it make sense to pin its version to `3.21.0` to match what it used to be, and generate a new lock file?

packages/protobuf-bench/README.md
```diff
@@ -10,5 +10,5 @@ server would usually do.
 
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
-| protobuf-es         | 74,055 b      | 36,960 b | 9,651 b |
+| protobuf-es         | 74,495 b      | 36,880 b | 9,644 b |
```
timostamm, line 13: I think this is caused by `packages/protobuf-bench/node_modules/@bufbuild/protobuf` in the lock file. It should be `node_modules/@bufbuild/protobuf`.

I guess the import paths make their way into the bundle, increasing the size.

Perhaps @smaye81 can lend you a hand?

smaye81: @timostamm @seankimdev Should be good now.  I basically just checked out the package-lock from main and reran `make`.  Looks like the benchmark values are back to being unchanged.

seankimdev: Thanks Steve!

timostamm (approved):

PR #275, opened 2022-11-02
Make tsNocheck works correctly

yukukotani:
`/* @ts-nocheck */` won't disable type check. `// @ts-nocheck` is the correct syntax.


timostamm (approved):
Thank you for the fix, @yukukotani!

This needs some updates to tests. You can run `make` to see where, but you need [bazel and other tools](https://github.com/bufbuild/protobuf-es/blob/main/.github/CONTRIBUTING.md) installed.

Let's get this merged and I'll follow up with an update to the tests.

yukukotani:
@timostamm Oops, Sorry I didn't notice CONTRIBUTING.md exists! I'll do that myself next time.

PR #277, opened 2022-11-02
allow async code generators

fubhy:
I am writing a code generator for generating tests for another code generator (lol). I want to ingest test cases by reading a file with `readline` using `AsyncIterator` (test cases encoded from https://github.com/bufbuild/protoc-gen-validate/tree/main/tests/harness in jsond format (line by line)).

In general, I think we should allow this because why not? :-)


fubhy:
Note: Tests are failing due to https://github.com/bufbuild/protobuf-es/pull/275

smaye81:
See discussion in https://github.com/bufbuild/protobuf-es/issues/278.  This is not something we are going to implement at this time.

PR #279, opened 2022-11-02
Proposal: Add print function with tagged template literal

yukukotani:
Adds `GeneratedFile#printTag` to improve the human-readability of plugins.

Please note that this PR is just a proposal. Also, the naming of `printTag` is debatable.

## Example

Extracted from [here](https://github.com/bufbuild/protobuf-es/blob/1d27516e6455675e778a807d8865b1bdf8c6d97a/packages/protoplugin-example/src/protoc-gen-twirp-es.ts#L68)

### Before

```ts
f.print("    async ", localName(method), "(request: ", method.input, "): Promise<", method.output, "> {");
```

### After

```ts
f.printTag`    async ${localName(method)}(request: ${method.input}): Promise<${method.output}>{`
```


smaye81:
Hi @yukukotani.  Thanks for the proposal!  This is actually something we've discussed as well.  I think we'd want this as an overload on the `GeneratedFile` type.  This way we can call both functions `print`.  Any interest in refactoring this PR?

yukukotani:
@smaye81 
> I think we'd want this as an overload on the GeneratedFile type. This way we can call both functions print.

Sounds good! I've done the refactoring.

packages/protoplugin/src/ecmascript/generated-file.ts
```diff
@@ -66,7 +66,12 @@ export interface GeneratedFile {
    * - ImportSymbol: Adds an import statement and prints the name of the symbol.
    * - DescMessage or DescEnum: Imports the type if necessary, and prints the name.
    */
-  print(...any: Printable[]): void;
+  print(...printables: Printable[]): void;
+
+  /**
+   * Add a line of code to the file with tagged template literal.
```
smaye81, line 72: `Add a line of code to the file with tagged template literal and an optional array of Printables.  See print(Printable[]) for behavior when printing Printable items.`

smaye81 (changes requested):

packages/protoplugin-test/src/transpile.test.ts
```diff
@@ -185,4 +185,21 @@ describe("transpile", function () {
       "export declare function foo(): Foo;",
     ]);
   });
+
+  test("print with tagged template literal works correctly", () => {
+    const linesOf = transpile((schema) => {
```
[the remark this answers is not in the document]
smaye81: Actually the more I think about it, I'm not entirely convinced we need the `Printable[]` parameter with the template string call.  Let me discuss it with the team and I'll get back to you.  For now, just ignore the mixing template literal and regular string test I mentioned above.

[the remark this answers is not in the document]
smaye81: OK, so after discussion I realized I was incorrect and we actually need that second parameter.  So ignore my second point instead.  Apologies for the confusion.

packages/protoplugin-test/src/tagged-template-literal.test.ts
```diff
@@ -0,0 +1,86 @@
+// Copyright 2021-2022 Buf Technologies, Inc.
```
smaye81, line 1: I would move these tests back into `transpile.test.ts`.  Not sure why, but it looks like the tests are split now between that file and this one.  I'd just create a nested `describe` block called `print with tagged template literal` inside the main `transpile` one.  

Also let's keep the `transpile` helper in the transpile test file since moving the tests back will keep everything contained in one place.

yukukotani: @smaye81 
I split them because `transpile.test.ts` seems a bit long for me. But I'm not very picky about it, so fixed in [7c5fb57](https://github.com/bufbuild/protobuf-es/pull/279/commits/7c5fb57bbc131c5cdcd32faaa0f54424a47d3463).

smaye81:
So apologies for the churn on this.  After discussing it with the team, I think we want to move the generated file logic into a `generated-file.test.ts` for consistency and proper separating of concerns.  The `transpile.test.ts` file should only focus on testing transpilation and in this instance, we don't need transpilation, just generated file content from the `print` function.  

Could you create a new `generated-file.test.ts` test file alongside the others and move all your print testing there instead?  Inside the test file, you can add a new helper that looks as follows:

```ts
/**
 * Generates a single file using the plugin framework and the given print function.
 * The returned string array represents the generated file content created using printFn.
 */
function generateSingleFile(printFn: (f: GeneratedFile) => void): string[] {
  const req = new CodeGeneratorRequest({
    parameter: `target=ts`,
  });
  const plugin = createEcmaScriptPlugin({
    name: "test-plugin",
    version: "v99.0.0",
    generateTs: (schema: Schema) => {
      const f = schema.generateFile("test.ts");
      printFn(f);
    },
  });
  const res = plugin.run(req);
  if (res.file.length !== 1) {
    throw new Error(`no file generated`);
  }

  const content = res.file[0].content ?? "";
  return content.trim().split("\n");
}
```

Then your tests will look something like this:

```ts
  test("print with tagged template literal works correctly", () => {
    const lines = generateSingleFile((f) => {
      const Foo = f.import("Foo", "foo");
      f.print`export function foo(): ${Foo} { return new ${Foo}(); };`;
    });
    expect(lines).toStrictEqual([
      'import {Foo} from "foo";',
      "",
      "export function foo(): Foo { return new Foo(); };",
    ]);
  });
```

packages/protoplugin/src/ecmascript/generated-file.ts
```diff
@@ -66,7 +66,12 @@ export interface GeneratedFile {
    * - ImportSymbol: Adds an import statement and prints the name of the symbol.
    * - DescMessage or DescEnum: Imports the type if necessary, and prints the name.
    */
-  print(...any: Printable[]): void;
+  print(...printables: Printable[]): void;
+
+  /**
+   * Add a line of code to the file with tagged template literal and an optional array of Printables. See print(Printable[]) for behavior when printing Printable items.
```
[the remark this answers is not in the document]
yukukotani: 773d475

yukukotani:
@smaye81 No problem, thanks for the detail suggestion!

packages/protoplugin-test/src/transpile.test.ts
```diff
@@ -185,4 +185,21 @@ describe("transpile", function () {
       "export declare function foo(): Foo;",
     ]);
   });
+
+  test("print with tagged template literal works correctly", () => {
```
smaye81, line 189: Can you move this test to `generated-file.test.ts` also?

yukukotani: @smaye81 Oops it's duplicated with [this test](https://github.com/yukukotani/protobuf-es/blob/773d47585173870442936e08861757875dffc489/packages/protoplugin-test/src/generated-file.test.ts#L45:L45). I've removed it.

PR #281, opened 2022-11-02
show test failure for float bug

fubhy:


smaye81:
This actually should've been caught [here](https://github.com/bufbuild/protobuf-es/blob/main/packages/protoplugin-test/src/custom-options.test.ts#L84) but we use `4.5` as the float value and from what I can tell, the issue is with precisions over 1 decimal place.

PR #284, opened 2022-11-02
fix typo

fubhy:


timostamm (approved):
A typo squashed _and_ 2 bytes saved :) TY!

PR #288, opened 2022-11-03
add unpack function to google.protobuf.Any

fubhy:
Adds a `unpack` convenience function to the `google.protobuf.Any` message type.

Before, in order to consume the underlying message, you'd have to manually do sth. like this:

```typescript
  const any = ... // Some arbitrary `google.protobuf.Any` any.
  const message = registry.findMessage(typeName);
  const instance = new message!();
  any.unpackTo(instance);
```

With this change, that becomes more convenient:

```typescript
  const any = ... // Some arbitrary `google.protobuf.Any` any.
  const message = any.unpack(registry);
```


smaye81:
> @smaye81 Before I add a test case... Is this what you had in mind?

Yep, it is.  Can you also make sure that it handles unpacking an empty `Any`?  Maybe returning `Message | undefined`.  Also please verify that as part of the test coverage.

Also, could you add a description to the PR?  Thanks!

fubhy:
@smaye81 Sure thing. Adding the tests tomorrow. Calling it a night for now :-)

Issue #283, opened 2022-11-02
Too strict type checking during serialization


rideg:
I started migrating my project to protobuf-es from protobuf-javascript.
I received an error message while calling Message.toBinary()

**cannot unwrap field value, package.Message does not define a field wrapper**

It happens because [here](https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf/src/private/field-wrapper.ts#L42). It uses an instanceof check to ensure that the object in the property is really an instance of the filed descriptor type.

However my situation is the following:

I have the proto definitions in a moduleA
I have moduleB that is compiled independently
I have moduleC that depends on moduleB and it receives proto objects from it the wraps them into a proto object instanciated locally, then tries to send them on the wire.

The code is something like this:

```Typescript

import {Wrapper} from "moduleA/wrapper_pb";
import {Service} from "moduleB/server";

function wrap(s: Service): Uint8Array {
   const wrap = new Wrapper();
   const msg = s.getMsg(); // this returns {Msg} from "moduleA/msg_pb"
   wrap.msg = msg;
  return wrap.toBinary(); // this is where I get the error
}
```

The reason for the error is that Msg the msg class has two definitions: one in moduleB and one in moduleC and however they are equivalent the instanceof operator will return false since the prototypes are different.

It probably should be enough to check if the object's type's name is equal to the field descriptor's type name or something.
This use case works fine with the protobuf-javascript.

**Edit:** replaced protobuf-ts with protobuf-javascript

timostamm:
Thanks for the report, Sandor! To be fair, protobuf-ts doesn't have this feature 😄

To better understand how you got into this situation: I assume you generated code for google/protobuf/wrappers.proto?

rideg:
Hey Timo!
Thanks for the quick reply!
Yes you're right. I did not use protobuf-ts, but protobuf-javascript (--js_out + -grpc-web_out params), I was confused. I'll update the issue.

So no I did not generate code for google/protobuf/wrappers.proto. All protos belong to my own project.
They look something like this:

```proto
syntax = "proto3";

package mypackage;

message Msg {
   string field = 1;
}

message Wrapper {
   string some_data = 1;
   Msg msg = 2;
}
```

The way it works in my project is that I create a js module (moduleA) for the protos. Then I have another js module (moduleB) that uses the proto module. Besides these I have another js module (moduleC) that depends on the previous two.
[...]

Issue #230, opened 2022-09-24
Inherit top-level type (e.g. PlainMessage)


faustbrian:
I have some user-facing functions that take `PlainMessage<T>` as input. The problem is that `PlainMessage` doesn't seem to trickle down to properties that themselves are messages. Take the below snippets as an example:

**Protobuf**
```proto
syntax = "proto3";

option optimize_for = SPEED;

message Op {
  string denomination = 1;
}

message Msg {
  repeated Op ops = 1;
}
```

```ts
// Type
type x = PlainMessage<Msg>

// Expected
type x = {
    ops: PlainMessage<Op>[];
}

// Actual
type x = {
    ops: Op[];
}
```

When using `PlainMessage<T>` I would expect all properties of `T` that are also messages to be also referenced as `PlainMessage` rather than expecting the _(full)_ `Message`. Is this an intentional design choice and can this be worked around without a custom type outside the library or is this an oversight? Thanks in advanced!

timostamm:
Hey Brian, thanks for this issue!

The behavior is a design choice - the type is an exact representation for a message cloned with the spread operator: `let plain: PlainMessage<Example> = {...example};`

I'm aware that this is not always helpful in practice. Making `PlainMessage<T>` recursive is definitely an option, but I'm not quite sure yet that it is the best option. So before we consider this, we will be looking into generating an `interface IExample` for every `message Example` alongside the class, which behaves exactly like a recursive `PlainMessage<T>`. 

The rationale for `interface IExample` is that a simple interface can be a bit more approachable and can provide better TypeScript error messages than a mapped type. It also means we would generate a bit more code, so we obviously want to consider this carefully. 

Expect to see some updates regarding this soonish. In the meantime, I don't see any complications from using your own recursive version of `PlainMessage<T>`. If you do, please do let us know about how it works out for you!

timostamm:
A recursive version of `PlainMessage`:

```ts
import type { Message } from "@bufbuild/protobuf";

export type Plain<T extends Message> = {
  [P in keyof T as T[P] extends Function ? never : P]: PlainField<T[P]>;
};

// prettier-ignore
type PlainField<F> =
  F extends (Date | Uint8Array | bigint | boolean | string | number) ? F
  : F extends Array<infer U> ? Array<PlainField<U>>
  : F extends ReadonlyArray<infer U> ? ReadonlyArray<PlainField<U>>
  : F extends Message ? Plain<F>
[...]

PR #289, opened 2022-11-03
Allow serializing identical messages

rideg:
This PR fixes the issues explained in [issue/283](https://github.com/bufbuild/protobuf-es/issues/283) by checking the if the incoming value is a Message<T> and its typeName equals to the field descriptor's typeName.


packages/protobuf/src/message.ts
```diff
@@ -132,6 +132,25 @@ export class Message<T extends Message<T> = AnyMessage> {
...
+    typeof value.clone === "function" &&
+    typeof value.fromBinary === "function" &&
+    typeof value.fromJson === "function" &&
+    typeof value.fromJsonString === "function" &&
+    typeof value.toBinary === "function" &&
+    typeof value.toJson === "function" &&
+    typeof value.toJsonString === "function" &&
+    typeof value.getType === "function"
+  );
+}
```
fubhy, line 152: Hmm... I don't think this is the ideal approach to this. I'd suggest to instead implement a generic ttype guard on the MessageType instead and, in there, only loosely check that the given Mesasge has a getType function and that it returns the same message type (typeName comparison)

```typescript
is(message: unknown): message is Message<ThisMessageType> {
  // Check that `message` has a `getType` function and that that returns
  // a `MessageType` with `typeName === this.typeName`
}
```

packages/protobuf/src/private/field-wrapper.ts
```diff
@@ -45,6 +45,11 @@ export function wrapField<T extends Message<T>>(
   if (type.fieldWrapper) {
     return type.fieldWrapper.wrapField(value);
   }
+  // Last resort check if the value is indeed a Message and has the same
+  // type as the field.
+  if (isMessage<T>(value) && value.getType().typeName === type.typeName) {
+    return value;
+  }
```
fubhy, line 52: ... So here, you'd instead do:

```typescript
if (type.is(value)) {
  return value;
}
```

Note: Not sure if the name `is` is a good choice, but you get the idea.

I am not a maintainer though... So :-)

fubhy (changes requested):

fubhy:
I just ran into this too while trying to stub out a slim Message object in a test.

Generally I agree that `instanceof` should be used carefully, especially in areas of a library that are part of the public interface. But I also wouldn't go down the path of adding too aggressive runtime duck type checking as in that `isMessage` function.

I think it's good enough to check whether there's a `getType` function and whether it returns an object with a `typeName` property that matches!

rideg:
@fubhy Thanks for checking my PR. Great suggestions! I changed accordingly. I came with "conforms" as a method name. However I need some help because I had to change the generator code as well, but I don't know how to regenerate the well known protobuf files and now everything is broken due to the missing methods there.

fubhy:
> @fubhy Thanks for checking my PR. Great suggestions! I changed accordingly. I came with "conforms" as a method name. However I need some help because I had to change the generator code as well, but I don't know how to regenerate the well known protobuf files and now everything is broken due to the missing methods there.

Yeah that's a bit of a chicken-and-egg problem... The `make` routine depends on the build output of `tsc` to run the protoc generator. In turn, `protoc-gen-es` depends on `@bufbuild/protobuf` which hosts some of the code that's generated by `protoc-gen-es`. So once that package contains code that doesn't pass the typecheck, you get stuck ... Sadly, `tsc` doesn't have a `--transpileOnly` option (ignore type checking during build).

Not sure how the buf.build team handles that locally but I simply reverted all local changes to `./packages/protobuf` and then run `make` again.

@timostamm @smaye81 I think it would be good to configure `tsconfig.json` for development to have path aliases set up to the `src` of each package in the monorepo and then use a `ts-node` shebang for something like `./bin/protoc-gen-es.ts` with `ts-node` configured to `transpileOnly`. Just for local development...

I've configured my plugin repo similarly FYI.

packages/protobuf/src/private/util.ts
```diff
@@ -66,6 +66,12 @@ export interface Util {
     b: T | PlainMessage<T> | undefined | null
   ): boolean;
 
+  /**
+   * Checks if the given value conformst the given type.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  conforms<T extends Message<T>>(type: MessageType<T>, value: any): value is T;
```
fubhy, line 73: I'd use `unknown` here instead of `any`. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

packages/protoc-gen-es/src/declaration.ts
```diff
@@ -101,6 +101,8 @@ function generateMessage(schema: Schema, f: GeneratedFile, message: DescMessage)
   f.print()
   f.print("  static fromJsonString(jsonString: string, options?: Partial<", JsonReadOptions, ">): ", message, ";")
   f.print()
+  f.print("  static conforms(value: any): value is ", message, ";")
```
fubhy, line 104: Same here (unknown instead of any)

packages/protoc-gen-es/src/typescript.ts
```diff
@@ -124,6 +124,10 @@ function generateMessage(schema: Schema, f: GeneratedFile, message: DescMessage)
   f.print("    return new ", message, "().fromJsonString(jsonString, options);")
   f.print("  }")
   f.print()
+  f.print("  static conforms(value: any): value is ", message, " {")
```
fubhy, line 127: and here :)

fubhy (changes requested):

rideg:
Ok, it seems I fixed all the issues. All tests pass and no there are no lint errors or warnings. The main changes are in the following files:

In ```protoc-gen-es```:
* packages/protoc-gen-es/src/declaration.ts
* packages/protoc-gen-es/src/javascript.ts
* packages/protoc-gen-es/src/typescript.ts

In  ```@bufbuild/protobuf```:
* packages/protobuf/src/message-type.ts
* packages/protobuf/src/private/field-wrapper.ts
* packages/protobuf/src/private/message-type.ts
* packages/protobuf/src/private/util-common.ts
* packages/protobuf/src/private/util.ts

In ```protobuf-test```:
* packages/protobuf-test/src/tobinary.test.ts

In the ```docs```:
* docs/runtime_api.md

I think during the next release the minor version might need to be bumped due to the API change, but I guess that is not part of this PR.

packages/protobuf/src/private/util.ts
rideg: Good point! Updated accordingly!

packages/protoc-gen-es/src/declaration.ts
rideg: done

packages/protoc-gen-es/src/typescript.ts
rideg: done

rideg:
I force re pushed all commits because I had the wrong git user.name and user.email set.

timostamm:
Sandor, thank you for looking into this! 

In the issue you put up, we take the wrong code path if `instanceof` fails, which causes a crash. This is very specific to our "unboxing" of the well-known wrapper types ([docs](https://github.com/bufbuild/protobuf-es/blob/main/docs/generated_code.md#message-fields)). We should be more tolerant there for sure, this was an oversight.

In other cases where `instanceof` might fail, we fall back on a code path that simply creates a new instance. This is a pretty robust way to handle duplicate generated code.

The best way forward here is to fix the code paths that need to be more robust, and _not_ add a type guard on `MessageType`.

fubhy:
@timostamm Are you against adding such a type guard on the generated code altogether or just for the sake of solving the original issue? I'd definitely see use for this type guard (see example below): 

```typescript
const conforming = new Example();
const nonConforming = new OtherType();
Example.conforms(conforming); // true
Example.conforms(nonConforming); // false
```

@rideg Apologies for leading you down the wrong path here!

timostamm:
I'm against a type guard in the generated code because it will only help in a very narrow band of use cases, where you have a class, but it might not have the identity you expect. For cases like this, it is also possible to override `instanceof`, for example:

```ts
Object.defineProperty(SomeMessage, Symbol.hasInstance, {
  value(instance: unknown) {
    return instance instanceof Message && instance.getType().typeName == SomeMessage.typeName;
  },
});
```

To be truly useful, fields must also be compared, otherwise it would give a false positive for a message with a removed field, for example. So far, I don't believe we need this complexity though. We can simply make our code paths a bit smarter to solve the original issue.

Apologies to both of you for not chiming in earlier.

timostamm:
There is another kind of type guard that would work well in combination with the recursive `Plain<T>` or message interfaces as described in https://github.com/bufbuild/protobuf-es/issues/230#issuecomment-1257544673: 

```ts
function isMessage<T extends Message<T>>(arg: unknown, type: MessageType<T>): arg is Plain<T>;
```

This could be a welcome addition for folks who want to deconstruct and construct messages without caring about classes. But this is also something best added as a standalone function instead of a method on `MessageType`, if only to keep it tree-shakeable.

By the way, protobuf-**ts** has them - see the [manual](https://github.com/timostamm/protobuf-ts/blob/master/MANUAL.md#message-type-guards). They were added because it only generates interfaces, not classes, and there is no equivalent to `instanceof`. They are also quite expensive to run, and can't distinguish between a `google.protobuf.Duration` and a `google.protobuf.Timestamp`.

fubhy:
Ok I see where you are coming from. Thanks for clarifying!

rideg:
Sure, thanks for the responses, reviews. I am going to close this PR as it is not necessary anymore.

PR #291, opened 2022-11-07
Make wrappers more lenient

### LINKED ISSUE #283: Too strict type checking during serialization

timostamm:
Fixes #283 by removing code path that errors. We can easily and reliably determine whether a field is already wrapped by checking if it is `instanceof Message`.

### DIFF
--- a/packages/protobuf/src/google/protobuf/wrappers_pb.ts
+++ b/packages/protobuf/src/google/protobuf/wrappers_pb.ts
@@ -79,8 +79,8 @@ export class DoubleValue extends Message<DoubleValue> {
   ]);
 
   static readonly fieldWrapper = {
-    wrapField(value: number | DoubleValue): DoubleValue {
-      return value instanceof DoubleValue ? value : new DoubleValue({value});
+    wrapField(value: number): DoubleValue {
+      return new DoubleValue({value});
     },
     unwrapField(value: DoubleValue): number {
       return value.value;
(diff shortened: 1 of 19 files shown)


smaye81 (approved):

packages/protobuf/src/google/protobuf/wrappers_pb.ts
```diff
@@ -79,8 +79,8 @@ export class DoubleValue extends Message<DoubleValue> {
   ]);
 
   static readonly fieldWrapper = {
-    wrapField(value: number | DoubleValue): DoubleValue {
-      return value instanceof DoubleValue ? value : new DoubleValue({value});
+    wrapField(value: number): DoubleValue {
+      return new DoubleValue({value});
     },
```
timostamm, line 84: This simplifies the generated code by making the internal `function wrapField` idempotent instead of this FieldWrapper. I don't like changing this, but it the generated code does look quite a bit nicer...

PR #295, opened 2022-11-10
Updates docs for print in plugins

smaye81:
https://github.com/bufbuild/protobuf-es/pull/279 adds an overload to `print` for using a template literal tagged function.  This updates the docs to illustrate that.


timostamm (approved):

docs/writing_plugins.md
```diff
@@ -166,22 +168,47 @@ for (const file of schema.files) {
...
+
+You can also pass a template literal to the function and use string interpolation as you would do in regular JavaScript:
+
+For example:
 
+```ts
+const name = "UserService";
+f.print`export class ${name}Client {     
+    console.log('Hello world');
+}`;
```
timostamm, line 204: Ah, hold on. Should we really recommend multiline strings? This code will always be in a function body: 

```ts
  const name = "UserService";
  f.print`export class ${name}Client {     
      console.log('Hello world');
  }`;
```

Which means it will generate:

```ts
export class UserServiceClient {
      console.log('Hello world');
  }
```

(extra leading whitespace on line 2 and 3)

smaye81: Gotcha.  Updated.

timostamm (approved):

PR #296, opened 2022-11-10
Improve TS compatibility checks

smaye81:
This makes the TS compatibility checks a bit more robust by fixing a few issues in the codebase as well as changing our TypeScript test configs to include all files rather than just the test files.  

Note that the `long` package needed downgraded to `5.0.1` (and pinned to patches only) because of an issue in that library with 5.2 and above.  See this issue:  https://github.com/dcodeIO/long.js/issues/109


timostamm (approved):

packages/protobuf-test/typescript/tsconfig.4_1_2.json
```diff
@@ -1,6 +1,16 @@
...
-  // These are the default compiler options for TypeScript v4.1.2, created with `tsc --init`
+  "include": ["../src/**/*"],
+  "exclude": [
+    // Exclude WKTs and their usage in tests because they use syntax not
+    // compatible with older versions of TypeScript.  WKTs used by actual
+    // consumers of this library are generated with ts-nocheck so its safe
+    // to ignore them here and still get version compatibility assurance.
+    "../src/gen/ts/google/protobuf/*.ts",
+    "../src/google/protobuf/*.test.ts",
+    "../src/message.test.ts"
```
timostamm, line 10: We should be able to include message.test.ts, if we change the test to use `Timestamp` from `@bufbuild/protobuf` instead of the locally generated file.

packages/protobuf-test/typescript/tsconfig.4_2_4.json
```diff
@@ -1,6 +1,16 @@
...
-  // These are the default compiler options for TypeScript v4.2.4, created with `tsc --init`
+  "include": ["../src/**/*"],
+  "exclude": [
+    // Exclude WKTs and their usage in tests because they use syntax not
+    // compatible with older versions of TypeScript.  WKTs used by actual
+    // consumers of this library are generated with ts-nocheck so its safe
+    // to ignore them here and still get version compatibility assurance.
+    "../src/gen/ts/google/protobuf/*.ts",
+    "../src/google/protobuf/*.test.ts",
+    "../src/message.test.ts"
```
timostamm, line 10: Same here, should be able to include message.test.ts.

packages/protobuf-test/typescript/tsconfig.4_1_2.json
```diff
@@ -1,6 +1,16 @@
 {
-  "include": ["../src/gen/js/extra/*.js", "../src/gen/ts/extra/*.ts"],
-  // These are the default compiler options for TypeScript v4.1.2, created with `tsc --init`
+  "include": ["../src/**/*"],
+  "exclude": [
+    // Exclude WKTs and their usage in tests because they use syntax not
+    // compatible with older versions of TypeScript.  WKTs used by actual
+    // consumers of this library are generated with ts-nocheck so its safe
+    // to ignore them here and still get version compatibility assurance.
```
timostamm, line 7: I think our ts-nocheck would only help the other way round - if we were missing an `override`. Perhaps reword like so?:

```suggestion
    // Exclude WKTs and their usage in tests because they use syntax not
    // compatible with older versions of TypeScript.  WKTs are typically not
    // generated, but imported from @bufbuild/protobuf, which is compatible.
    // Users of this version of TypeScript must use them.
```

packages/protobuf-test/typescript/tsconfig.4_2_4.json
```diff
@@ -1,6 +1,16 @@
 {
-  "include": ["../src/gen/js/extra/*.js", "../src/gen/ts/extra/*.ts"],
-  // These are the default compiler options for TypeScript v4.2.4, created with `tsc --init`
+  "include": ["../src/**/*"],
+  "exclude": [
+    // Exclude WKTs and their usage in tests because they use syntax not
+    // compatible with older versions of TypeScript.  WKTs used by actual
```
timostamm, line 5: Same as above.

timostamm (commented):
Thanks for adding more files to the compilation! Two nits though:

PR #297, opened 2022-11-10
makes generated imports look more normal

dimitropoulos:
The rationale for this change is that we should take the opportunity, when it arises, to make generated code look natural and typical (if possible).  While this isn't a logical change, it's hopefully worthwhile in the pursuit of making the generated outputs feel more normal.

Here's an example:

```diff
- import {Message, proto3} from "@bufbuild/protobuf";
+ import { Message, proto3 } from "@bufbuild/protobuf";
```

Note that this falls in line with [Prettier's default](https://prettier.io/docs/en/options.html#bracket-spacing) as well.  In fact, that's how I found it. I was generating code and trying to aim for a final output that matched prettier's default formatting and this was the only thing that I wasn't able to make match as a downstream consumer of protobuf-es.

### DIFF
--- a/packages/protoplugin/src/ecmascript/generated-file.ts
+++ b/packages/protoplugin/src/ecmascript/generated-file.ts
@@ -190,7 +190,7 @@ function elToContent(el: El[], importerPath: string): string {
       const p = names.map(({ name, alias }) =>
         alias == undefined ? name : `${name} as ${alias}`
       );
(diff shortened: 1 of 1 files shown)


dimitropoulos:
closing in favor of https://github.com/bufbuild/protobuf-es/pull/298 so I can switch from the remote to here

PR #298, opened 2022-11-10
makes generated imports look more normal

dimitropoulos:
The rationale for this change is that we should take the opportunity, when it arises, to make generated code look natural and typical (if possible). While this isn't a logical change, it's hopefully worthwhile in the pursuit of making the generated outputs feel more normal.

Here's an example:

```
- import {Message, proto3} from "@bufbuild/protobuf";
+ import { Message, proto3 } from "@bufbuild/protobuf";
```

Note that this falls in line with [Prettier's default](https://prettier.io/docs/en/options.html#bracket-spacing) as well. In fact, that's how I found it. I was generating code and trying to aim for a final output that matched prettier's default formatting and this was the only thing that I wasn't able to make match as a downstream consumer of protobuf-es.

### DIFF
--- a/packages/protobuf-bench/report.mjs
+++ b/packages/protobuf-bench/report.mjs
@@ -1,5 +1,5 @@
-import {buildSync} from "esbuild";
-import {brotliCompressSync} from "zlib";
+import { buildSync } from "esbuild";
+import { brotliCompressSync } from "zlib";
 
 const protobufEs = gather("src/entry-protobuf-es.ts");
 const googleProtobuf = gather("src/entry-google-protobuf.js");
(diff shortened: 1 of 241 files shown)


packages/protobuf-bench/report.mjs
```diff
@@ -20,38 +20,37 @@ server would usually do.
...
-    const bundleMinified = build(entryPoint, true, "esm");
-    const compressed = compress(bundleMinified);
-    return {
-        entryPoint,
-        size: formatSize(bundle.byteLength),
-        minified: formatSize(bundleMinified.byteLength),
-        compressed: formatSize(compressed.byteLength),
-    };
+  const bundle = build(entryPoint, false, "esm");
+  const bundleMinified = build(entryPoint, true, "esm");
```
timostamm, line 25: Oops. Can I ask you to add `mjs` [here](https://github.com/bufbuild/protobuf-es/blob/2657ac74842880a3871de76a708f791d73058388/Makefile#L196)?

timostamm (commented):
Thanks for the diligent update, including the docs ❤️

Looks like we missed the single mjs file in the formatter setup. Do you mind fixing it while you're at it?

packages/protobuf-bench/report.mjs
dimitropoulos: sure! [`2035624` (#298)](https://github.com/bufbuild/protobuf-es/pull/298/commits/2035624b20cdfc35b4b2becb09872d528378a44f)

timostamm (approved):

PR #301, opened 2022-11-14
[WIP]: Extension support

srikrsna-buf:
WIP extensions support. Experimental. This assumes that field names will not conflict.

### DIFF
--- a/packages/protobuf-test/extra/proto2-extend.proto
+++ b/packages/protobuf-test/extra/proto2-extend.proto
@@ -1,0 +1,33 @@
+// Copyright 2021-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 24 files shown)


timostamm (commented):
Thanks for the initiative, @srikrsna-buf. 

Keeping track of extensions on the message type is a nice approach. But we can't ask users to maintain a registry for Any _and_ extension sets. I think we might be able to auto-register them, but we have to avoid imports for side-effects, and we have to be cautious with circular imports. 

One major benefit of the [simple approach](https://github.com/bufbuild/protobuf-es/pull/231) is that it doesn't add the the bundle size at all if you don't use extensions. I think we might be able to pull this off with this approach as well (by moving the code into the extension type, and only sharing an interface with the binary/json format).

I think we have to investigate this route (the "traditional" way) further. Thanks for putting up the idea with the extension set.

timostamm:
proto2 extensions can be very useful, but we are not going to support the feature for v1. Closing this PR, but we are going to keep the branch.

PR #303, opened 2022-11-15
Add an unpack function to google.protobuf.Any

smaye81:
This incorporates work done in https://github.com/bufbuild/protobuf-es/pull/288.  The extra additions were:

- Adding unit tests
- Adding generation code for JavaScript and declaration files.

### DIFF
--- a/packages/protobuf/src/google/protobuf/any_pb.ts
+++ b/packages/protobuf/src/google/protobuf/any_pb.ts
@@ -16,13 +16,14 @@
 // @generated from file google/protobuf/any.proto (package google.protobuf, syntax proto3)
 /* eslint-disable */
 
-import type { PartialMessage, PlainMessage } from "../../message.js";
-import { Message } from "../../message.js";
-import { proto3 } from "../../proto3.js";
-import type { JsonReadOptions, JsonValue, JsonWriteOptions } from "../../json-format.js";
-import type { MessageType } from "../../message-type.js";
-import type { FieldList } from "../../field-list.js";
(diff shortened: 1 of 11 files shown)


packages/protobuf/src/google/protobuf/any_pb.ts
```diff
@@ -216,6 +217,17 @@ export class Any extends Message<Any> {
...
   }
 
+  unpack(registry: IMessageTypeRegistry): Message | undefined {
+    if (this.typeUrl === "") {
+      return undefined;
+    }
+    const messageType = registry.findMessage(this.typeUrlToName(this.typeUrl));
+    if (!messageType) {
+      throw new Error(`cannot unpack message: ${this.typeUrl} is not in the type registry`);
+    }
```
timostamm, line 227: Is it the right call to raise an error if the message wasn't found?

Users will already have to handle the `undefined` return value already for the case that the Any is empty. I have the gut feeling that it's best to only have users handle one error path, and return `undefined` if the message type is not in the registry.

smaye81: I had the same thought actually.  I figured it was keeping consistent with the `is` implementation, which returns false on an empty `Any`, but can throw an error with a bad type URL (see `typeUrlToName`).  Should we also change that or does an error in that instance make more sense?

smaye81: Sounds good.  Updated!

timostamm (approved):

PR #305, opened 2022-11-16
fix map equality

### LINKED ISSUE #304: map equality is incorrect
The current implemention of `equals` for maps does not detect new keys: https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf/src/private/util-common.ts#L166

```
          case "map":
            const keys = Object.keys(va);
            if (keys.some((k) => vb[k] === undefined)) {
              return false;
            }
            switch (m.V.kind) {
              case "message":
                const messageType = m.V.T;
                return keys.every((k) => messageType.equals(va[k], vb[k]));
              case "enum":
                return keys.every((k) =>
                  scalarEquals(ScalarType.INT32, va[k], vb[k])
                );
              case "scalar":
                const scalarType = m.V.T;
                return keys.every((k) =>
                  scalarEquals(scalarType, va[k], vb[k])
                );
            }
            break;
```

These are considered equivalent:

```
  console.log(
    new MyMessage({ myMap: { a: 'b' } }).equals(
      new MyMessage({ myMap: { a: 'b', c: 'd' } })
    )
  );
```

This is because only the keys from `va` are tested, whereas all the keys from both `va` and `vb` need to be tested:

```
union(Object.keys(va), Object.keys(vb))
```

calebdoxsey:
Fix map equality when keys are added to the second map.

Fixes #304

### DIFF
--- a/packages/protobuf/src/private/util-common.ts
+++ b/packages/protobuf/src/private/util-common.ts
@@ -164,7 +164,10 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
             throw new Error(`oneof cannot contain ${s.kind}`);
           case "map":
             const keys = Object.keys(va);
-            if (keys.some((k) => vb[k] === undefined)) {
+            if (
+              keys.some((k) => vb[k] === undefined) ||
+              Object.keys(vb).some((k) => va[k] === undefined)
+            ) {
               return false;
             }
             switch (m.V.kind) {
--- a/packages/protobuf-test/src/equals.test.ts
+++ b/packages/protobuf-test/src/equals.test.ts
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { MapsMessage as TS_MapsMessage } from "./gen/ts/extra/msg-maps_pb.js";
+import { MapsMessage as JS_MapsMessage } from "./gen/js/extra/msg-maps_pb.js";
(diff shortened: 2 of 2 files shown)


packages/protobuf/src/private/util-common.ts
```diff
@@ -164,7 +164,10 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
             throw new Error(`oneof cannot contain ${s.kind}`);
           case "map":
             const keys = Object.keys(va);
-            if (keys.some((k) => vb[k] === undefined)) {
+            if (
+              keys.some((k) => vb[k] === undefined) ||
+              Object.keys(vb).some((k) => va[k] === undefined)
+            ) {
               return false;
             }
```
timostamm, line 172: Thinking about this a bit, I'm not sure we gain much from checking for `undefined` in the first place. We can just as well rely on `scalarEquals()` and `messageType.equals()`, which already handle `undefined`.

```diff
-           const keys = Object.keys(va);
-           if (
-            keys.some((k) => vb[k] === undefined) ||
-             Object.keys(vb).some((k) => va[k] === undefined)
-           ) {
-             return false;
-           }
+           const keys = Object.keys(va).concat(Object.keys(vb));
```

Do you mind updating this?

timostamm (commented):
Thank you for spotting this, Caleb 🙏 

The fix looks correct to me, but I believe it can be simplified further, see comment below.

calebdoxsey:
@timostamm thanks. I simplified the code and added a few more tests.

timostamm (approved):
LGTM!

PR #308, opened 2022-11-21
Make PlainMessage recursive

smaye81:
This adds recursiveness to `PlainMessage` so that it's functionality on fields is applied to nested message fields also.


packages/protobuf-test/src/message.test.ts
```diff
@@ -12,23 +12,57 @@
...
     expect(plainTimestamp.toBinary).toBeDefined();
-  });
-  test("removes wkt methods from type system", () => {
     // Custom methods of well-known types are removed as well.
     // @ts-expect-error TS2339
     expect(plainTimestamp.toDate).toBeDefined();
   });
 });
+
+describe("PartialMessage", () => {
```
timostamm, line 37: constructor.test.ts already tests most of this, because constructors take a partial message as input.

Perhaps it makes sense to move the additional tests for PartialMessage there? The message being tested (protobuf_test_messages.proto3.TestAllTypesProto3) also has fields for all WKT.

packages/protobuf-test/extra/example.proto
```diff
@@ -15,8 +15,11 @@
 syntax = "proto3";
 package docs;
 
+import "google/protobuf/timestamp.proto";
+
 message Example {
   string foo = 1;
   bool bar = 2;
   Example baz = 3;
+  google.protobuf.Timestamp created = 4;
```
timostamm, line 24: I think we may end up putting all possible field types in here over time. We might just as well use protobuf_test_messages.proto3.TestAllTypesProto3 right away, and keep this as a simple example.

packages/protobuf-test/src/message.test.ts
smaye81: I switched to use `TestAllTypesProto3`, but if it's cool with you, I think I'd like to leave the tests here.  The `constructor.test` stuff is so constructor-specific and I like having this separate file to focus on the removal of standard/wkt methods and the recursiveness of that functionality.

timostamm (commented):
I don't think it is a good idea to add a `describe("PartialMessage")`, while we still cover the majority of what the type does in constructor.test.ts. Creating instances and testing for existence of properties at run time also seems distracting when we want to test a type. This makes our coverage muddy.
I would approve a `describe("PartialMessage")` if it was our single group of tests for this type, and avoided the problems described.

For PlainMessage, we have previously introduced a test to guard against a regression we once had, and this adds coverage for a change we are making. This is practical, but creating instances and testing for existence of properties seems distracting here as well. The use of describeMT() is also distracting, because we create a type dynamically from a descriptor set, which is not necessary at all, and because we now operate on a union type of the .ts and .d.ts types.

Instead, why don't we test by assigning an object literal to a PlainMessage? This provides pretty decent coverage for the mapped type, and also proves that the methods are removed:

```ts
test("PlainMessage", () => {
  const t: PlainMessage<TestAllTypesProto3> = {
    optionalInt32: 0,
    // ...
    optionalBool : false,
    optionalString : "",
    optionalBytes : new Uint8Array(0),
    optionalNestedEnum: TestAllTypesProto3_NestedEnum.FOO,
    // ...
  };
  t.optionalNestedMessage = {
    a: 123
  };
  t.repeatedBoolWrapper = [
    {value: true}
  ];
  t.optionalAny = {
    value: new Uint8Array(0),
    typeUrl: ""
  }
  expect(t).toBeDefined();
})
```

This example is incomplete, there are dozens of fields missing. But this will let you ensure that optional fields remain optional, arrays and map remain required, that it recurses correctly, including arrays, maps, and oneofs, etc. And it is dead simple, because it focuses on types only.

WDYT?

packages/protobuf-test/src/message.test.ts
```diff
@@ -12,23 +12,155 @@
...
+      FIELDName12: 0,
+      FieldName13: 0,
+      FieldName14: 0,
+      fieldName15: 0,
+      fieldName16: 0,
+      fieldName17: 0,
+      FieldName18: 0,
+    };
+
+    msg.optionalTimestamp = Timestamp.now();
```
timostamm, line 153: This assignment does not help us much. Consider the following example:

```ts
type A = {
  a: string;
}
type B = {
  a: string;
  b: string;
}
let a: A = {
  a: ""
}
let b: B = {
  a: "",
  b: ""
}
a = b;
```

`b` can be assigned to `a`, even though it has properties that `A` doesn't have. TypeScript simply does structural matching here, and only looks for the properties needed to satisfy `A`.

But consider this:

```ts
a = {
  a: "",
  b: "" // TS2322: Type '{ a: string; b: string; }' is not assignable to type 'A'.
        // Object literal may only specify known properties, and 'b' does not exist in type 'A'.
}
```

When assigning an object literal to a type, TypeScript _does_ check for excess properties. And we can make use of that fact to prove that a PlainMessage<Timestamp> field does not have any methods:

```suggestion
    msg.optionalTimestamp = {
      nanos: 0,
      seconds: protoInt64.zero
    };
```

Please add the same for the fields 
- `optionalDuration`
- `optionalAny`
- `optionalNestedMessage`

With that, we are starting to get some meaningful coverage. But PlainMessage is recursive now, which applies to repeated fields, maps, and oneofs. To cover those cases as well, please add assignments for the following fields:

- `repeatedDuration`
- `mapStringNestedMessage`
- `oneofField` for the case `"oneofNestedMessage"`

packages/protobuf-test/src/message.test.ts
```diff
@@ -12,23 +12,155 @@
...
+
+    expect(msg).toBeDefined();
+
+    // @ts-expect-error TS2339
+    expect(msg.toBinary).toBeUndefined();
```
timostamm, line 158: The object literal assignment we did above already proved this. This line can be removed.

packages/protobuf-test/src/message.test.ts
```diff
@@ -12,23 +12,155 @@
...
     // We want to test that the type system sees this function as undefined even though it's still actually there.  So
     // we expect TS error  TS2339, but add a simple test so Jest doesn't complain there's no expectations.
     // @ts-expect-error TS2339
-    expect(plainTimestamp.toBinary).toBeDefined();
-  });
-  test("removes wkt methods from type system", () => {
-    // Custom methods of well-known types are removed as well.
-    // @ts-expect-error TS2339
-    expect(plainTimestamp.toDate).toBeDefined();
+    expect(msg.optionalTimestamp.toDate).toBeDefined();
```
timostamm, line 164: With the change suggested above, we are already proving this, and this line can be removed.

packages/protobuf/src/message.ts
```diff
@@ -154,17 +154,29 @@ export class Message<T extends Message<T> = AnyMessage> {
...
+  : F extends OneofSelectedMessage<infer C, infer V> ? { case: C; value: PlainField<V> }
+  : F extends { case: string | undefined; value?: unknown } ? F
+  : F extends { [key: string|number]: Message<infer U> } ? { [key: string|number]: PlainField<U> }
+  : F ;
+
 /**
  * PartialMessage<T> constructs a type from a message. The resulting type
  * only contains the protobuf field members of the message, and all of them
- * are optional.
+ * are optional.  The optionality of the fields is the major difference between PartialMessage and PlainMessage.
```
timostamm, line 179: Please wrap to 80c. The optionality is the _only_ difference.

timostamm (commented):
In the docs, we claim that PlainMessage<T> is assignable to T. 

Can you add a test for that?

packages/protobuf-test/src/message.test.ts
smaye81: Makes sense.  Thanks for the explanation.  Updated the PR.

timostamm (approved):

docs/runtime_api.md
```diff
@@ -579,24 +579,17 @@ sendExample(new Example());
...
-provided. And since it is not recursive, message fields must provide the exact type.
-For example:
+provided. For example:
 
 ```typescript
 let plain: PlainMessage<Example> = {
   foo: "abc",
   bar: false,
-  baz: undefined,
+  baz: 42,
```
timostamm, line 588: `baz: 42` is a type error. At the top, we define it as a message field. Feel free to remove the line entirely if you think `baz: undefined` is unclear.

timostamm (approved):
Please fix the example in the docs (see comment below), but feel free to merge besides that.

PR #311, opened 2022-11-22
Remove assertion that "any `T` is assignable to `PlainMessage<T>` from runtime api docs

McTano:
Remove assertion that "any `T` is assignable to `PlainMessage<T>` from runtime api docs

The docs for the `PlainMessage` utility type currently ends with the sentence "Note that any `T` is assignable to `PlainMessage<T>`."

This seems to imply that the type argument `T` can be any type, not necessarily a Message, but the type definition does declare `<T extends Message>`.


docs/runtime_api.md
```diff
@@ -600,8 +600,6 @@ As such, `PlainMessage<T>` can be a great fit to use in throughout your business
 if that business logic is never concerned with serialization, and does not need 
 `instanceof`.
 
-Note that any `T` is assignable to `PlainMessage<T>`.
```
smaye81, line 603: Rather than delete this, let's maybe say:

> Note that any T that extends `Message` is assignable to `PlainMessage<T>`

smaye81:
Thanks for the PR @McTano.  Left one small comment.  Can you sign the CLA Agreement, also?

docs/runtime_api.md
smaye81: Maybe we just clarify based on what you said above:

> Note that any `T` (assuming `T` extends `Message`) is assignable to a variable of type `PlainMessage<T>`.

McTano: I like that.

smaye81 (approved):

PR #312, opened 2022-11-23
Directly Export PluginInit for Testing

dimitropoulos:
I needed this when writing tests for plugins.  We already indirectly export it:

```ts
type PluginInit = Parameters<typeof createEcmaScriptPlugin>[0]
```

(which, is what I had to do to keep going with my tests)

This PR just makes it directly exported.


timostamm (commented):
We shouldn't export `PluginInit` from the package directly. We might want to support generating code for other languages at some point, that's why all ECMAScript-specific code is in the subpath `ecmascript`, with the exception of `createEcmaScriptPlugin()`. If we export `PluginInit`, it will conflict with the type for `createLolcodePlugin()`.

Exporting the interface from `ecmascript` would avoid that, but I'm curious why you needed it for writing tests?

dimitropoulos:
cool cool, closing for now

PR #316, opened 2022-11-28
Add an FAQ to documentation

smaye81:
This adds an FAQ to our documentation and links to it from the main README.


docs/faq.md
```diff
@@ -0,0 +1,104 @@
...
+
+  > In general, you should consider using classes when you want to create objects that store their own internal data and expose a lot of behavior.
+
+  Which is exactly what our aim is when generating code from a Protobuf file.
+
+- Our messages contain numerous additional functions such as constructing an object, serializing to and from JSON, serializing to and from binary data, and cloning.  Most of this functionality is reusable, so using inheritance with a `Message` super class is beneficial.
+
+- We want the generated code to be approachable.  While the behavior of classes under the hood may be surprising as to how they're constructed and represented, they make for a very easy-to-read representation of a message.  The encapsulation they provide and some assurances that TypeScript provides on top of them make for a compelling argument to readability.
+
+So, in summary, yes, we know that JavaScript is not an OOP language and classes were shoehorned in, but our use case seems to be the reason they were added.
```
timostamm, line 61: Is JavaScript not object-oriented? It has first-class functions, but has _also_ had prototype-based object-orientation from the very start. To quote Douglas Crockford:

> What could be more object oriented than that?

It is a language that provides multiple paradigms. We picked the object-oriented approach to represent messages because it is a good fit.

docs/faq.md
```diff
@@ -0,0 +1,104 @@
...
+
+- How should they be generated?  As part of the current `protoc-gen-es` plugin?  In that case, generated code would include an additional interface alongside the class which could be confusing to users as to which one they should use.  
+
+  If we provided an option to generate interfaces, then in addition to the above problem, we now have a plugin option that could be confusing.  We also are striving to limit the number of options in our plugin as we find that only makes the plugin less-approachable.
+  
+  If we made this a separate plugin, then it seems to *really* confuse the matter because now users have to configure another plugin.  And because of the way plugins work, the separate plugin would generate new files presumably named something like `msg_interface_pb.ts`.  If users want to refactor to use the generated interface, they have to change their import statement path.  If users want to use both, they need two separate imports now.  Granted, these all may seem inconsequential at first, but they add additional overhead with arguably little payoff.
```
timostamm, line 69: I think the main thing users coming from GraphQL or JSON schema are wondering about is why we don't generate an interface _exclusively_. Do you think you can answer this question here before going into why _both_ class and interface is not working out?

timostamm (commented):
This adds a new section to the top-level readme. Can we move the section "What features are implemented" to the FAQ? That way, we keep the top-level readme concise, and it seems to be a good fit for the FAQ. WDYT?

Besides that, a couple of remarks from me, see comments below:

smaye81:
@timostamm Made the requested updates.  Let me know what you think.

timostamm (approved):
I think it's great, thank you for this FAQ :)

PR #317, opened 2022-11-29
Add plugin option `import_extension` to replace the `.js` extension in import paths

timostamm:
### `import_extension=.js`

By default, [protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es) (and all other plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin)) uses a `.js` file extensions in import paths, even in TypeScript files.

This is unintuitive, but necessary for [ECMAScript modules in Node.js](https://www.typescriptlang.org/docs/handbook/esm-node.html). Unfortunately, not all bundlers and tools have caught up yet, and Deno requires `.ts`. With this plugin option, you can replace `.js` extensions in import paths with the given value. For example, set

- `import_extension=` to remove the `.js` extension
- `import_extension=.ts` to replace the `.js` extension with `.ts`

### DIFF
--- a/packages/protoc-gen-es/README.md
+++ b/packages/protoc-gen-es/README.md
@@ -124,11 +124,26 @@ By default, we generate JavaScript and TypeScript declaration files, which
 produces the smallest code size. If you prefer to generate TypeScript, use
 `target=ts`.
 
+### `import_extension=.js`
+
+By default, [protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es)
+(and all other plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin))
+uses a `.js` file extensions in import paths, even in TypeScript files.
+
+This is unintuitive, but necessary for [ECMAScript modules in Node.js](https://www.typescriptlang.org/docs/handbook/esm-node.html).
+Unfortunately, not all bundlers and tools have caught up yet, and Deno 
+requires `.ts`. With this plugin option, you can replace `.js` extensions 
+in import paths with the given value. For example, set
+
+- `import_extension=` to remove the `.js` extension
+- `import_extension=.ts` to replace the `.js` extension with `.ts`
+
+
 ### `keep_empty_files=true`
 
-By default, [protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es) 
-(and all other plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin)) 
-omit empty files from the plugin output. This option disables pruning of 
-empty files, to allow for smooth interoperation with Bazel and similar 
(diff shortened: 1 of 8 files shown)


packages/protoplugin/src/ecmascript/generated-file.ts
```diff
@@ -283,6 +291,7 @@ type MakeImportStatementFn = (
 function processImports(
   el: El[],
   importerPath: string,
+  rewriteImportPath: RewriteImportPathFn,
```
timostamm, line 294: It's a bit inconvenient to carry this function around, but it gives us certainty that we touch all managed import statements.

packages/protoc-gen-es/README.md
```diff
@@ -124,11 +124,26 @@ By default, we generate JavaScript and TypeScript declaration files, which
...
+By default, [protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es)
+(and all other plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin))
+uses a `.js` file extensions in import paths, even in TypeScript files.
+
+This is unintuitive, but necessary for [ECMAScript modules in Node.js](https://www.typescriptlang.org/docs/handbook/esm-node.html).
+Unfortunately, not all bundlers and tools have caught up yet, and Deno 
+requires `.ts`. With this plugin option, you can replace `.js` extensions 
+in import paths with the given value. For example, set
+
+- `import_extension=` to remove the `.js` extension
```
timostamm, line 138: @mfridman is suggesting a more explicit `import_extension=none`.

packages/protoplugin-test/src/import_extension.test.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Generates a single file using the plugin framework and the given print function,
+ * passing the given options to the plugin.
+ * The returned string array represents the generated file content created using printFn.
+ */
+function generate(
+  printFn: (f: GeneratedFile, schema: Schema) => void,
+  options: string[]
+): string[] {
+  const req = new CodeGeneratorRequest({
```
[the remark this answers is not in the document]
timostamm: I'm very wary of using it everywhere because it makes the test depend on large parts of @bufbuild/protoplugin and @bufbuild/protobuf. Ideally, unit tests test only small units of code.

Removed it from two other locations in https://github.com/bufbuild/protobuf-es/pull/317/commits/fff972562e2ff408219ea8aa2953f9eaacf40985. The few tests that need a `Desc*` now just get one via getDescriptorSet().

packages/protoc-gen-es/README.md
timostamm: Updated in https://github.com/bufbuild/protobuf-es/pull/317/commits/374cfbef890bdfc16f7830748d047499adbefb59

PR #318, opened 2022-11-30
Improve JSON parse error message

timostamm:
We already provide helpful error messages when JSON input does not have the expected shape. But Message.fromJsonString() does not. This wraps errors raised by JSON.parse(), and adds a prefix with the message type name.

```ts
// currently:
Example.fromJsonString("this is not json") // Unexpected token h in JSON at position 1

// with this PR:
Example.fromJsonString("this is not json") // cannot decode Example from JSON: Unexpected token h in JSON at position 1

### DIFF
--- a/packages/protobuf-test/src/google/protobuf/struct.test.ts
+++ b/packages/protobuf-test/src/google/protobuf/struct.test.ts
@@ -71,6 +71,14 @@ describe("google.protobuf.Value", () => {
       });
       expect(value.toJsonString()).toBe("true");
     });
+    test("encoding unset value to JSON raises error", () => {
+      // Absence of any variant indicates an error.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/protobuf-test/src/google/protobuf/struct.test.ts [hunk 1]
```diff
@@ -71,6 +71,14 @@ describe("google.protobuf.Value", () => {
       });
       expect(value.toJsonString()).toBe("true");
     });
+    test("encoding unset value to JSON raises error", () => {
```
timostamm, line 74: Unrelated - just adding some coverage here because it is an odd case.

smaye81 (approved):
Nice improvement.

Issue #314, opened 2022-11-25
`T` is not assignable to `PartialMessage<T>` for messages with `Oneof` fields


MonoidMusician:
For messages with `Oneof`, the generated message type `T` is not assignable to `PartialMessage<T>` because the generated message type includes a case `{ case: undefined, value?: undefined }` that is not added by `PartialMessage`.

I have a patch for this which seems to fix it for our use-case (both of the additions are necessary):
https://github.com/id3as/protobuf-es/commit/c8994e4baa29e694a49ec277ba54896e203b0ad7

MonoidMusician:
Sure, here's a small repro, it just needs a message that contains a message with a `oneof` field: https://gist.github.com/MonoidMusician/e0d3d29c6ce6973dc41473aac011db7c

I don't have enough experience with TypeScript to know why it's dropping the undefined case, but the error is clearly that `Foo` contains `{ case: undefined; value?: undefined; }` and `PartialMessage<Foo>` doesn't.

```
Type '{ value: string; case: "fieldCase"; } | { case: undefined; value?: undefined; }' is not assignable to type '{ value: string; case: "fieldCase"; }'.
```

If I add another case, that shows up, so it isn't just dropping extra cases.
```
Type '{ value: string; case: "fieldCase"; } | { value: number; case: "otherCase"; } | { case: undefined; value?: undefined; }' is not assignable to type '{ value: string; case: "fieldCase"; } | { value: number; case: "otherCase"; }'.
```

smaye81:
Hmm, I still can't seem to recreate that issue here.  What version of Protobuf-ES are you using?  Also, what version of TypeScript?

MonoidMusician:
Huh, upgrading TypeScript `4.8.4 -> 4.9.3` fixes it. Not sure why that is.

jcready:
Is the fact that OP has `strictNullChecks` turned off the issue? https://gist.github.com/MonoidMusician/e0d3d29c6ce6973dc41473aac011db7c#file-tsconfig-json-L12

MonoidMusician:
Yes, changing that to true also seems to fix it on 4.8.4 … but leaves me with a different pile of errors that I don't want to fix right now 😂

smaye81:
Yeah I do see the errors now if I turn that off (and also `strictPropertyInitialization`).  @MonoidMusician are you seeing a different pile of errors within Protobuf-ES or your own code when you set that to true?

MonoidMusician:
[...]

PR #319, opened 2022-11-30
Improve TS docs on support

smaye81:
A user encountered [an issue](https://github.com/bufbuild/protobuf-es/issues/314) with building because they had `strictNullChecks` set to `false`.  We run our compatibility settings against the defaults generated by TypeScript's `tsc --init`.  The `strict: true` setting is a default, so `strictNullChecks: false` is actually not an option we should claim to support.

However, this should be made more clear in our docs and the README has been updated to reflect that.


README.md
```diff
@@ -64,7 +64,9 @@ The implementation is covered by the protocol buffers
 
 ### TypeScript
 
-The minimum version of TypeScript supported by Protobuf-ES is **4.1.2**.
+The minimum version of TypeScript supported by Protobuf-ES is **4.1.2**.  Our compatibility tests use the default 
+settings for `tsconfig.json` that are generated by running `tsc --init` (with a few minor exceptions).  As a result,
+for any version we support, we inherently support the default settings respective to that version of TypeScript.
```
timostamm, line 69: What are the exceptions?

smaye81: We specify `skipLibCheck: false` down to 4.4.4. and `moduleResolution: NodeNext` where it is supported.

timostamm: Yes, but both settings are actually _more_ restrictive than the defaults. So we are compatible with the default compiler settings for all versions of TypeScript >= 4.1.2.

I feel like I have to read the paragraph carefully, perhaps more than once, to understand what's supported. 

Isn't this simple alternative also true?

> The generated code is compatible with TypeScript v4.1.2 or later, with the default compiler settings.

smaye81: Sure.  I just wanted to be clear in case anyone noticed that we don't officially use the defaults in all versions.

timostamm: We already have comments in the tsconfig files that _should_ explain that. If it isn't clear enough, let's improve them. Or move the more restrictive settings to separate tsconfig files, or remove them completely.

smaye81: Sounds good.  Updated.

timostamm: TY. I only had reservations about the top-level readme - the details you added in https://github.com/bufbuild/protobuf-es/pull/319/commits/467154d3e97d77a982c6fa05bd113b3fb720f542#diff-c73cc5d535f51338cff7c9594228cae1b32f902aeeac134be925efc2a7c9af35 seem helpful!

Regarding strictNullChecks - protobuf-ts has a [note in it's manual](https://github.com/timostamm/protobuf-ts/blob/master/MANUAL.md#oneof-representation). Considering a user ran into the issue, perhaps we should add a similar note to [the docs here](https://github.com/bufbuild/protobuf-es/blob/main/docs/generated_code.md#oneof-groups)? We no longer have to explain that users need to set `strict` to true, it's all on by default in all TS version we support. But it'll be helpful for users who have strictNullChecks disabled for legacy reasons and run into this.

smaye81: Oh nice.  OK, I added that note in and put my coment on the test README back in.

timostamm (approved):

PR #335, opened 2022-12-06
Make toJSON private

smaye81:
This makes the `toJSON` method on `Message` private.  This is done for two reasons:

- Users should never need to invoke this directly.  They should use `JSON.stringify(msg)` to indirectly invoke it.
- Showing this function as part of the public API of `Message` is confusing when listed with the other `toJson` function.

### DIFF
--- a/packages/protobuf/src/message.ts
+++ b/packages/protobuf/src/message.ts
@@ -149,7 +149,7 @@ export class Message<T extends Message<T> = AnyMessage> {
    */
   //  eslint-disable-next-line @typescript-eslint/ban-ts-comment
   //  @ts-ignore
-  private toJSON(): JsonValue {
+  protected toJSON(): JsonValue {
     return this.toJson({
       emitDefaultValues: true,
     });
--- a/packages/protobuf/src/message.ts
+++ b/packages/protobuf/src/message.ts
@@ -141,8 +141,15 @@ export class Message<T extends Message<T> = AnyMessage> {
    * unpacked, and this is only possible with a type registry to look up the
    * message type.  As a result, attempting to serialize a message with this
    * type will throw an Error.
+   *
+   * Note that this function is private because users should not need to invoke
+   * it directly -- they can use JSON.stringify.  Additionally, if it is public
+   * it shows as part of the Message API to users which can be confusing when
+   * listed with the other toJson method
    */
-  toJSON(): JsonValue {
+  //  eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //  @ts-ignore
+  private toJSON(): JsonValue {
     return this.toJson({
       emitDefaultValues: true,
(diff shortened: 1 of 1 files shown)


packages/protobuf/src/message.ts
```diff
@@ -141,8 +141,15 @@ export class Message<T extends Message<T> = AnyMessage> {
...
+   *
+   * Note that this function is private because users should not need to invoke
+   * it directly -- they can use JSON.stringify.  Additionally, if it is public
+   * it shows as part of the Message API to users which can be confusing when
+   * listed with the other toJson method
    */
-  toJSON(): JsonValue {
+  //  eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //  @ts-ignore
+  private toJSON(): JsonValue {
```
timostamm, line 152: What do you think about making it `protected` to avoid the two linter comments?

timostamm: It's definitely worth making sure `protected` doesn't have unintended side-effects, but regarding subclassing, we clearly warn against doing that [here](https://github.com/bufbuild/protobuf-es/blob/b430d99fd8064b9073b2c817faa244257f5df5a2/packages/protobuf/src/message.ts#L37).

packages/protobuf/src/message.ts
```diff
@@ -141,8 +141,15 @@ export class Message<T extends Message<T> = AnyMessage> {
...
    * type will throw an Error.
+   *
+   * Note that this function is private because users should not need to invoke
+   * it directly -- they can use JSON.stringify.  Additionally, if it is public
+   * it shows as part of the Message API to users which can be confusing when
+   * listed with the other toJson method
    */
-  toJSON(): JsonValue {
+  //  eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //  @ts-ignore
```
timostamm, line 151: Don't need this with `protected`

```suggestion
```

packages/protobuf/src/message.ts
```diff
@@ -141,8 +141,15 @@ export class Message<T extends Message<T> = AnyMessage> {
    * unpacked, and this is only possible with a type registry to look up the
    * message type.  As a result, attempting to serialize a message with this
    * type will throw an Error.
+   *
+   * Note that this function is private because users should not need to invoke
+   * it directly -- they can use JSON.stringify.  Additionally, if it is public
+   * it shows as part of the Message API to users which can be confusing when
+   * listed with the other toJson method
```
timostamm, line 148: It's now "protected", no longer private. We should call it "method" instead of "function". Last sentence is missing a full stop.

smaye81: Ugh sorry.  That is my fault.

timostamm (approved):

PR #338, opened 2022-12-12
Updated Protobuf and TypeScript deps

smaye81:
This updates Protobuf-ES to the latest protocolbuffers/protobuf release (`21.11`) and the latest TypeScript release (`4.9.4`).

In doing so, this caused cascading issues during the build process.  A summary:

We included our own type definition for `TextEncoder` and `TextDecoder` so that we didn't have to pull in the entire DOM lib and inadvertently depend on something that didn't have widespread support.  But, `@types/node` actually added these same two types to their global scope in `v18.11.x` (see change [here](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/85092f93d4fa4f7139ca36c686d38ed93cf0ef54#diff-fe20cb7d2536cc42b8e485f38bacbb45f55937a656837f2108594f76b9680acdR1111-R1136)).  Therefore, when trying to compile the code with our custom type AND an upgraded version of `@types/node`, we saw compilation issues with duplicate definitions of `TextEncoder` and `TextDecoder`.  To get around this, we simply pinned `@types/node` to a specific version that didn't have these types defined.

The problem though is that TypeScript `4.9.x` upgraded to include the latest `@types/node` types as part of their built-in types.  So, upgrading TypeScript to `4.9.x` causes these same duplicate type issues to reoccur.

We are already including these `@types/node` types as part of our compilation process, so it seems to fix the issues if we just remove our custom type declarations and rely on the ones in `@types/node`.   This _should_ keep things as they are currently as:

1. we still rely on `@types/node`
2. we still don't pull in any of `lib.dom.ts`
3. and we still have a definition for the decoders.

As a result of the above, we were also able to remove the explicit `@types/node` typings from our package.json files.

Finally, there seems to be some TypeScript compatibility issues with `@types/jest` and TypeScript `4.1.2`.  If `@types/jest` is upgraded from `29.2.1` to anything greater (`29.2.2` fails for example), a ton of compilation issues start to show in our compatibility tests on `4.1.2` only.  This seems to be because `@types/jest` has TypeScript syntax that was introduced in `4.2`, such as [abstract constructor signatures](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html#abstract-construct-signatures) with their `SpiedClass`, `SpiedGetter`, and `SpiedSetter` types.

To fix this, we removed `@types/jest` altogether and instead are relying on Jest globals.  This allows us to only import what we use and since we aren't using the offending `SpiedClass` types above, we can fix the compilation issue on `4.1.2`.  This will also allow us to upgrade Jest whenever we want whereas before we would've been stuck on `29.2.1`.

This has the unfortunate side effect of having to add explicit imports to tests, but it prevents a ton of extra types from being sucked in, allows us to upgrade Jest freely, and still keeps us compatible with TypeScript `4.1.2`.


timostamm (commented):
There seems to be an issue with typings for Node.js and TextDecoder, see CI error.

packages/protobuf-test/typescript/tsconfig.4_4_4.json
```diff
@@ -10,6 +10,6 @@
     "strict": true,
     // To guard against regression and ensure we are remaining backwards
     // compatible, set the skipLibCheck flag to false explicitly.
-    "skipLibCheck": false
+    "skipLibCheck": true
```
smaye81, line 13: There is a compilation issue in `@jest/mocks` when this is set to false.

timostamm (approved):

PR #346, opened 2022-12-15
Add better docs to toJSON

smaye81:
This adds better documentation to the protected `toJSON` method.  Users should use `JSON.stringify` or `toJsonString` if they need stringified JSON.  If they want an actual JSON object, they should use `toJson`.

Also updated release notes to reflect that making `toJSON` protected was a breaking change:  https://github.com/bufbuild/protobuf-es/releases/tag/v0.5.0

### DIFF
--- a/packages/protobuf/src/message.ts
+++ b/packages/protobuf/src/message.ts
@@ -142,9 +142,9 @@ export class Message<T extends Message<T> = AnyMessage> {
    * message type.  As a result, attempting to serialize a message with this
    * type will throw an Error.
    *
-   * This method is protected because users should not need to invoke it
-   * directly -- they should use JSON.stringify or toJsonString for
-   * stringified JSON.  Alternatively, if actual JSON is desired, users should
+   * This method is protected because you should not need to invoke it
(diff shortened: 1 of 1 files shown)


packages/protobuf/src/message.ts
```diff
@@ -142,8 +142,10 @@ export class Message<T extends Message<T> = AnyMessage> {
    * message type.  As a result, attempting to serialize a message with this
    * type will throw an Error.
    *
-   * Note that this method is protected because users should not need to invoke
-   * it directly -- they can use JSON.stringify.
+   * This method is protected because users should not need to invoke it
+   * directly -- they should use JSON.stringify or toJsonString for
+   * stringified JSON.  Alternatively, if actual JSON is desired, users should
+   * use toJson.
```
timostamm, line 148: All JSDoc is documentation for our users. As a reader, it gives _you_ documentation on how to use it. WDYT about addressing users with "you" consistently?

timostamm (approved):

packages/protobuf/src/message.ts
smaye81: Works for me.  Updated.

Issue #347, opened 2022-12-19
Provide an option to use immutable default values for repeated & map fields


newyankeecodeshop:
Our protobuf object model includes several types that have several repeated properties. These properties are often set to the default (empty array) value because the properties are only used for certain states. Our browser-based app downloads thousands of these objects using `connect-web`, and this creates hundreds of thousands of empty repeated values. 

I propose that an option be added to `BinaryReadOptions` and `JsonReadOptions`:
`defaultsImmutable: boolean;`

When this option is on, default repeated and map objects are constructed using a single (frozen) `Array` or `Object` instance that is shared across all protobuf Message instances. The value is immutable, and thus there is no need to have multiple instances of an empty array or empty object.

This options provides our app with a major reduction in the amount of memory consumed. Over one million array objects were optimized (see screenshot from the Chrome memory view):
![Array proto optimization](https://user-images.githubusercontent.com/445421/208448401-9a656989-6476-4035-8e8b-302fff8a7e9c.jpg)

I added a single unit test for this, but more pervasive testing would be good. I'm happy to add that if you approve of this idea.

### Issue

The TypeScript declaration files use the inline member initialization syntax, so there was no way for me to implement this feature for TypeScript files. Fixing this will require more extensive changes, but I didn't want to implement those changes unless you approved of this feature.

smaye81:
Hi, @newyankeecodeshop thanks for the issue and PR.  Unfortunately, I don't think this is something we plan on implementing at this point.  I get what you're saying and there are some upsides to what you're doing, but this is a pretty drastic change and more importantly, seems to be a footgun for developers.  This would essentially make all modifying methods on the array (and map) inaccessible, which could be what you're aiming for here, but seems a bit specialized to your use case. 

So, it is hard to do this in a reasonable way with a runtime option and making this a plugin option would complicate the runtime and the code generator. Again, thank you for the suggestion, but this is something we won't implement right now.  Sorry!

newyankeecodeshop:
Thanks @smaye81 for the prompt review. I appreciate the feedback.

PR #348, opened 2022-12-19
Add option to use immutable default array,map

newyankeecodeshop:
This is an implementation of the idea described in issue #347 .

I look forward to your thoughts!


smaye81:
Discussed this on #347, but this isn't something we will be implementing at this time.  Thank you for the submission though.

PR #359, opened 2023-01-03
Pin Bazel version

smaye81:
This pins the Bazel version to `5.4.0`, which is a version compatible with our version of Protocol Buffers (`21.12`).  The upgrade to Bazel 6 causes our builds to fail and this ensures that we use `5.4.0` when building.

Note this assumes Bazelisk is being used to enforce the Bazel version.  However, we can't be sure all users are using Bazelisk with their Bazel install.  The contributing docs have been updated to reflect that `5.4.0` is the latest supported version.


gilwong00 (approved):
LGTM!

PR #374, opened 2023-01-27
Allow optional fields to be explicitly set to `undefined`

haines:
This PR allows fields to be set to `undefined` explicitly even when the TypeScript compiler option `exactOptionalPropertyTypes` is enabled.

Currently optional message fields are typed as

```typescript
field?: Type
```

but with `exactOptionalPropertyTypes` this means you can't set them to `undefined`; you have to omit them. This is a bit inconvenient if you're trying to pass through a value from a variable.

This PR changes the optional fields to be typed as

```typescript
field?: Type | undefined
```

which removes the restriction.

### DIFF
--- a/packages/protobuf-bench/README.md
+++ b/packages/protobuf-bench/README.md
@@ -10,5 +10,5 @@ server would usually do.
 
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
-| protobuf-es         | 74,150 b      | 36,950 b | 9,642 b |
+| protobuf-es         | 74,150 b      | 36,950 b | 9,636 b |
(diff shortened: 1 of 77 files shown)


smaye81:
Hey @haines thank you for the PR.  Thank you for being patient.

Our approach has been that [we support the default compiler settings](https://github.com/bufbuild/protobuf-es#typescript) for a particular version of TypeScript.  And since `exactOptionalPropertyTypes: true` is not a default setting, I don't think it's something we want to factor into the generated code.  It's just a can of worms we are trying to avoid because trying to handle the myriad options in tsconfig would be a difficult task.

That being said, if this becomes a default setting, we would certainly support it at that time.  Thanks again for the consideration.

PR #386, opened 2023-02-03
Strip rewrite_imports parameter

### LINKED ISSUE #343: Remove the rewrite_imports preamble
I think we should consider removing the `rewrite_imports` preamble from the generated code. It's more of an implementation detail (and feature) of the BSR. By removing this we get:

1. cleaner generated code.
2. ability to improve the underlying glob patterns upstream without affecting the imports. We likely don't want to do this, but if we do then it'll be seamless to the user without diff changes.
3. this is helpful for debugging, but an issue with this chunk of code will still result in the wrong import path. So that's the thing the end-user should be focusing on, and less on this option.

```
// @generated by protoc-gen-connect-web v0.3.3 with parameter "rewrite_imports=./pet/v1/**/*_pb.js:@buf/acme_petapis.bufbuild_es,rewrite_imports=./payment/v1alpha1/**/*_pb.js:@buf/acme_paymentapis.bufbuild_es,rewrite_imports=./payment/v1alpha1/**/*_connectweb.js:@buf/acme_paymentapis.bufbuild_connect-web,rewrite_imports=./google/type/**/*_pb.js:@buf/googleapis_googleapis.bufbuild_es,rewrite_imports=./google/type/**/*_connectweb.js:@buf/googleapis_googleapis.bufbuild_connect-web"
// @generated from file pet/v1/pet.proto (package pet.v1, syntax proto3)
/* eslint-disable */
// @ts-nocheck
```

smaye81:
Fixes #343.

This strips the `rewrite_imports` parameter from the generated file preamble as that is mostly an implementation detail and can be very noisy  (and not very helpful) to consumers/readers of the generated code.


packages/protoplugin/src/ecmascript/gencommon.ts
```diff
@@ -59,7 +59,12 @@ export function makeFilePreamble(
   writeLeadingComments(file.getSyntaxComments());
   builder.push(`// @generated by ${pluginName} ${pluginVersion}`);
   if (parameter !== undefined) {
-    builder.push(` with parameter "${parameter}"`);
+    // rewrite_imports can be noisy and is more of an implementation detail
+    // so we strip it out of the preamble
+    const stripped = stripParameter(parameter, "rewrite_imports");
```
timostamm, line 64: It doesn't seem quite right to parse the "parameter" here. It is just a convention that options are delimited with a comma, and I think it's better to have just one piece of code dealing with the ambiguitiy.

We already parse the parameter in `createEcmaScriptPlugin()`. Let's just return the parameter we want to render there, pass it as a new argument to `createSchema()` (I'd suggest the name `pluginParameter`), and then to `createGeneratedFile()`.

This should be fairly clean, and we won't stumble over a user-defined plugin option that happens to start with "rewrite_imports".

smaye81: The parameter is actually already passed to `createSchema` along with other parameters.  Do you mean that this new `pluginParameter` argument represents a string of the plugin parameters we want to render in the preamble?

timostamm: Yes.

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -291,14 +300,20 @@ function parseParameter(
         }
         break;
     }
+    if (printToFile) {
+      renderParams.push(`${key}=${value}`);
```
timostamm, line 304: What if my custom plugin has an option that is a simple flag, not a key-value pair?

Consider:

```yaml
opt: foo,bar=,baz=abc
``` 

I think the return type of splitParameter has to change, so that we can distinguish between `foo` - a flag, and `bar=`, a key-value with an empty value.

Might be most straight-forward to return:

```ts
{ key: string; value: string; raw: string }[]
``` 

With `raw` being the unparsed element, that we can easily join again.

packages/protobuf/src/google/protobuf/api_pb.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @generated by protoc-gen-es v1.0.0 with parameter "bootstrap_wkt=true,ts_nocheck=false,target=ts"
+// @generated by protoc-gen-es v1.0.0 with parameter "bootstrap_wkt=truets_nocheck=falsetarget=ts"
```
timostamm: Not sure what's happening here, but we lost all commas 😆

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -313,6 +329,7 @@ function splitParameter(
     return {
       key: i === -1 ? pair : pair.substring(0, i),
       value: i === -1 ? "" : pair.substring(i + 1),
+      raw: pair,
```
timostamm, line 332: Small code style nit: Can be

```suggestion
      raw,
```

if you rename the variable `pair`.

packages/protoplugin/src/ecmascript/gencommon.ts
```diff
@@ -58,7 +58,7 @@ export function makeFilePreamble(
   };
   writeLeadingComments(file.getSyntaxComments());
   builder.push(`// @generated by ${pluginName} ${pluginVersion}`);
-  if (parameter !== undefined) {
+  if (parameter !== undefined && parameter !== "") {
```
timostamm, line 61: We never pass `undefined` to this function. Please change line 37 to be `parameter: string`, without `| undefined`.

Then this can be:

```suggestion
  if (parameter !== "") {
```

timostamm (approved):

PR #387, opened 2023-02-06
Support size-delimited messages

timostamm:
This implements support for size-delimited messages described in https://github.com/protocolbuffers/protobuf/issues/10229

A size-delimited message is a varint size in bytes, followed by exactly that many bytes of a message serialized with the binary format. This implementation is compatible with the counterparts in [C++](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/delimited_message_util.h), [Java](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/AbstractParser.html#parseDelimitedFrom-java.io.InputStream-), [Go](https://github.com/golang/protobuf/issues/1382), and others.

You create such a message with `protoDelimited.enc`:

```typescript
import { protoDelimited } from "@bufbuild/protobuf";

const bytes = protoDelimited.enc(new User({ firstName: "John" }));
const user = protoDelimited.dec(User, bytes);
```

With `protoDelimited.decStream`, you can parse messages from a stream. The method expects an `AsyncIterable<Uint8Array>` as a stream input, so it works with Node.js streams out of the box, and can be easily adapted to other stream APIs:

```typescript
import { protoDelimited } from "@bufbuild/protobuf";
import { createReadStream, createWriteStream } from "fs";
import { tmpdir } from "os";
import { join } from "path";

// Let's write a couple of messages to a file
const ws = createWriteStream("protoDelimited.bin", {encoding: "binary"});
ws.write(protoDelimited.enc(new User({ firstName: "John" })));
ws.write(protoDelimited.enc(new User({ firstName: "Max" })));
ws.write(protoDelimited.enc(new User({ firstName: "Max" })));
ws.end();
ws.close();

// Now we can parse them from the stream
const readStream = createReadStream("protoDelimited.bin");
for await (const user of protoDelimited.decStream(User, readStream)) {
  console.log(user);
}
```


packages/protobuf-test/src/proto-delimited.test.ts
```diff
@@ -0,0 +1,216 @@
...
+          throw new Error("incomplete data");
+        }
+      }
+      it("should decode messages", async function () {
+        const decoded: TestAllTypesProto3[] = [];
+        for await (const m of decStream(stream, TestAllTypesProto3)) {
+          decoded.push(m);
+        }
+        expect(decoded).toStrictEqual(testMessages);
+      });
```
timostamm, line 213: An example for using `protoDelimited.peekSize` to read from an `AsyncIterable<Uint8Array>`.

akshayjshah: This is surprisingly readable!

akshayjshah (commented):
I'm almost completely unable to parse the actual code, but the unit tests look nice! As long as we're following the proposal in the discussion you linked, this lgtm. (I assume that short names like `enc` and `dec` match the rest of the code.)

timostamm:
Have added the `decStream` method to the object after running some tests. This PR also adds documentation now.

smaye81 (approved):

PR #391, opened 2023-02-15
adds `safeIdentifier` and `safeObjectProperty` helpers

dimitropoulos:
It came up in https://github.com/bufbuild/connect-query/pull/19 that we need to export the functionality of safely modifying names for identifiers.

This PR adds `safeIdentifier` and `safeObjectProperty` helpers which do exactly that.  All prior behavior should be the same (except for these functions now being isolated, tested, and exported).


packages/protobuf/src/private/names.ts
```diff
@@ -89,7 +80,7 @@ export function localFieldName(protoName: string, inOneof: boolean) {
     // oneof member names are not properties, but values of the `case` property.
     return name;
   }
-  if (reservedObjectProperties[name] || reservedMessageProperties[name]) {
+  if (isReservedObjectProperty(name) || reservedMessageProperties[name]) {
     name = name + "$";
```
dimitropoulos, line 84: I didn't update `reservedMessageProperties` like the other two (`reservedObjectProperties` and `reservedIndent`) in the name of keeping scope down, but if you'd like me to make them all look the same I'd be happy to.

packages/protobuf/src/private/names.ts
```diff
@@ -275,3 +266,43 @@ const reservedMessageProperties: { [k: string]: boolean } = {
...
+  word: T | ReservedObjectProperty
+): word is ReservedObjectProperty =>
+  Boolean(reservedObjectProperties[word as ReservedObjectProperty]);
+
+export const safeObjectProperty = <T extends string>(
+  word: T
+): T extends ReservedObjectProperty ? Fallback<T> : T => {
+  if (isReservedObjectProperty(word)) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any -- required for narrowing
+    return fallback(word) as any;
```
dimitropoulos, line 286: I didn't take a ton of time to mess with this: all I want is that the types reflect what's going on with the data, and they do (see the tests, which solidify this).

packages/protobuf/src/private/names.ts
```diff
@@ -246,17 +237,17 @@ const reservedIdent: { [k: string]: boolean } = {
   globalThis: true,
   Uint8Array: true,
   Partial: true,
-};
+} satisfies { [k: string]: boolean };
```
timostamm, line 240: `satisfies` is nice, but only available from TS v4.9. Please revert so we retain the option to publish types for older compilers.

packages/protobuf/src/private/names.ts
timostamm: Typically, you have a name from the schema that is a `string`, so there is not a lot of benefit to the types. Readability of this code is important, and the types do not increase readability. Also not great to need to disable the linter. Not worth the trade-off in my book. Can you remove?

packages/protobuf/src/private/names.ts
timostamm: Let's make it consistent for sure.

I think this should probably boil down to `return safeMessageProprty(safeObjectProperty(name));` Please don't export safeMessageProperty().

Please check how that affects bundle size with `make bench`.

packages/protobuf/src/private/names.ts
```diff
@@ -275,3 +266,43 @@ const reservedMessageProperties: { [k: string]: boolean } = {
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any -- required for narrowing
+    return fallback(word) as any;
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-explicit-any -- required for narrowing
+  return word as any;
+};
+
+type ReservedIdentifier = keyof typeof reservedIdent;
+
+const isReservedIdentifier = <T extends string>(
```
timostamm, line 294: We just want to conditionally append an `$` and a clear code path is valuable here. Ideally, the naming logic is easy to port to other languages and other code generators, so that they can interact. Can you remove is*()?

packages/protobuf/src/private/names.ts
dimitropoulos:
```diff
  | code generator      | bundle size             | minified               | compressed         |
  |---------------------|------------------------:|-----------------------:|-------------------:|
+ | protobuf-es         | 86,823 b      | 36,972 b | 9,654 b |
- | protobuf-es         | 86,785 b      | 36,950 b | 9,642 b |
  | protobuf-javascript | 375,550 b  | 271,672 b | 43,769 b |
```

as of e4f43261f355c39a539528e0fa16409b80e8c044

packages/protobuf/src/private/names.ts
dimitropoulos: you're quick! they were removed before I saw this comment e4f43261f355c39a539528e0fa16409b80e8c044

dimitropoulos:
Now that we found that [bug](https://github.com/bufbuild/protobuf-es/blob/ee1f610542f13ffa3dd9b1486dbe136606f87708/packages/protobuf/src/private/names.ts#L185) re: prototype inheritance being relied on, it looks like it might have been relied on because one of the fixtures fails now.  That suggest to me that we'll need to update something in the generator.  Would you mind helping me locate what?fixtures fails now.  That suggest to me that we'll need to update something in the generator.  Would you mind helping me locate what needs to be updated?

Screenshot_20230215_122115

packages/protobuf/src/private/names.ts
```diff
@@ -180,98 +168,144 @@ function protoCamelCase(snakeCase: string): string {
...
+/**
+ * Will wrap names that are Object prototype properties or names reserved
+ * for `Message`s.
+ */
+const safeMessageProperty = (name: string): string => {
+  if (
+    reservedObjectProperties.has(name) ||
+    reservedMessageProperties.has(name)
+  ) {
+    fallback(name);
```
timostamm, line 284:
```suggestion
    return fallback(name);
```

packages/protobuf/src/private/names.ts
```diff
@@ -180,98 +168,144 @@ function protoCamelCase(snakeCase: string): string {
...
+ * Names that can be used for identifiers or class properties
+ */
+export const safeIdentifier = (name: string): string => {
+  if (
+    reservedObjectProperties.has(name) ||
+    reservedIdentifiers.has(name)
+  ) {
+    return fallback(name);
+  }
+  return name;
```
timostamm, line 310:
```suggestion
export const safeIdentifier = (name: string): string => {
  if (reservedIdentifiers.has(name)) {
    return fallback(name);
  }
  return name;
```

With this change, and the one above, all tests pass for me. Can you try that? Please make sure to run `make` and commit all diffs.

packages/protobuf/src/private/names.ts
```diff
@@ -50,32 +50,23 @@ export function localName(
       const pkg = desc.file.proto.package;
       const offset = pkg === undefined ? 0 : pkg.length + 1;
       const name = desc.typeName.substring(offset).replace(/\./g, "_");
-      if (reservedIdent[name]) {
-        return name + "$";
-      }
-      return name;
+      return safeObjectProperty(safeIdentifier(name));
```
timostamm, line 53: Can you add a comment here? For example:

> For services, we only care about safe identifiers, not safe object properties, 
> but we have shipped v1 with a bug that respected object properties, and we
> do not want to introduce a breaking change.

packages/protobuf-bench/README.md
```diff
@@ -10,5 +10,5 @@ server would usually do.
 
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
-| protobuf-es         | 86,785 b      | 36,950 b | 9,642 b |
+| protobuf-es         | 87,019 b      | 36,999 b | 9,692 b |
```
timostamm, line 13: Can you look into this? We should not increase the bundle size. I know it's tedious for 50 bytes compressed, but we don't make any logical changes here... I'm happy to help looking at the bundle.

timostamm (commented):
Code looks really nice! Two comments though:

packages/protobuf-bench/README.md
dimitropoulos: I definitely need help: I cannot for the life of me determine what's going on.  see https://github.com/bufbuild/protobuf-es/tree/dimitri/test-safe-names and note that in the last commit the bundle.js file didn't change, despite the size being reported to be different.  I also know it's not from the exports or the test files (by process of elimination).

timostamm: You didn't see changes because gather() is called twice, once for protobuf-es, once for protobuf-javascript, but you always write to bundle.js, overwriting the protobuf-es bundle output with the protobuf-javascript output, which does not change with this PR.

timostamm: Took a quick stab at bundle size in https://github.com/bufbuild/protobuf-es/pull/391/commits/7c274cf8ec879ad73c3da34756f3a9c300d0eac0, but only managed to shave off two bytes from the compressed. That doesn't seem worth it, I think the readability of the safe*Property functions and the Set are more important. Reverted in https://github.com/bufbuild/protobuf-es/pull/391/commits/b58ffd7535e6b0caeaad44fb46d4c49e7e729484.

timostamm (approved):

packages/protobuf-bench/README.md
dimitropoulos: yes, I _just_ got that working, yeah.

Screenshot_20230315_142335

packages/protobuf/src/private/names.ts
dimitropoulos: (thank you! :smile: I was gonna add it after we figured out the other thing)

PR #393, opened 2023-02-17
Update to use Buf from NPM

smaye81:
This updates Protobuf-ES to use the Buf NPM package in examples, docs, and the Makefile.

### DIFF
--- a/packages/protobuf-example/package.json
+++ b/packages/protobuf-example/package.json
@@ -7,10 +7,11 @@
     "build": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "add-person": "node dist/esm/add-person.js addressbook.bin",
     "list-people": "node dist/esm/list-people.js addressbook.bin",
-    "buf:generate": "buf generate"
+    "generate": "npx buf generate"
   },
   "type": "module",
   "dependencies": {
+    "@bufbuild/buf": "^1.14.0",
     "@bufbuild/protobuf": "1.0.0",
     "@bufbuild/protoc-gen-es": "1.0.0",
     "typescript": "^4.9.4"
--- a/packages/protobuf-example/package.json
+++ b/packages/protobuf-example/package.json
@@ -7,7 +7,7 @@
     "build": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "add-person": "node dist/esm/add-person.js addressbook.bin",
     "list-people": "node dist/esm/list-people.js addressbook.bin",
-    "generate": "npx buf generate"
+    "generate": "buf generate"
   },
   "type": "module",
   "dependencies": {
--- a/packages/protoplugin-example/README.md
+++ b/packages/protoplugin-example/README.md
@@ -25,9 +25,11 @@ cd packages/protoplugin-example
 npm run build
 ```
 
-To run the plugin (i.e. generate files):
+To run the plugin (i.e. generate files), use the following command. This will generate files based on the 
+[Eliza module](https://buf.build/bufbuild/eliza) in the Buf Schema Registry (BSR).  You can change this path to generate
+additional files locally or from the BSR.
 
-`npm run generate`
+`npx buf generate buf.build/bufbuild/eliza`
 
 To run the example webpage and see the generated code in action:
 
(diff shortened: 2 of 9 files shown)


packages/protobuf-example/package.json
```diff
@@ -7,10 +7,11 @@
     "build": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "add-person": "node dist/esm/add-person.js addressbook.bin",
     "list-people": "node dist/esm/list-people.js addressbook.bin",
-    "buf:generate": "buf generate"
+    "generate": "npx buf generate"
```
timostamm, line 10:
```suggestion
    "generate": "buf generate"
```

packages/protoplugin-test/package.json
```diff
@@ -5,8 +5,8 @@
     "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
     "build": "npm run build:esm+types",
     "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
-    "buf:build": "buf build -o descriptorset.bin",
-    "buf:generate": "buf generate",
+    "buf:build": "npx buf build -o descriptorset.bin",
+    "generate": "npx buf generate",
```
timostamm, line 9:
```suggestion
    "buf:build": "buf build -o descriptorset.bin",
    "generate": "buf generate",
```

packages/protoplugin-example/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "clean": "rm -rf src/gen",
     "build": "../../node_modules/typescript/bin/tsc --noEmit",
-    "buf:generate": "buf generate buf.build/bufbuild/eliza",
+    "generate": "npx buf generate buf.build/bufbuild/eliza",
```
timostamm, line 9:
```suggestion
    "generate": "buf generate buf.build/bufbuild/eliza",
```

packages/protoplugin-example/README.md
```diff
@@ -25,7 +27,7 @@ npm run build
 
 To run the plugin (i.e. generate files):
 
-`npm run buf:generate`
+`npm run generate`
```
timostamm, line 30: For this plugin example, I suggest we use:

```suggestion
`npx buf generate buf.build/bufbuild/eliza`
```

In a project, it makes sense to put the command into an npm script, because you typically want to set it up once and forget. 

But in this example, it would be nice to suggest the user to change the command, and maybe generate other BSR modules, from Git, or from local proto files.

WDYT?

smaye81: That's fine.  I actually had it this way and changed it because I felt putting the URL there would be confusing and that it would have to be explained, which would then inflate the README example.

timostamm (approved):

packages/protoplugin-example/README.md
smaye81: @timostamm Updated.  Lmk if the wording sounds OK.

timostamm (approved):

packages/protobuf-example/package.json
```diff
@@ -7,10 +7,11 @@
     "build": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "add-person": "node dist/esm/add-person.js addressbook.bin",
     "list-people": "node dist/esm/list-people.js addressbook.bin",
-    "buf:generate": "buf generate"
+    "generate": "buf generate"
```
dimitropoulos, line 10: as written, this would still use a `buf` command in a path (and not from the NPM), wouldn't it?

packages/protoplugin-test/package.json
```diff
@@ -6,7 +6,7 @@
     "build": "npm run build:esm+types",
     "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types",
     "buf:build": "buf build -o descriptorset.bin",
-    "buf:generate": "buf generate",
+    "generate": "buf generate",
```
dimitropoulos, line 9: same comment here as [above](https://github.com/bufbuild/protobuf-es/pull/393/files#r1113188788)

packages/protobuf-example/package.json
timostamm: Same as with any other npm script, this runs with a modified $PATH, looking at `node_modules/.bin` first.

dimitropoulos: ohhh right right, gotcha, thanks.

PR #394, opened 2023-02-22
Upgrade to protobuf v22.0

smaye81:
This upgrades Protobuf-ES to v22.0.  A few changes were involved.  Notably:

- https://github.com/protocolbuffers/protobuf/commit/ca1cb1ba80ef18f5dccfb5b6ee7fa623ba6caab5 added checks to expect a failure when attempting to serialize `NaN` or `Infinity` with the `Value` WKT.  This adds these checks to our generated `Value` code to adhere to the conformance spec.

- [This PR](https://github.com/protocolbuffers/protobuf/commit/b2aca61f4dff165d0e6c320a5e582fe26989c95f#diff-4c6ba3d87bce4389f7883a1cffd8ec741f76a4ec05bba4a38a5018a8484cbad9) added conformance checks to check for negative values in the `Duration` WKT.  This adds additional checks and unit tests to our generated `Duration` code to handle negative durations and adhere to the conformance spec.

- protoc was [changed](https://protobuf.dev/programming-guides/migration/#json-field-names) to check the `json_name` option, and give errors for case-sensitive conflict, which caused some of our tests to fail.  Tests have been updated.

Note that v22.0 also mentions the new `editions` syntax, although this is not used anywhere.  Protobuf-ES validates the `syntax` value is only either `proto2` or `proto3` so we are handling this effectively at this point.  An example error message if `editions` is used:

```
proto/address_book.proto:15:10:syntax value must be "proto2" or "proto3"
```


packages/protobuf-test/src/gen/js/google/protobuf/struct_pb.js
```diff
@@ -126,6 +131,9 @@ Value.prototype.toJson = function toJson(options) {
 Value.prototype.fromJson = function fromJson(json, options) {
   switch (typeof json) {
     case "number":
+      if (!isFinite(json)) {
+          throw new Error("cannot decode google.protobuf.Value from JSON " + proto3.json.debug(json));
+      }
```
timostamm, line 136: It is not possible to serialize Infinity or NaN to JSON. Under which circumstances do you expect to see a NaN or Infinity here, when parsing JSON?

packages/protobuf-test/src/gen/js/google/protobuf/struct_pb.js
```diff
@@ -113,7 +113,12 @@ Value.prototype.toJson = function toJson(options) {
     case "nullValue":
       return null;
     case "boolValue":
+      return this.kind.value;
     case "numberValue":
+      if (!isFinite(this.kind.value)) {
```
timostamm, line 118: Best to use the explicit Number.isFinite here.

Previously, the switch statement matched three cases to limit the number of returns required. Can we retain this? Fine to have a fall-through here, we do it in other places as well.

packages/protobuf/src/google/protobuf/duration_pb.ts
```diff
@@ -131,7 +130,7 @@ export class Duration extends Message<Duration> {
     if (typeof match[2] == "string") {
       const nanosStr = match[2] + "0".repeat(9 - match[2].length);
       this.nanos = parseInt(nanosStr);
-      if (longSeconds < protoInt64.zero) {
+      if (longSeconds < protoInt64.zero || match[1] === '-0') {
```
timostamm, line 133: I don't understand it :sweat_smile: Can you explain?

packages/protobuf-test/extra/msg-json-names.proto
```diff
@@ -22,9 +22,9 @@ message JsonNamesMessage {
     repeated string repeated_scalar_field = 2 [json_name = "repeatedScalarFieldJsonName"];
 
     string a = 3 [json_name = "sameJsonName"];
-    string b = 4 [json_name = "sameJsonName"];
+    string b = 4 [json_name = "b"];
```
timostamm, line 25: The fields 3 to 6 were there to show name clashes. If you simply change them to avoid conflicts, there is no longer any point in having them here. This means the code is really difficult to understand because it has no point. 

Please make sure that the name clashes shown by these fields are addressed by a compiler error or warning now, and remove them.

packages/protobuf-test/src/gen/ts/google/protobuf/descriptor_pb.ts
```diff
@@ -164,12 +164,21 @@ export class FileDescriptorProto extends Message<FileDescriptorProto> {
 
   /**
    * The syntax of the proto file.
-   * The supported values are "proto2" and "proto3".
+   * The supported values are "proto2", "proto3", and "editions".
```
timostamm, line 167: Are we already handling this in descriptor-set.ts and in protoplugin?

packages/protobuf-test/src/gen/js/google/protobuf/struct_pb.js
smaye81: I had this there for consistency with to/fromJson, but point taken.  I removed it from the above location.

packages/protobuf-test/src/gen/ts/google/protobuf/descriptor_pb.ts
smaye81: We were.  I also added a small commentary on this in the main description.

packages/protobuf/src/google/protobuf/duration_pb.ts
smaye81: Sure.  The regex above, if I'm following correctly, stores the whole number part in `match[1]` and the fractional part in `match[2]`.  In the case of a duration that is > -1 and < 0, then the whole number part (`match[1]`) will be `-0`.  

`BigInt` does not support the notion of a signed zero value (i.e. -0), so when we pass this value to `protoInt64.parse` the returned value is simply 0.  If we simply test `longSeconds` to be < `protoInt64.zero`, that won't work.  We also obviously can't check for this to be `<=`.  So we need a way to know if the whole number part is a string of `-0`.  

We could also test `longSeconds` to equal -0, but that seems error-prone to me over just evaluating an explicit string of `'-0'`.

smaye81: That works too.  Probably a little easier to read.  I just got a bit nervous with dealing with -0 as a number and it felt safer to compare a string.

timostamm: Good call, but thankfully Object.is has [well-defined behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#description).

timostamm (approved):

PR #398, opened 2023-02-23
Fix type declaration emitting when using NodeNext module resolution

### LINKED ISSUE #396: Type resolution for generated code struggles when using `NodeNext`
`The inferred type of 'runtime' cannot be named without a reference`

The generated code from `protoc-gen-es` currently causes problems with NodeNext module resolution. I'm not a tsconfig wizard (who really is though?) so there may be a misconfiguration on my side. However, `tsc` struggles with the generated `runtime` property on messages.

```typescript
  ...

  static readonly runtime = proto3; // This is where it errors with "The inferred type of 'runtime' cannot be named without a reference".
  static readonly typeName = "sf.substreams.v1.Clock";
  static readonly fields: FieldList = proto3.util.newFieldList(() => [
    { no: 1, name: "id", kind: "scalar", T: 9 /* ScalarType.STRING */ },
    { no: 2, name: "number", kind: "scalar", T: 4 /* ScalarType.UINT64 */ },
    { no: 3, name: "timestamp", kind: "message", T: Timestamp },
  ]);

  ...
```

Adding an explicit type hint like for FieldList below solves it, but I'm not sure if that's the route you want to take. If it is, I'm happy to whip up a PR for that.

I put together a simple reproduction here:  https://github.com/fubhy/protoc-gen-es-repro

fubhy:
Removes `--noEmit` when testing compilation with the various tsc versions and configures `declarations` and `declarationMap` to be emitted.

Adds a ProtoRuntime typehint to the `runtime` property on generated messages.

Fixes #396

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules
 /packages/*/dist
-/packages/protobuf-test/typescript/out
 /packages/protobuf-test/descriptorset.bin
 /packages/protoplugin-test/descriptorset.bin
 /.tmp
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 /packages/*/dist
+/packages/protobuf-test/typescript/out
 /packages/protobuf-test/descriptorset.bin
 /packages/protoplugin-test/descriptorset.bin
 /.tmp
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,7 @@ test-ts-compat: $(GEN)/protobuf-test node_modules
 		formatted=$$(echo "$${number}" | sed -r 's/[\.]/_/g'); \
 		dirname=packages/protobuf-test/typescript ; \
 		echo "Testing TypeScript `node_modules/ts$$formatted/bin/tsc --version`" ; \
-		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --noEmit || exit ; \
+		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --outDir $$dirname/out/$$formatted || exit ; \
 	done
 
 .PHONY: lint
--- a/Makefile
+++ b/Makefile
@@ -183,9 +183,9 @@ test-conformance: $(BIN)/conformance_test_runner $(BUILD)/protobuf-conformance
 test-ts-compat: $(GEN)/protobuf-test node_modules 
 	@for number in $(TS_VERSIONS) ; do \
 		formatted=$$(echo "$${number}" | sed -r 's/[\.]/_/g'); \
-		dirname=packages/protobuf-test/typescript ; \
+		dirname=packages/protobuf-test ; \
 		echo "Testing TypeScript `node_modules/ts$$formatted/bin/tsc --version`" ; \
-		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --outDir $$dirname/out/$$formatted || exit ; \
+		node_modules/ts$$formatted/bin/tsc -p $$dirname/typescript/tsconfig.$${formatted}.json --outDir $$dirname/dist/typescript/$$formatted || exit ; \
 	done
 
 .PHONY: lint
--- a/packages/protobuf-test/typescript/tsconfig.4_7_4.json
+++ b/packages/protobuf-test/typescript/tsconfig.4_7_4.json
@@ -12,6 +12,8 @@
     // compatible, set the skipLibCheck flag to false explicitly.
     "skipLibCheck": false,
     // To test forward-compatibility, set to NodeNext
-    "moduleResolution": "NodeNext"
+    "moduleResolution": "NodeNext",
+    "declaration": true,
+    "declarationMap": true
   }
 }
--- a/packages/protobuf-test/typescript/tsconfig.4_7_4.json
+++ b/packages/protobuf-test/typescript/tsconfig.4_7_4.json
@@ -13,6 +13,8 @@
     "skipLibCheck": false,
     // To test forward-compatibility, set to NodeNext
     "moduleResolution": "NodeNext",
+    // Certain errors are only triggered by actually emitting declaration files,
+    // see https://github.com/bufbuild/protobuf-es/pull/398
     "declaration": true,
     "declarationMap": true
   }
--- a/packages/protoc-gen-es/src/typescript.ts
+++ b/packages/protoc-gen-es/src/typescript.ts
@@ -101,7 +101,7 @@ function generateMessage(schema: Schema, f: GeneratedFile, message: DescMessage)
   f.print("  }");
   f.print();
   generateWktMethods(schema, f, message);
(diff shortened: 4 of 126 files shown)


packages/protobuf/src/index.ts
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+export type { ProtoRuntime } from "./private/proto-runtime.js";
```
fubhy, line 15: Not sure if you want to export like that from a "private" module.

Makefile
```diff
@@ -185,7 +185,7 @@ test-ts-compat: $(GEN)/protobuf-test node_modules
 		formatted=$$(echo "$${number}" | sed -r 's/[\.]/_/g'); \
 		dirname=packages/protobuf-test/typescript ; \
 		echo "Testing TypeScript `node_modules/ts$$formatted/bin/tsc --version`" ; \
-		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --noEmit || exit ; \
+		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --outDir $$dirname/out/$$formatted || exit ; \
```
timostamm, line 188: We already have the `dist` directory with ephemeral artifacts (this is not published, and will never be).

Can you change this to emit into this directory? Than we don't need to ignore a new one. 

```suggestion
		node_modules/ts$$formatted/bin/tsc -p $$dirname/tsconfig.$${formatted}.json --outDir packages/protobuf-test/dist/$$formatted || exit ; \
```

.eslintrc.cjs
```diff
@@ -8,6 +8,7 @@ module.exports = {
     node: true,
   },
   ignorePatterns: [
+    "packages/protobuf-test/typescript/out/**",
```
timostamm, line 11: We can avoid adding a new one, see comment on Makefile.

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 /packages/*/dist
+/packages/protobuf-test/typescript/out
```
timostamm, line 3: We can avoid adding a new one, see comment on Makefile.

.prettierignore
```diff
@@ -3,4 +3,5 @@
 /packages/*/dist
 /packages/*/src/gen
 /packages/protobuf/src/google
+/packages/protobuf-test/typescript/out
```
timostamm, line 6: We can avoid adding a new one, see comment on Makefile.

packages/protobuf-test/typescript/tsconfig.4_7_4.json
```diff
@@ -12,6 +12,8 @@
     // compatible, set the skipLibCheck flag to false explicitly.
     "skipLibCheck": false,
     // To test forward-compatibility, set to NodeNext
-    "moduleResolution": "NodeNext"
+    "moduleResolution": "NodeNext",
+    "declaration": true,
+    "declarationMap": true
```
timostamm, line 17: Can you add a comment here explaining why we set the option? 

```suggestion
    // Certain errors are only triggered by actually emitting declaration files, see https://github.com/bufbuild/protobuf-es/pull/398
    "declaration": true,
    "declarationMap": true
```

Please add the same comment to the other config files.

packages/protoc-gen-es/src/typescript.ts
```diff
@@ -101,7 +102,7 @@ function generateMessage(schema: Schema, f: GeneratedFile, message: DescMessage)
   f.print("  }");
   f.print();
   generateWktMethods(schema, f, message);
-  f.print("  static readonly runtime = ", protoN, ";");
+  f.print("  static readonly runtime: ", ProtoRuntime, " = ", protoN, ";");
```
timostamm, line 105: I noticed that declaration.ts generates `runtime: typeof proto3 = proto3` here. 

And this actually works for typescript.ts as well. This passes the tests: 

```suggestion
  f.print("  static readonly runtime: typeof ", protoN ," = ", protoN, ";");
```

I think this is preferable over `ProtoRuntime`. We will add another runtime soon, and `ProtoRuntime` would no longer fit, and we would have to export even more types.

Can you make this change? That means we don't actually have to export `ProtoRuntime` anymore, and we don't have to loop it through codegen-info.ts and runtime-imports.ts, which is great.

packages/protobuf/src/codegen-info.ts
```diff
@@ -63,24 +64,25 @@ export const codegenInfo: CodegenInfo = {
...
-    JsonReadOptions:      {typeOnly: true,  privateImportPath: "./json-format.js",   publicImportPath: packageName},
-    JsonWriteOptions:     {typeOnly: true,  privateImportPath: "./json-format.js",   publicImportPath: packageName},
-    JsonValue:            {typeOnly: true,  privateImportPath: "./json-format.js",   publicImportPath: packageName},
-    JsonObject:           {typeOnly: true,  privateImportPath: "./json-format.js",   publicImportPath: packageName},
-    protoInt64:           {typeOnly: false, privateImportPath: "./proto-int64.js",   publicImportPath: packageName},
-    ScalarType:           {typeOnly: false, privateImportPath: "./field.js",         publicImportPath: packageName},
-    MethodKind:           {typeOnly: false, privateImportPath: "./service-type.js",  publicImportPath: packageName},
-    MethodIdempotency:    {typeOnly: false, privateImportPath: "./service-type.js",  publicImportPath: packageName},
-    IMessageTypeRegistry: {typeOnly: true,  privateImportPath: "./type-registry.js", publicImportPath: packageName},
+    ProtoRuntime:         {typeOnly: true,  privateImportPath: "./private/proto-runtime.js",  publicImportPath: packageName},
```
timostamm, line 67: With `typeof protoN` (see typescript.ts), we don't need this!

packages/protobuf/src/index.ts
timostamm: Would be best to split into a ./proto-runtime.ts with just the interface. But with `typeof protoN` (see typescript.ts), we don't need this!

timostamm (commented):
Can confirm that https://github.com/bufbuild/protobuf-es/pull/398/commits/d99b803b26b57030c03fadede6846797088f8e10 fails the tests 👍 

Seems that `declarationMap` is not necessary to trigger, but I think it's a good call to leave it in. 

Can you make the simplification with `typeof protoN` and re-using the `dist` directory and add explanatory comments suggested below?

Makefile
fubhy: Agreed. I'm going to re-use the `dirname` variable there then though if you don't mind!

packages/protoc-gen-es/src/typescript.ts
fubhy: Ah yeah, that's much better.

fubhy:
@timostamm Thanks for the review. I force-pushed two commits on top of the test-failure demonstration commit to make it easier to re-review for you (considering all the noise with the generated files).

timostamm (approved):
Thank you for this fix!

PR #403, opened 2023-02-28
Upgrade @bufbuild/buf

smaye81:
This upgrades the Buf CLI npm package in all packages with Protobuf-ES.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5951,7 +5951,7 @@
     "packages/protobuf-bench": {
       "name": "@bufbuild/protobuf-bench",
       "dependencies": {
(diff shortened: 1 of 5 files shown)


timostamm (commented):
No changes to package-lock.json?

smaye81:
I ran `make` after these updates assuming it would update, but realized it only does `npm ci`.  One sec -- updating.

smaye81:
> I ran `make` after these updates assuming it would update, but realized it only does `npm ci`. One sec -- updating.

Done.

timostamm (approved):

PR #431, opened 2023-03-14
Enables exactOptionalPropertyTypes

dimitropoulos:
To quote from the TypeScript team lead:

> If we were to start TypeScript over again, we believe the behavior of strictOptionalProperties would be on by default
> https://github.com/microsoft/TypeScript/issues/44421

This [configuration](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) brings a pretty critical element to the type-soundness (critical because it drastically changes the behavior of operations that check for keys).  [SameValueZero](https://262.ecma-international.org/7.0/#sec-samevaluezero) checks remain unchanged, anyway, so there's basically no downside to having this rule enabled.  The [implementation](https://github.com/microsoft/TypeScript/pull/43947) from Anders provides some more context, as well.

Thankfully, this codebase was already pretty close to the mark, with only a few little minor things to update to increase strictness.  Once this PR merges, I'd like to do to same with the connect-es repo.

### DIFF
--- a/packages/protobuf-example/tsconfig.json
+++ b/packages/protobuf-example/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "target": "es2016",
     "strict": true,
+    "exactOptionalPropertyTypes": true,
     "moduleResolution": "Node"
   }
 }
--- a/packages/protobuf/src/create-descriptor-set.ts
+++ b/packages/protobuf/src/create-descriptor-set.ts
@@ -213,6 +213,10 @@ function addEnum(
   cart: Cart
 ): void {
   assert(proto.name, `invalid EnumDescriptorProto: missing name`);
+  const sharedPrefix = findEnumSharedPrefix(
+    proto.name,
+    proto.value.map((v) => v.name ?? "")
+  );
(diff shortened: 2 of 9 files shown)


dimitropoulos:
(also: note: https://github.com/bufbuild/protobuf-es/pull/422 was found by enabling this setting)

timostamm:
> If we were to start TypeScript over again, we believe the behavior of strictOptionalProperties would be on by default

I'd love that, but they didn't start over, and the rest of the type system doesn't work nicely with the option, requiring expressions like `...(idempotency !== undefined ? { idempotency } : {})`. 

I don't think anybody really likes such an expression, and we had a feature request to accept `| undefined` in the message initializer to avoid them (https://github.com/bufbuild/protobuf-es/pull/374). 

We should definitely support strictOptionalProperties once it becomes the default. Before that happens, we're better off avoiding the over-engineering.

dimitropoulos:
@timostamm that's very easy, all we need to do is change the type to reflect that a literal undefined is acceptable and your concern is addressed.  e90f7f15bbc4d7e30ed836538fa085e332246a43 :)

PR #436, opened 2023-03-24
Document how to use multiple plugin options

timostamm:
It is not obvious how to give multiple plugin options from our readme. This updates the example buf.gen.yaml to use a yaml list, and adds a comment.

This also updates the example to make use of `buf` on npmjs.com

### DIFF
--- a/packages/protoc-gen-es/README.md
+++ b/packages/protoc-gen-es/README.md
@@ -15,12 +15,17 @@ npm install @bufbuild/protobuf
 ```
 
 We use peer dependencies to ensure that code generator and runtime library are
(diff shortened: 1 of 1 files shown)


packages/protoc-gen-es/README.md
```diff
@@ -32,26 +37,15 @@ plugins:
...
-
-```json
-{
-  "name": "your-package",
-  "version": "1.0.0",
-  "scripts": {
-    "generate": "buf generate"
-  },
-  // ...
-}
```
timostamm, line 48: We don't need the npm script anymore with @bufbuild/buf and npx.

srikrsna-buf (approved):

PR #451, opened 2023-04-03
Upgrade to TypeScript 5.0

smaye81:
This upgrades Protobuf-ES to use TypeScript 5.0.3. 

The only major change needed for the upgrade was to accommodate the new [`verbatimModuleSyntax`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax) flag. This required removing the usage of `importsNotUsedAsValues` from our main tsconfig and applying the new flag to any build commands generating ESM output. Note that we can't use this in our main `tsconfig.base.json` because it isn't permitted when the module output is CJS.

For consistency, this flag was used in all the CLI commands for generating ESM. The only place where it was put into the actual `tsconfig.json` was for the `protoplugin-example`.

Finally, this also removes 5.0.0 from our TS compatibility tests in favor of the last major / minor version of 4.9.5.


timostamm (approved):
It's really unfortunate that the new flag complicates dual-package builds. I don't see any substantially better solution with `tsc`. LGTM!

PR #459, opened 2023-04-17
Improve binary read performance

kbongort:
This PR makes a couple of improvements to binary read performance:

1. One of the slowest things during a read with many messages is initializing a BinaryReader for each message with its DataView and TextDecoder. This PR adds a `messageField()` method to BinaryReader that reads a size-delimited message field without constructing a new BinaryReader. (`Message.fromBinary` is refactored to share code). A knock-on effect of this change is that we pass `options` along without repeated calls to `makeReadOptions`, which otherwise needlessly spreads `options` into `readDefaults` when `options` is already equivalent to `readDefaults`.
1. readScalar was calling `reader[method]()` which seems to incur an implicit call to .bind() on the reader method (i.e. it is equivalent to `reader[method].bind(reader)()`. By using an inline switch statement we can avoid that indirection and use a faster path.

I put together a simple benchmark (https://github.com/kbongort/protobuf-es/pull/1) using the following message definitions:
```
message User {
  string first_name = 1;
  string last_name = 2;
  bool active = 3;
  User manager = 4;
  repeated string locations = 5;
  map<string, string> projects = 6;
  int32 number = 7;
}

message UserList {
  repeated User users = 1;
}
```

The benchmark uses both a more complete version of the User message and one with only a few fields set:
```
const userMessage = new User({
  firstName: "Jane",
  lastName: "Doe",
  active: true,
  manager: { firstName: "Jane", lastName: "Doe", active: false },
  locations: ["Seattle", "New York", "Tokyo"],
  projects: { foo: "project foo", bar: "project bar" },
});

const smallUserMessage = new User({
  manager: { active: true },
  number: 2,
});
```

After warming up the cache by reading a serialized userMessage 1000 times, it measures the time to read `userMessage`, to read a UserList of 1000 `userMessage`s, and to read a UserList of 1000 `smallUserMessage`s. The amount of time it takes to call `JSON.parse()` on the JSON version of the message is provided as a point of comparison. Without the changes in this PR, the results are:
```
Benchmark: Single user
	Time to parse JSON: 1135 µs
	Time to parse from binary: 3548 µs
	Ratio: 312%
Benchmark: List of 1000 users
	Time to parse JSON: 920 ms
	Time to parse from binary: 2957 ms
	Ratio: 321%
Benchmark: List of 1000 small users
	Time to parse JSON: 260 ms
	Time to parse from binary: 923 ms
	Ratio: 355%
```

After the changes in this PR, the results are:
```
Benchmark: Single user
	Time to parse JSON: 1088 µs
	Time to parse from binary: 2755 µs
	Ratio: 253%
Benchmark: List of 1000 users
	Time to parse JSON: 921 ms
	Time to parse from binary: 1729 ms
	Ratio: 188%
Benchmark: List of 1000 small users
	Time to parse JSON: 255 ms
	Time to parse from binary: 237 ms
	Ratio: 93%
```

The three test cases show speedups of 29%, 71%, and 290% respectively. The effect of (1) and is especially pronounced when there are many small messages, as in the last benchmark test.

### DIFF
--- a/packages/protobuf/src/private/binary-format-common.ts
+++ b/packages/protobuf/src/private/binary-format-common.ts
@@ -145,14 +145,14 @@ export function makeBinaryFormatCommon(): Omit<BinaryFormat, "writeMessage"> {
             if (repeated) {
               // safe to assume presence of array, oneof cannot contain repeated values
               (target[localName] as any[]).push(
-                messageType.fromBinary(reader.bytes(), options)
+                reader.messageField(new messageType(), options)
               );
             } else {
               if (target[localName] instanceof Message) {
-                target[localName].fromBinary(reader.bytes(), options);
+                reader.messageField(target[localName], options);
               } else {
-                target[localName] = messageType.fromBinary(
-                  reader.bytes(),
+                target[localName] = reader.messageField(
+                  new messageType(),
                   options
                 );
                 if (
@@ -235,8 +235,34 @@ function readMapEntry(
 }
 
 function readScalar(reader: IBinaryReader, type: ScalarType): any {
-  let [, method] = scalarTypeInfo(type);
-  return reader[method]();
+  switch (type) {
+    case ScalarType.STRING:
+      return reader.string();
+    case ScalarType.BOOL:
+      return reader.bool();
+    case ScalarType.DOUBLE:
+      return reader.double();
+    case ScalarType.FLOAT:
+      return reader.float();
+    case ScalarType.INT32:
+      return reader.int32();
+    case ScalarType.INT64:
+      return reader.int64();
+    case ScalarType.UINT64:
+      return reader.uint64();
+    case ScalarType.FIXED64:
+      return reader.fixed64();
+    case ScalarType.BYTES:
+      return reader.bytes();
+    case ScalarType.FIXED32:
+      return reader.fixed32();
+    case ScalarType.SFIXED32:
+      return reader.sfixed32();
+    case ScalarType.SFIXED64:
+      return reader.sfixed64();
+    case ScalarType.SINT64:
+      return reader.sint64();
+  }
 }
 
 export function writeMapEntry(
--- a/packages/protobuf/src/binary-encoding.ts
+++ b/packages/protobuf/src/binary-encoding.ts
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import type { BinaryReadOptions } from "./binary-format.js";
 import {
   varint32read,
   varint32write,
   varint64read,
   varint64write,
 } from "./google/varint.js";
+import { Message } from "./message.js";
 import { assertFloat32, assertInt32, assertUInt32 } from "./private/assert.js";
 import { protoInt64 } from "./proto-int64.js";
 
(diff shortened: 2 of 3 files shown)


packages/protobuf/src/private/binary-format-common.ts
```diff
@@ -235,8 +235,38 @@ function readMapEntry(
 }
 
 function readScalar(reader: IBinaryReader, type: ScalarType): any {
-  let [, method] = scalarTypeInfo(type);
-  return reader[method]();
```
newyankeecodeshop, line 239: It might be lower impact to use `reader[method].call(reader)`

kbongort: That doesn't work either. Here's a flame graph with the switch statement:
image

And with .call():
image

Note that in the top chart, `string()` is >90% of the time spent in `readScalar()`. In the lower chart, around 1/3 of the time is spent in V8 something-something (I wish I could get more detail on what's going on there, but it's out of my depth and Google hasn't been helpful). The benchmark also comes out significantly worse.

newyankeecodeshop: Oh I suspect the performance issue is the use of `scalarTypeInfo()`, which requires allocating an array to hold the return values, and does `const method = ScalarType[type].toLowerCase()` to determine the method. ToLowerCase is probably consuming some additional time.

kbongort: It's not `scalarTypeInfo()`; you can see that function in the second flame graph:

image

My PR description includes a link to my benchmarking code, so I'd encourage you to clone this PR and that one, and merge them together to test any theories you have. I'm using https://clinicjs.org/flame/ to make flame graphs and invoking the benchmark with `clinic flame -- node --no-turbo-inlining dist/esm/benchmark.js` in order to keep functions separate in the result.

smaye81:
Hi @kbongort appreciate your patience on this. I know we spoke offline last week. We have been gathering some baseline metrics on our performance to see all the areas that need addressed. I can confirm that this does improve binary read performance by an average of ~2.5x.  

However, one concern is the change to modify the `IBinaryReader` interface. This is an interface we expose from the library and changing it will result in breaking changes. Are you up for potentially modifying the PR so that these changes can be made in a non-breaking way without affecting the interface contract?

kbongort:
> Hi @kbongort appreciate your patience on this. I know we spoke offline last week. We have been gathering some baseline metrics on our performance to see all the areas that need addressed. I can confirm that this does improve binary read performance by an average of ~2.5x.
> 
> However, one concern is the change to modify the `IBinaryReader` interface. This is an interface we expose from the library and changing it will result in breaking changes. Are you up for potentially modifying the PR so that these changes can be made in a non-breaking way without affecting the interface contract?

What would you suggest?

smaye81:
Let me talk it over with the team. We know that the `fromBinary > readMessage` path is a bottleneck, but I'm not sure if there is a good way to reuse the `BinaryReader` like you're doing without breaking one of the interfaces.

In the meantime, would you mind splitting out the two improvements (`BinaryReader` vs. `readScalar`). I think we can probably get the latter implemented and at least get a quick win.

kbongort:
> Let me talk it over with the team. We know that the `fromBinary > readMessage` path is a bottleneck, but I'm not sure if there is a good way to reuse the `BinaryReader` like you're doing without breaking one of the interfaces.
> 
> In the meantime, would you mind splitting out the two improvements (`BinaryReader` vs. `readScalar`). I think we can probably get the latter implemented and at least get a quick win.

What type of breakage are you concerned with, exactly? This does add a couple of methods to IBinaryReader. Are you worried that clients may have IBinaryReader implementations that will no longer conform to the interface, or something else?

kbongort:
I've updated the PR to not modify IBinaryReader. Since the new methods are implemented strictly in terms of existing ones, it's possible to refactor them into static utility methods. So that's what I did.

kbongort:
@smaye81 PTAL – Let me know if you have any further concerns, thanks!

kbongort:
Hi Steve,

I believe that I've addressed your concerns with this PR. Can you please
take another look?

Thanks,
Ken

smaye81:
Hi @kbongort -- yes, apologies for the delay. we have been a bit heads-down on some other work at the moment, but will take a look at this hopefully this week.  thanks again for the PR.

cc @timostamm

PR #460, opened 2023-04-17
Add additional information to docs for serialization

smaye81:
This expands a bit on the documentation for serialization. It explains the individual methods in a bit more detail and lists their respective options for customizing the serialization behavior.

### DIFF
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -237,7 +237,8 @@ The `toBinary` function also accepts an options object which can be used to cust
 
 ```typescript
 {
-   // Indicates whether to include unknown fields in the serialized output. Defaults to `true`.
+   // Indicates whether to include unknown fields in the serialized output. 
+   // Defaults to `true`.
    // For more details see https://developers.google.com/protocol-buffers/docs/proto3#unknowns
    writeUnknownFields?: boolean;
    
@@ -258,7 +259,8 @@ The `fromBinary` function also accepts an options object which can be used to cu
 
 ```typescript
 {
-   // Indicates whether to retain unknown fields during parsing and include them in the serialized output. Defaults to `true`.
+   // Indicates whether to retain unknown fields during parsing and include them 
+   // in the serialized output. Defaults to `true`.
    // For more details see https://developers.google.com/protocol-buffers/docs/proto3#unknowns
    readUnknownFields?: boolean;
    
@@ -279,18 +281,23 @@ The `toJson` function also accepts an options object which can be used to custom
 
 ```typescript
 {
-   // Indicates whether to emit fields with default values. Fields with default values are omitted by default in proto3 JSON output. 
-   // This option overrides this behavior and outputs fields with their default values.
+   // Indicates whether to emit fields with default values. 
+   // Fields with default values are omitted by default in proto3 JSON output. 
+   // This option overrides this behavior and outputs fields with 
+   // their default values.
    emitDefaultValues?: boolean;
    
    // Indicates whether to emit enum values as integers instead of strings.
-   // The name of an enum value is used by default in JSON output. An option may be provided to use the numeric value of the enum value instead.
+   // The name of an enum value is used by default in JSON output. An option 
+   // may be provided to use the numeric value of the enum value instead.
    enumAsInteger?: boolean;
    
-   // Whether to use the proto field name instead of converting to lowerCamelCase.
-   // By default the proto3 JSON printer should convert the field name to lowerCamelCase and use that as the JSON name. 
-   // An implementation may provide an option to use the proto field name as the JSON name instead. 
-   // Proto3 JSON parsers are required to accept both the converted lowerCamelCase name and the proto field name.
+   // Whether to use the proto field name instead of converting 
+   // to lowerCamelCase. By default the proto3 JSON printer should 
+   // convert the field name to lowerCamelCase and use that as the JSON name. 
+   // An implementation may provide an option to use the proto field name as 
+   // the JSON name instead. Proto3 JSON parsers are required to accept both 
+   // the converted lowerCamelCase name and the proto field name.
    useProtoFieldName?: boolean;
 }
 ```
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -12,7 +12,7 @@ provided by the library.
   - [Accessing oneof groups](#accessing-oneof-groups)
   - [Cloning messages](#cloning-messages)
   - [Comparing messages](#comparing-messages)
-- [Serialization and parsing](#serialization-and-parsing)
+- [Serialization](#serialization)
 - [Using enumerations](#using-enumerations)
 - [Well-known types](#well-known-types)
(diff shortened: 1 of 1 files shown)


smaye81:
@timostamm I wasn't sure of how you'd like to present the options objects, so open to suggestions on a layout.

timostamm (commented):
Some users have been surprised by our toJSON() implementation. We have JSDoc [here](https://github.com/bufbuild/protobuf-es/blob/772930c9852b7b50fb2a4e95b646462fbbfd3bf1/packages/protobuf/src/message.ts#L135-L154). Can we make this easier to understand and also call it out here?

smaye81:
Incorporated your feedback @timostamm. Let me know what you think.

docs/runtime_api.md
```diff
@@ -221,16 +227,20 @@ User.equals(user, null); // false
 User.typeName; // docs.User
 ```
 
-### Serializing messages
+## Serializing messages
 
-Serializing to the binary format is very straight-forward:
+All messages provide functions for serializing and parsing data between the
```
timostamm, line 232:
```suggestion
All messages provide methods for serializing and parsing data between the 
```

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
 User.fromBinary(protoBase64.dec(base64));
 ```
 
-Serializing to the JSON format is equally simple:
+### `toBinary`
+
+The `toBinary` function accepts an options object which can be used to
```
timostamm, line 255:
```suggestion
The `toBinary` method accepts an options object which can be used to 
```

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
...
 const json = user.toJson();
 User.fromJson(json);
 ```
 
-But the result will be a [JSON value][src-json-value] – a primitive JavaScript that can 
-be converted to a JSON string with the built-in function `JSON.stringify()`. For 
-convenience, we also provide methods that include the stringify step:
+### `toJson`
+
+The `toJson` function also accepts an options object which can be used to
```
timostamm, line 290:
```suggestion
The `toJson` method also accepts an options object which can be used to 
```

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
...
+when `JSON.stringify` is used. For more information, see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior).
+However, you should never need to invoke this function, which is why
+it is marked as protected. 
+
+As a rule of thumb, use `JSON.stringify` or `toJsonString` for stringified JSON. 
+If actual JSON is desired, use `toJson`.
+
+### `fromJson`
 
+The `fromJson` function also accepts an options object for customizing the
```
timostamm, line 346:
```suggestion
The `fromJson` method also accepts an options object for customizing the 
```

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
...
+   A convenience property for the space option to `JSON.stringify`.
+   It indicates the number of space characters to be used as indentation, clamped to 10 
+   (that is, any number greater than 10 is treated as if it were 10). 
+   Values less than 1 indicate that no space should be used.
+   For more information, see the [`JSON.stringify` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify).
+
+Note that messages also contain a protected `toJSON` function, which is invoked
+when `JSON.stringify` is used. For more information, see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior).
+However, you should never need to invoke this function, which is why
+it is marked as protected.
```
timostamm, line 339: My first question would be _why_ we have a toJSON, and what it does. What do you think about:

```suggestion
Note that all message have a `toJSON` method as well. If you serialize a message with `JSON.stringify()`, it will automatically call this method. By providing `toJSON`, we make sure that messages are always serialized to the well-defined protobuf JSON format. Otherwise, `JSON.stringify()` would crash on BigInt values, and would not serialize `oneof`, enumerations, and other types correctly. Serializing a message with `JSON.stringify()` is equivalent to `message.toJsonString({emitDefaultValues: true})`.
```

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
...
+// const json = user.toJson();
+// const jsonStr = JSON.stringify(json);
+
+const jsonStr = user.toJsonString();
+```
+
+Since `toJsonString` calls `toJson` under the hood, it accepts the same options 
+object as `toJson` as well as additional options for the stringify step:
+
+- `emitDefaultValues?: boolean`<br/>
```
timostamm, line 329: Shouldn't that be "prettySpaces"?

docs/runtime_api.md
```diff
@@ -240,32 +250,134 @@ const base64: string = protoBase64.enc(bytes)
...
+   A convenience property for the space option to `JSON.stringify`.
+   It indicates the number of space characters to be used as indentation, clamped to 10 
+   (that is, any number greater than 10 is treated as if it were 10). 
+   Values less than 1 indicate that no space should be used.
+   For more information, see the [`JSON.stringify` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify).
```
timostamm, line 334: Very wordy. WDYT about:

```suggestion
   A convenience property for the `space` option to `JSON.stringify`, which controls indentation for prettier output. 
   See the [`JSON.stringify` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#parameters).
```

timostamm (commented):
This is great, but the methods look out of place in the ToC (see screenshot below). We don't do this for any of the other methods or for other objects, and I think we should stick with the conversational style, with using topic for headlines, and explaining methods via examples. 

What do you think about having just two sections "JSON" and "Binary" instead.

Screenshot 2023-04-20 at 20 43 00

docs/runtime_api.md
smaye81: Ugh yes, fixed.

docs/runtime_api.md
smaye81: Sure. There's two additional points that I think we should make though:

- By providing `toJSON`, we make sure that messages are always serialized to the well-defined protobuf JSON format _because it calls our `toJson` implementation_. I don't want to give the impression that just because we declared the method, we automatically get well-defined protobuf JSON.
- Users shouldn't need to use this method directly.

I augmented what you have with these two points. Let me know what you think.

timostamm (approved):

PR #467, opened 2023-04-24
Add additional context about JSON

smaye81:
This adds an FAQ about JSON to our FAQ page. It also adds some additional context about initializer objects when constructing messages.


docs/faq.md
```diff
@@ -115,6 +115,61 @@ In the above code, `plainFoo` will be a type with only its fields and `oneOf` gr
 
 Great segue!  Our [docs](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#advanced-typescript-types) provide a good explanation for their usage and example use cases.
 
+### Why doesn't Protobuf-ES simply generate interfaces for the JSON representation?
```
timostamm, line 118: This is great, but can we list the reasons a bit more concisely? For example:

1. Protobuf has several types that do not have a corresponding JSON type
	- list them here
2. There is no single JSON representation - implementation can provide serialization options
	- give examples here
3. JSON parsing is lax. For example, parsers must accept fields with their lowerCamelCase name, or the original Protobuf field name.
	- give more examples here

As a result, there is no single JSON type for a protobuf message.

Issue #494, opened 2023-05-31
Incompatibility between `protobufjs` and `@bufbuild/protobuf`


fubhy:
It's currently not possible to load a `.proto` file with `protobufjs` and turn it into a descriptor with `@bufbuild/protobuf`.

So far, debugging lead me to this:

1. `protobufjs` does not resolve field type names for message type references to their fully qualified names.
2. `@bufbuild/protobuf` however expects those to be fully resolved in `newFields` https://github.com/bufbuild/protobuf-es/blob/ae29a60c7c6b396da73c14bf858758884845843a/packages/protobuf/src/create-descriptor-set.ts#L537

This leads to `error: invalid FieldDescriptorProto: type_name <TYPE NAME> not found`

Even adding the fully qualified names in the source `.proto` doesn't resolve it which leads me to believe that `protobufjs` actually strips those (expecting them to be resolved by parent lookup in the consumer?)

timostamm:
The "JSON descriptors" protobuf.js uses and produces are conceptually similar to the google/protobuf/descriptor.proto messages, but they are not the same. Unfortunately, they only contain a subset of the information in descriptor.proto, and the format is unique to protobuf.js.

[protobufjs/ext/descriptor](https://github.com/protobufjs/protobuf.js/tree/master/ext/descriptor) is an experimental extension to convert the protobuf.js descriptors to protobuf messages.

fubhy:
I see. Are you aware of any alternatives by chance? I'm trying to load raw `.proto` files at runtime and use them with `@bufbuild/proto` if at all possibe.

fubhy:
I fear I might_ have to resort to `exec` with `buf build` to produce the file descriptor set :-/

timostamm:
I'm not aware of a really awesome solution. I would try to avoid going from protobuf.js descriptors to descriptor.proto to @bufbuild/protobuf messages. Because protobuf.js descriptors are a subset of descriptor.proto, some data needs to be synthesized to produce descriptor.proto, and this is likely to cause problems. I have a hunch that relative references in type names are just the tip of the ice berg.

@bufbuild/protobuf messages only need a subset of the information in descriptor.proto. protobuf.js descriptor _might_ have all necessary data to go from them to @bufbuild/protobuf directly. But I would take a closer look first, especially regarding syntax. (@bufbuild/protobuf makes a distinction between proto2 and proto3 that protobuf.js might just not care about.)

[...]

PR #495, opened 2023-05-31
fix: resolve field type name in file descriptor

fubhy:
Ref: https://github.com/bufbuild/protobuf-es/issues/494

This is merely meant to demonstrate the problem I'm facing. I'm absolutely not sure if this needs fixing here or in `protobufjs`. Are fields referencing local types supposed to be fully resolved to their fully qualified names in a file descriptor? It seems that `protobufjs` thinks they shouldn't be because it apparently strips them even if you provide them in the source `.proto` manually.

### DIFF
--- a/packages/protobuf/src/create-descriptor-set.ts
+++ b/packages/protobuf/src/create-descriptor-set.ts
@@ -534,7 +534,21 @@ function newField(
           ...getMapFieldTypes(mapEntry),
         };
       }
-      const message = cart.messages.get(trimLeadingDot(proto.typeName));
+
+      const trimmedName = trimLeadingDot(proto.typeName);
+      let message = cart.messages.get(trimmedName);
+      if (message === undefined && !trimmedName.includes('.')) {
+        const parts = parent.typeName.split('.');
+        while (parts.length > 0) {
+          const parentName = parts.join('.');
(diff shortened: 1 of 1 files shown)


timostamm:
> Are fields referencing local types supposed to be fully resolved to their fully qualified names in a file descriptor?

From our documentation about [Compilation and Descriptors](https://protobuf.com/docs/descriptors#type-references):

> Type references are found in various fields in descriptors, all mapping to a TypeName production in the grammar. The value stored in the descriptors may initially be a relative reference, as that may be what appears in the source file. But a properly compiled file descriptor will undergo a link step, wherein all relative references are resolved.
> 
> At the end of this process, all relative references in a descriptor should be replaced with fully-qualified references, which include a leading dot (.).

fubhy:
Gotcha. So I'll likely have to do a preprocessing pass where I take the protobufjs descriptor and "fix" it before using it with @bufbuild/protobuf. Unless you are aware of any alternatives? ;-)

timostamm:
> So I'll likely have to do a preprocessing pass where I take the protobufjs descriptor and "fix" it 

Definitely an option, but not without caveats. See https://github.com/bufbuild/protobuf-es/issues/494#issuecomment-1569954634

PR #510, opened 2023-06-21
Failing for toJson functionality for value of null

dilipvamsi:
I was extending the message created from protos to create an Entity in typeorm which returns the values as null for nullable fields.
As null was not handled. It's throwing error saying **Cannot read properties of null (reading 'toJson')**

### DIFF
--- a/packages/protobuf/src/private/json-format-proto2.ts
+++ b/packages/protobuf/src/private/json-format-proto2.ts
@@ -33,6 +33,7 @@ export function makeJsonFormatProto2(): JsonFormat {
       value: any,
       options: JsonWriteOptions
     ): JsonValue | undefined {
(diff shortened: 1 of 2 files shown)


smaye81:
Hey @dilipvamsi appreciate the PR, but I don't think we'd want to land this. This wouldn't be the right place to make this check, plus it only fixes the JSON format. 

Typically, the general practice for modeling your DB schema is to create your own plugin which generates the code mirroring your DB structure. This allows you to generate your DB code separately and doesn't tie your API schema to the DB. We have extensive docs [here](https://github.com/bufbuild/protobuf-es/blob/main/docs/writing_plugins.md) for writing your own plugin.

Going to close this, but if you need any other help with the plugin, feel free to reach out.

PR #511, opened 2023-06-27
Add `toPlainMessage`

srikrsna-buf:
Add `toPlainMessage` to convert `Message` objects to their `PlainMessage` variants. 

Helps in cases like https://github.com/bufbuild/connect-es/issues/505.

### DIFF
--- a/packages/protobuf/src/to-plain-message.ts
+++ b/packages/protobuf/src/to-plain-message.ts
@@ -1,0 +1,85 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* eslint-disable @typescript-eslint/no-explicit-any,@typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-member-access,@typescript-eslint/no-unsafe-return,@typescript-eslint/no-unsafe-argument,no-case-declarations */
+
+import { Message } from "./message.js";
+import type { AnyMessage, PlainMessage } from "./message.js";
+
+/**
+ * toPlainMessage returns a new object by striping
+ * all methods from a message, leaving only fields and
+ * oneof groups. It is recursive, meaning it applies this
+ * same logic to all nested message fields as well.
+ */
+export function toPlainMessage<T extends Message<T>>(
+  message: T
+): PlainMessage<T> {
+  if (typeof structuredClone === "function") {
+    return structuredClone(message) as PlainMessage<T>;
+  }
+  const type = message.getType();
+  const source = message as AnyMessage;
+  const target = {} as PlainMessage<AnyMessage>;
+  for (const member of type.fields.byMember()) {
+    const localName = member.localName;
+    switch (member.kind) {
+      case "scalar":
+      case "enum":
+        target[localName] = source[localName];
+        break;
+      case "message":
+        if (member.repeated) {
+          target[localName] = (source[localName] as any[]).map((val) =>
+            toPlainMessage(val)
+          );
+        } else if (source[localName] !== undefined) {
+          target[localName] = toPlainMessage(source[localName]);
+        } else {
+          target[localName] = undefined;
+        }
+        break;
+      case "oneof":
+        const _case = source[localName].case;
+        if (_case === undefined) {
+          target[localName] = { case: undefined };
+          break;
+        }
+        let value = source[localName].value;
+        const sourceField = member.findField(_case);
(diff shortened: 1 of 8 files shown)


packages/protobuf/src/to-plain-message.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+  const type = message.getType();
+  const source = message as AnyMessage;
+  const target = {} as PlainMessage<AnyMessage>;
+  for (const member of type.fields.byMember()) {
+    const localName = member.localName;
+    switch (member.kind) {
+      case "scalar":
+      case "enum":
+        target[localName] = source[localName];
```
timostamm, line 40: This will keep the reference for repeated fields (and for Uint8Array), where `structuredClone` will create a copy.  I don't think it is mandatory to guarantee a deep clone, but the behavior should be consistent. So the options are:
1. deep clone like structuredClone
2. remove structuredClone, and always use our implementation

It is just a matter of time that structuredClone becomes widely available, so option 1 has the advantage that we can drop our implementation at some point in the future, which is great. 

So I tend to think that option 1 is the best choice, unless there is a significant performance penalty with structuredClone... 

With a naive test, they look pretty much identical, so let's go with option 1:

```
### toPlainMessage
protobuf_test_messages.proto3.TestAllTypesProto3 with toPlainMessage) x 13,217 ops/sec ±0.15% (98 runs sampled)
protobuf_test_messages.proto3.TestAllTypesProto3 with structuredClone() x 13,259 ops/sec ±0.17% (98 runs sampled)
spec.RepeatedScalarValuesMessage with toPlainMessage) x 255,900 ops/sec ±0.51% (96 runs sampled)
spec.RepeatedScalarValuesMessage with structuredClone() x 257,787 ops/sec ±0.50% (95 runs sampled)
```

packages/protobuf/src/to-plain-message.ts
```diff
@@ -0,0 +1,85 @@
...
+        break;
+      case "message":
+        if (member.repeated) {
+          target[localName] = (source[localName] as any[]).map((val) =>
+            toPlainMessage(val)
+          );
+        } else if (source[localName] !== undefined) {
+          target[localName] = toPlainMessage(source[localName]);
```
timostamm, line 48: Can you add a test that uses [WrappersMessage](https://github.com/bufbuild/protobuf-es/blob/f955b2913d3186c31cbd129f8eb5f434af456c8b/packages/protobuf-test/extra/wkt-wrappers.proto#L20)?

It might be helpful too peek at [Message.clone](https://github.com/bufbuild/protobuf-es/blob/f955b2913d3186c31cbd129f8eb5f434af456c8b/packages/protobuf/src/private/util-common.ts#L181-L227).

timostamm (commented):
Looks great, but I don't think we're handling wrapper messages correctly yet, and we should keep behavior consistent whether structuredClone is available or not.

packages/protobuf/src/to-plain-message.ts
timostamm: Looks like out implementation is up to 10x faster:

```
### toPlainMessage
protobuf_test_messages.proto3.TestAllTypesProto3 with toPlainMessage) x 27,494 ops/sec ±0.50% (94 runs sampled)
protobuf_test_messages.proto3.TestAllTypesProto3 with structuredClone() x 13,168 ops/sec ±0.21% (95 runs sampled)
spec.RepeatedScalarValuesMessage with toPlainMessage) x 2,239,265 ops/sec ±0.14% (100 runs sampled)
spec.RepeatedScalarValuesMessage with structuredClone() x 255,246 ops/sec ±0.53% (98 runs sampled)
```

<details>
    <summary>payloads</summary>

protobuf_test_messages.proto3.TestAllTypesProto3:
```json
{"optionalNestedMessage":{"a":123,"corecursive":{"optionalNestedMessage":{"a":456,"corecursive":{"optionalNestedMessage":{}}}}},"repeatedNestedMessage":[{"a":123,"corecursive":{"repeatedNestedMessage":[{"a":456,"corecursive":{"repeatedNestedMessage":[{}]}}]}}],"mapStringString":{"a":"A"},"oneofNestedMessage":{}}
```

spec.RepeatedScalarValuesMessage:
```json
{"doubleField":[0.75,0,1],"floatField":[0.75,-0.75],"int64Field":["-1","-2"],"uint64Field":["1","2"],"int32Field":[-123,500],"fixed64Field":["1","99"],"fixed32Field":[123,999],"boolField":[true,false,true],"stringField":["hello","world"],"bytesField":["aGVsbG8gd29ybGQ="],"uint32Field":[123,123],"sfixed32Field":[-123,-123,-123],"sfixed64Field":["-1","-2","100"],"sint32Field":[-1,-2,999],"sint64Field":["-1","-99","99"]}
```
</details>

srikrsna-buf: Is this also a function of runtime? Or we can safely assume we will always be faster than `structuredClone`

timostamm: AFAIK the serialization is implemented natively in V8. It seems likely that our implementation will be faster for the foreseeable future. Still think it's useful to have the same or very similar cloning behavior in our implementation, which I believe we do now.

Leaving it up to you whether to always use our implementation, or only use it as a fallback. Seems like it would be faster and simpler to always use ours...

PR looks great to me.

srikrsna-buf: Removed it, no point in having it if we are faster, users can always directly use that if they don't want to include this chunk of code.

smaye81 (approved):

PR #533, opened 2023-07-25
Allow `number[]` for `bytes` fields in constructor

srikrsna-buf:
Follow up work to #511. This doesn't change the types but loosens the runtime to also accept `number[]` alongside `Uint8Array` values for `bytes` fields.

### DIFF
--- a/packages/protobuf/src/private/util-common.ts
+++ b/packages/protobuf/src/private/util-common.ts
@@ -80,8 +80,8 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
                 let target = s[localName];
                 if (member.V.T === ScalarType.BYTES) {
                   target = {};
-                  for (const k of Object.keys(s[localName])) {
-                    target[k] = toU8Arr(s[localName][k] as ArrayLike<number>);
+                  for (const [k, v] of Object.entries(s[localName])) {
+                    target[k] = toU8Arr(v as ArrayLike<number>);
                   }
                 }
                 Object.assign(t[localName], target);
--- a/packages/protobuf/src/private/util-common.ts
+++ b/packages/protobuf/src/private/util-common.ts
@@ -213,7 +213,7 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
         const source = (message as AnyMessage)[member.localName];
         let copy: any;
         if (member.repeated) {
-          copy = (source as any[]).map((e) => cloneSingularField(e));
+          copy = (source as any[]).map(cloneSingularField);
         } else if (member.kind == "map") {
           copy = any[member.localName];
           for (const [key, v] of Object.entries(source)) {
--- a/packages/protobuf/src/private/util-common.ts
+++ b/packages/protobuf/src/private/util-common.ts
@@ -79,12 +79,10 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
               case "enum":
                 let target = s[localName];
                 if (member.V.T === ScalarType.BYTES) {
-                  target = Object.fromEntries(
-                    Object.entries(target).map(([k, v]) => [
-                      k,
-                      toU8Arr(v as ArrayLike<number>),
-                    ])
-                  );
+                  target = {};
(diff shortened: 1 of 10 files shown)


packages/protobuf/src/private/util-common.ts
```diff
@@ -55,18 +54,48 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
...
+                target = new Uint8Array(target);
+              } else if (member.repeated) {
+                const val = [];
+                for (const v of target) {
+                  val.push(v instanceof Uint8Array ? v : new Uint8Array(v));
+                }
+                target = val;
+              }
+            }
+            t[localName] = target;
```
timostamm, line 81: WDYT about introducing a function that converts to Uint8Array if necessary? 


```ts
function toU8Arr(input: ArrayLike<number>): Uint8Array {
  if (input instanceof Uint8Array) {
    return input;
  }
  return new Uint8Array(input);
}
```

This removes the if-else code paths from the call sites. Here, it would allow us to use a ternary operator:

```suggestion
            let target = s[localName];
            if (member.T === ScalarType.BYTES) {
              target = member.repeated ? target.map(toU8Arr) : toU8Arr(target);
            }
            t[localName] = target;
```

I think other call-sites would also benefit, with slightly easier to read code and smaller bundle size.

packages/protobuf/src/private/util-common.ts
```diff
@@ -55,18 +54,39 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
...
+                let target = s[localName];
+                if (member.V.T === ScalarType.BYTES) {
+                  target = Object.fromEntries(
+                    Object.entries(target).map(([k, v]) => [
+                      k,
+                      toU8Arr(v as ArrayLike<number>),
+                    ])
+                  );
+                }
+                Object.assign(t[localName], target);
```
timostamm, line 89: We cannot use Object.entries unfortunately. Our target is ES2017, and the method was only added in ES2019.

We should investigate why this didn't raise an error.

packages/protobuf/src/private/util-common.ts
```diff
@@ -186,19 +213,19 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
         const source = (message as AnyMessage)[member.localName];
         let copy: any;
         if (member.repeated) {
-          copy = (source as any[]).map((e) => cloneSingularField(member, e));
+          copy = (source as any[]).map((e) => cloneSingularField(e));
```
timostamm, line 216: TY for the cleanup!

```suggestion
          copy = (source as any[]).map(cloneSingularField);
```

packages/protobuf/src/private/util-common.ts
timostamm: Apologies, I meant Object.**from**Entries 🙈  Object.entries is fine, it's part of ES2017.

packages/protobuf/src/private/util-common.ts
```diff
@@ -55,18 +54,37 @@ export function makeUtilCommon(): Omit<Util, "newFieldList" | "initFields"> {
...
               case "enum":
-                Object.assign(t[localName], s[localName]);
+                let target = s[localName];
+                if (member.V.T === ScalarType.BYTES) {
+                  target = {};
+                  for (const [k, v] of Object.entries(s[localName])) {
+                    target[k] = toU8Arr(v as ArrayLike<number>);
+                  }
+                }
+                Object.assign(t[localName], target);
```
timostamm, line 87: We should avoid modifying the input. What about

```suggestion
                if (member.V.T === ScalarType.BYTES) {
                  for (const [k, v] of Object.entries(s[localName])) {
                    t[localName][k] = toU8Arr(v as ArrayLike<number>);
                  }
                } else {
                  Object.assign(t[localName], s[localName]);
                }
```

?

srikrsna: We are not modifying the response, we reset target at line 82 to `{}`

timostamm: I missed that one too 🙈. Discussed offline, we went with the one with smaller bundle size.

PR #535, opened 2023-07-26
Update debugJsonValue to accept unknown

GauBen:
Hey! We started migrating a large code base to protobuf, we'll probably be uncovering a few bugs on the  way.

This merge request addresses an issue regarding `debugJsonValue`:

```
TypeError: Cannot read properties of undefined (reading 'toString')
     at Object.debugJsonValue [as debug]
     ...
```

### DIFF
--- a/packages/protobuf-bench/README.md
+++ b/packages/protobuf-bench/README.md
@@ -10,5 +10,5 @@ server would usually do.
 
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
-| protobuf-es         | 88,672 b      | 37,773 b | 9,865 b |
+| protobuf-es         | 88,669 b      | 37,770 b | 9,843 b |
 | protobuf-javascript | 394,384 b  | 288,761 b | 45,123 b |
--- a/packages/protobuf/src/private/json-format-common.ts
+++ b/packages/protobuf/src/private/json-format-common.ts
@@ -328,7 +328,7 @@ export function makeJsonFormatCommon(
   };
 }
 
-function debugJsonValue(json: JsonValue): string {
+function debugJsonValue(json: unknown): string {
   if (json === null) {
     return "null";
   }
@@ -337,10 +337,8 @@ function debugJsonValue(json: JsonValue): string {
       return Array.isArray(json) ? "array" : "object";
     case "string":
       return json.length > 100 ? "string" : `"${json.split('"').join('\\"')}"`;
-    case "undefined":
-      return "undefined";
     default:
-      return json.toString();
+      return String(json);
   }
 }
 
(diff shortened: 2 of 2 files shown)


timostamm:
If this happened during parsing of a valid JSON value, it is a bug. 

Note that `fromJson` _only_ parses from a JSON value (anything returned by JSON.parse). If you coerce the type system and pass `undefined` or an object containing `undefined`, you run into undefined behavior. For example:

```ts
import { Struct } from "@bufbuild/protobuf";
Struct.fromJson(undefined); 
```

The compiler should error with `TS2345: Argument of type 'undefined' is not assignable to parameter of type 'JsonValue'` on line 2. If you run this without compiling, you will see the error you mentioned. 

Can you provide details on how this error occurred?

GauBen:
It was indeed not a valid JSON, but called with undefined as first parameter: `Message.fromJson(somethingThatCouldReturnUndefined())`.

timostamm:
> It was indeed not a valid JSON, but called with undefined as first parameter: `Message.fromJson(somethingThatCouldReturnUndefined())`.

That should be caught by the TypeScript compiler:

```ts
import { Duration, JsonValue } from "@bufbuild/protobuf";
function somethingThatCouldReturnUndefined(): JsonValue | undefined {
  return undefined;
}
Duration.fromJson(somethingThatCouldReturnUndefined());
// TS2345: Argument of type 'JsonValue | undefined' is not assignable to
// parameter of type 'JsonValue'. Type 'undefined' is not assignable to type
// 'JsonValue'.
```


We could change the signature:

```diff
- fromJson(jsonValue: JsonValue, options?: Partial<JsonReadOptions>)
+ fromJson(jsonValue: JsonValue | undefined, options?: Partial<JsonReadOptions>)
```

But it would be _very_ unusual for a function to accept `undefined` as an argument only to reject it by raising an error. Depending on context, maybe it could help to JSON.stringify() the return value of somethingThatCouldReturnUndefined() and use Message.fromJsonString()? This effectively passes the value through a JSON roundtrip, guaranteeing that the value is valid JSON.

GauBen:
> That should be caught by the TypeScript compiler

It wasn't because one of the branch was returning `any` (like `JSON.parse` does), making `JsonValue` noneffective

I'd advocate for `fromJson` to keep its actual signature and making `debugJsonValue` more resilient by accepting `unknown` as argument instead of `JsonValue` and performing additional runtime checks

timostamm:
> making debugJsonValue more resilient by accepting unknown as argument

👍 that is reasonable. 

Changing the default case to return `String(json)` instead of `json.toString` should be a simple solution for everything that falls through. Would be happy to accept this improvement.

GauBen:
Updated with this improvement!

timostamm (approved):
Thank you 🙂 

Pushed up a commit with updated bundle size benchmark (`make bench`).

GauBen:
It made the package 3 bytes smaller, which is an absolute win

PR #536, opened 2023-07-28
Integrate `arethetypeswrong`

### LINKED ISSUE #456: Verify our exports with arethetypeswrong
@fubhy pointed out https://arethetypeswrong.github.io/, written by one of the TypeScript maintainers.

A [project](https://github.com/arethetypeswrong/arethetypeswrong.github.io) that:
> attempts to analyze npm package contents for issues with their TypeScript types, particularly ESM-related module resolution issues.

We should take a look whether we can use this tool to verify that our build artifacts are exported correctly for the cases we do not cover ourselves in the tests. This will also be interesting for our other repositories (including @bufbuild/connect, which uses subpath exports).

smaye81:
Fixes #456.

In light of the CLI of `arethetypeswrong` being [published](https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/15#issuecomment-1575691921), this adds integration with it to the Protobuf-ES repo.

The integration showed errors with [@bufbuild/protobuf](https://arethetypeswrong.github.io/?p=%40bufbuild%2Fprotobuf%401.3.0) and [@bufbuild/protoplugin](https://arethetypeswrong.github.io/?p=%40bufbuild%2Fprotoplugin%401.3.0). The same change fixed errors in both packages.

Note the copying of the generated declaration file. In TypeScript [docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing), it seems to imply that the contents of these can be the same.

> It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them. Every declaration file is interpreted either as a CommonJS module or as an ES module, based on its file extension and the "type" field of the package.json, and this detected module kind must match the module kind that Node will detect for the corresponding JavaScript file for type checking to be correct. Attempting to use a single .d.ts file to type both an ES module entrypoint and a CommonJS entrypoint will cause TypeScript to think only one of those entrypoints exists, causing compiler errors for users of the package.

### DIFF
--- a/packages/protoplugin/package.json
+++ b/packages/protoplugin/package.json
@@ -13,17 +13,30 @@
     "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
     "build": "npm run build:cjs && npm run build:esm+types",
     "build:cjs": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types"
+    "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types && cp ./dist/types/index.d.ts ./dist/types/index.d.cts && cp ./dist/types/ecmascript/index.d.ts ./dist/types/ecmascript/index.d.cts",
+    "lint:types": "attw --pack"
   },
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/esm/index.js",
-      "require": "./dist/cjs/index.js"
+      "import": {
+        "types": "./dist/types/index.d.ts",
+        "default": "./dist/esm/index.js"
+      },
+      "require": {
+        "types": "./dist/types/index.d.cts",
+        "default": "./dist/cjs/index.js"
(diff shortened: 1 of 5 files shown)


packages/protoplugin/package.json
```diff
@@ -13,17 +13,30 @@
     "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
     "build": "npm run build:cjs && npm run build:esm+types",
     "build:cjs": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types"
+    "build:esm+types": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm --declaration --declarationDir ./dist/types && cp ./dist/types/index.d.ts ./dist/types/index.d.cts && cp ./dist/types/ecmascript/index.d.ts ./dist/types/ecmascript/index.d.cts",
```
timostamm, line 16: That's not very maintainable :sweat_smile: 

When I looked into this previously, I noticed that it is actually pretty cheap to emit types in addition to js. So I think it should be possible to simply emit types when building cjs _and_ esm, and have a .d.ts file next to each .js file.

smaye81: Updated `packages/protobuf` accordingly (I think). Is that what you mean? If so, I can do the same to `protoplugin`.

PR #563, opened 2023-09-25
Update protoplugin example

timostamm:
- Update buf.gen.yaml to use the new path syntax added in buf v1.14.0 to make the example run on windows.
- Add tests for the generated code.
- Update the documentation for writing plugins with guidance for testing.
- Update the example package to be self-sufficient and update the README.


emcfarlane (approved):
lgtm!

Issue #509, opened 2023-06-19
Dual package hazard


fubhy:
I just ran into the dual package hazard problem with `@bufbuild/protobuf` and the exported WKT types. I was checking `instanceof` on a Timestap type and that failed because the object instance I got was from the CJS export (through a CJS library) and I was comparing it against `Timestamp` via my local ESM import.

Personally, I think it's time to drop CJS altogether. Top-level await has been available for quite some time now and is a perfectly acceptable escape hatch for anyone who needs to import ESM modules into CJS via dynamic import.

smaye81:
Hey @fubhy unfortunately I think this might be an issue with the ecosystem. But, can you share a bit more details with what happened? What CJS library was it that pulled in the CJS export, etc.?

Dropping CJS would be nice, but not sure we can without some tests to ensure we're not breaking a bunch of bundlers out there. I can maybe look into this with our connect-es-integration repo to see how things would behave.

fubhy:
Hey @smaye81. Sorry for the late reply!

I can see how it might be difficult for y'all to entirely drop CJS support now that it's already "live" with both ESM & CJS.

Maybe, instead of dropping it entirely you'd consider publishing the ESM package as a CJS wrapper instead? That would eliminate the dual package hazard effectively too... Admittedly at a bit of a maintenance burden in terms additional build & build config shenanigans.

Particularly for `@bufbuild/protobuf` though, I think this issue can't be ignored... I ran into it in an internal monorepo with some shared dependencies between some backend & frontend services. Due to some configuration issues involving some legacy packages, the shared package with the protobuf definitions was loaded twice. Once from CJS and once from ESM.

Given how "low level" particularly the "@bufbuild/protobuf" is for anything that has anything to do with connect, I'd be willing to bet that this is destined to happen more frequently as adoption increases. Especially as people might start to publish & reuse protos & utils via shared packages. A single misconfigured (or intentionally CJS only) package and you run into this. Mind you: this happened inside a single (admittedly large) monorepo with only local lib & app packages for me and it was already hard to debug ;-)

fubhy:
... This part in particular:

```
[...]

PR #566, opened 2023-09-29
Add ESM wrapper to avoid dual package hazard

timostamm:
Fixes https://github.com/bufbuild/protobuf-es/issues/509 with the solution proposed in https://github.com/bufbuild/protobuf-es/issues/509#issuecomment-1634670621.

We've tested the approach in https://github.com/connectrpc/examples-es/pull/1002 in various versions of Node.js, esbuild, parcel, rollup, vite, and webpack. Tree-shaking continues to work as expected with the "module" condition, and the dual package hazard is resolved with the esm wrappers in the "import" condition.


smaye81 (approved):

Andarist:
Excellent work on this! :)

PR #584, opened 2023-10-10
Update to google protobuf v24.4

timostamm:
google/protobuf/descriptor.proto and google/protobuf/compiler/plugin.proto have changed for editions.

Some unittest proto files have been updated with editions. Since our plugin does not support them yet, we skip generating them. These proto files are only used to test code generation. We will include them again once we have sufficient support for editions.


smaye81 (approved):

smaye81:
nit: PR name should be 24.4

PR #588, opened 2023-10-13
Make `toPlainMessage` idempotent

haines:
This PR makes it possible to call `toPlainMessage` on a `PlainMessage<T>` object as well as a `Message<T>` instance.

The motivation for this change is the lack of exact types in TypeScript. Specifically, if I define a function that accepts a `PlainMessage<T>` as an argument, it's valid to pass a `Message<T>` instance as well, because it has all the properties necessary to satisfy the `PlainMessage<T>` interface.

If I need to ensure that I actually have a `PlainMessage<T>` inside the function (e.g. for the reasons described in https://github.com/connectrpc/connect-es/issues/505), at the moment I have to type the parameter as `T | PlainMessage<T>`, and check with `instanceof Message` before calling `toPlainMessage`.

### DIFF
--- a/packages/protobuf-test/src/to-plain-message.test.ts
+++ b/packages/protobuf-test/src/to-plain-message.test.ts
@@ -261,6 +261,14 @@ describe("toPlainMessage", () => {
     const act = toPlainMessage(new WrappersMessage(exp));
     expect(act).toEqual(exp);
   });
+  describe("on plain messages", () => {
+    const exp: PlainMessage<MessageFieldMessage> = {
+      messageField: undefined,
+      repeatedMessageField: [],
+    };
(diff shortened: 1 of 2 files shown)


timostamm (commented):
Makes sense to me to support this for convenience 👍 Thank you for adding the test.

Could you add the following to the JSDoc?

```

If the argument is already a plain message, it is
returned as is.
```

timostamm (approved):
LGTM!

haines:
Thanks!

timostamm:
This was just released in [v1.4.0](https://github.com/bufbuild/protobuf-es/releases/tag/v1.4.0).

PR #611, opened 2023-11-07
build(protobuf, protoplugin): reorder exports field for native esm compatibility

### LINKED ISSUE #610: Cannot find module '/app/server/node_modules/@bufbuild/protobuf/dist/proxy/index.js' imported from /app/server/chunks/app/server.mjs
Hi. I am using Nuxt 3 for my application where I use this package. The problem is that after version 1.3.3. I started getting this error in production. From what I found, the proxy directory is missing from the build. I think this change is the cause: https://github.com/bufbuild/protobuf-es/pull/566

Do you have an idea how to resolve this?

pi0:
Resolves #610

By using `import` export condition first, we ask bundlers such as rollup to prefer an export that is compatible with native ESM (ie: directly importing in Node.js when externalized in build). 

See https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1799898255 for more context.

### DIFF
--- a/packages/protobuf/package.json
+++ b/packages/protobuf/package.json
@@ -25,9 +25,9 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
-      "module": "./dist/esm/index.js",
+      "import": "./dist/proxy/index.js",
       "require": "./dist/cjs/index.js",
-      "import": "./dist/proxy/index.js"
+      "module": "./dist/esm/index.js"
     }
   },
   "devDependencies": {
--- a/packages/protoplugin/package.json
+++ b/packages/protoplugin/package.json
@@ -21,14 +21,14 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
-      "module": "./dist/esm/index.js",
+      "import": "./dist/proxy/index.js",
       "require": "./dist/cjs/index.js",
-      "import": "./dist/proxy/index.js"
+      "module": "./dist/esm/index.js"
     },
     "./ecmascript": {
-      "module": "./dist/esm/ecmascript/index.js",
+      "import": "./dist/proxy/ecmascript/index.js",
       "require": "./dist/cjs/ecmascript/index.js",
-      "import": "./dist/proxy/ecmascript/index.js"
+      "module": "./dist/esm/ecmascript/index.js"
     }
   },
   "typesVersions": {


packages/protobuf/package.json
```diff
@@ -25,9 +25,9 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
-      "module": "./dist/esm/index.js",
+      "import": "./dist/proxy/index.js",
       "require": "./dist/cjs/index.js",
-      "import": "./dist/proxy/index.js"
+      "module": "./dist/esm/index.js"
```
timostamm, line 30: This will cause webpack, rollup, wmr ([ref](https://webpack.js.org/guides/package-exports/#conditions)), esbuild ([ref](https://esbuild.github.io/api/#how-conditions-work)) to choose "import" over "module".

I don't think we can do that, because dist/proxy/index.js is an ESM wrapper that re-exports CJS to avoid the dual package hazard. I'm pretty sure it will cause problems for tree shaking, and possibly other issues.

For reference, here is how the current exports were tested: https://github.com/connectrpc/examples-es/pull/1002

I think some other solution is necessary to fix the issue, while maintaining behavior for other consumers. I'm not entirely sure which one yet 🙂

pi0: I'm pretty sure we can find some better solutions to support dual package hazard and replace proxy with an alternative, yet make a fully standard ESM package with a standard `import` export condition that works natively. (if this is your main concern to resolve) -- Feel free to direct me in discord (pi0), i can share my 50 cents of experience with dealing with ESM/CJS dual exports.

pi0:
Updated PR as per https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1802152966 suggestion to use `node` export condition as a quick fix.

Note: I have not e2e tested it against nitro. In theory it should work as we apply `node` export condition when tracing. @DavidDomkar if not too much trouble, would be nice if you can check by patching in `node_modules`.

DavidDomkar:
@pi0 thanks for modifying the PR with a quick fix. I can confirm that when I patch the package in node_modules with changes from this PR, the problem disappears. 🎉🎉

smaye81:
Hi, @pi0 appreciate your patience on this. We've been testing this extensively in [Connect-ES](https://github.com/connectrpc/connect-es) also because we've noticed some similar issues with exports there. After a lot of investigation, we landed on the format specifed in this PR: https://github.com/connectrpc/connect-es/pull/921.

Would you be able to modify this PR so that the exports are in the same order and configured the same way?

smaye81:
Hi @pi0. I actually pushed up the changes to your branch, I hope you don't mind. We'd like to get this fixed in all our npm packages as soon as possible. If things seem OK to you, we can merge this branch in.

cc @timostamm

nicksnyder:
Hi @pi0 , can you sign the CLA?

pi0:
Thanks for following and considering the change ❤️ https://github.com/bufbuild/protobuf-es/pull/645 looks good!

smaye81:
Also, FYI, this was released in [v1.5.1](https://www.npmjs.com/package/@bufbuild/protobuf) today.

PR #612, opened 2023-11-15
Add tooling script to cache key

timostamm:
In packages/upstream-protobuf, we're fetching `protoc` and other dependencies, and cache them based on the version. Since the logic of the script can affect the outcome, this is brittle. This PR adds the script itself to the cache key.

### DIFF
--- a/packages/upstream-protobuf/bin/upstream-files.mjs
+++ b/packages/upstream-protobuf/bin/upstream-files.mjs
@@ -43,7 +43,6 @@ async function main(args) {
             exitUsage();
     }
     stdout.write(protoInclude.files.join(" "));
-    exit(0);
 }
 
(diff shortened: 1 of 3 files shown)


packages/upstream-protobuf/bin/upstream-files.mjs
```diff
@@ -43,7 +43,6 @@ async function main(args) {
             exitUsage();
     }
     stdout.write(protoInclude.files.join(" "));
-    exit(0);
```
timostamm, line 46: Unrelated. It's possible that Node.js exits before it has flushed what we wrote to stdout in the line above. Removing this line let's Node.js exit on its own.

gilwong00 (approved):

PR #615, opened 2023-11-16
Make edition feature defaults from `protoc` available

timostamm:
The recent [release v25](https://github.com/protocolbuffers/protobuf/releases/tag/v25.0) of `protoc` makes [edition feature resolution](https://github.com/protocolbuffers/protobuf/blob/main/docs/design/editions/protobuf-editions-design-features.md#specification-of-an-edition) available via a flag:
`--experimental_edition_defaults_out` outputs a [google.protobuf.FeatureSetDefaults](https://github.com/protocolbuffers/protobuf/blob/1a5cdb937826dbabc2c05872c5d7f1d918dea861/src/google/protobuf/descriptor.proto#L1026) message, which can be used to seed feature resolution.

This PR adds functionality to the internal upstream-protobuf package to make it easy to shell out to `protoc`, and to generate the defaults.

### DIFF
--- a/packages/protobuf-test/src/json-names.test.ts
+++ b/packages/protobuf-test/src/json-names.test.ts
@@ -13,115 +13,58 @@
 // limitations under the License.
 
 import { expect, test } from "@jest/globals";
-import { mkdtempSync, readFileSync, writeFileSync } from "fs";
-import { join } from "path";
-import { tmpdir } from "os";
-import { spawnSync } from "child_process";
 import { FileDescriptorSet, proto3, ScalarType } from "@bufbuild/protobuf";
+import { UpstreamProtobuf } from "upstream-protobuf";
 
-test("JSON names equal protoc", () => {
-  expect(getProtocJsonName("foo_bar")).toBe("fooBar");
-  expectRuntimeJsonNameEqualsProtocJsonName("foo_bar");
-  expectRuntimeJsonNameEqualsProtocJsonName("__proto__");
-  expectRuntimeJsonNameEqualsProtocJsonName("fieldname1");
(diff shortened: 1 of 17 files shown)


packages/protobuf-test/src/json-names.test.ts
```diff
@@ -13,115 +13,58 @@
 // limitations under the License.
 
 import { expect, test } from "@jest/globals";
-import { mkdtempSync, readFileSync, writeFileSync } from "fs";
-import { join } from "path";
-import { tmpdir } from "os";
-import { spawnSync } from "child_process";
 import { FileDescriptorSet, proto3, ScalarType } from "@bufbuild/protobuf";
+import { UpstreamProtobuf } from "upstream-protobuf";
```
timostamm, line 17: Since we're here, I've also added the `compileToDescriptorSet` method, which gives us the generic option to generate descriptors from inline protobuf source, and makes this test a bit more readable.

jhump:
I'm a little confused by the commit history. IIUC, it looks like the meat of the changes are in e16130a36d4e00b8c6e24997ff2085760482ce44, but that was apparently merged to main already in #616. (Oops?)

The diff for this PR shows zero files changed, so I guess that's why.

Anyhow, I looked at that commit, and it looked sane to me 😄!

timostamm:
> The diff for this PR shows zero files changed, so I guess that's why.

Welp. I split the noisy formatter fixes into separate PRs for a better commit history :neutral_face:

Yes, this was supposed to be just https://github.com/bufbuild/protobuf-es/commit/e16130a36d4e00b8c6e24997ff2085760482ce44.

smaye81 (approved):

PR #621, opened 2023-11-17
Expose syntax "editions" in descriptor sets

timostamm:
This change exposes the editions syntax in a `DescriptorSet`, the wrapper around protobuf file descriptor messages provided by `createDescriptorSet` from `@bufbuild/protobuf`.

The type `DescFile` receives a new property:

```ts
  /**
   * The edition of the protobuf file. Will be EDITION_PROTO2 for syntax="proto2",
   * EDITION_PROTO3 for syntax="proto3";
   */
  readonly edition: Omit<Edition, ...>;
```

The `syntax` property is updated to include `"editions"` in the union type:

```ts
  /**
   * The syntax specified in the protobuf source.
   */
  readonly syntax: "proto3" | "proto2" | "editions";
```

For example, a file with `syntax="proto3"` will have the properties `syntax: "proto3"` and `edition: Edition.EDITION_PROTO3`. A file with `edition="2023"` will have the properties `syntax: "editions"` and `edition: Edition.EDITION_2023`.

FeatureSets are not exposed yet. `protoc-gen-es` cannot generate code for editions yet. This change just allows `createDescriptorSet` to wrap descriptors using editions.

### DIFF
--- a/packages/protobuf/src/create-descriptor-set.ts
+++ b/packages/protobuf/src/create-descriptor-set.ts
@@ -19,6 +19,7 @@ import type {
   OneofDescriptorProto,
 } from "./google/protobuf/descriptor_pb.js";
 import {
+  Edition,
   FieldDescriptorProto_Label,
   FieldDescriptorProto_Type,
   FieldOptions_JSType,
@@ -45,7 +46,10 @@ import type {
 import { LongType, ScalarType } from "./field.js";
 import { MethodIdempotency, MethodKind } from "./service-type.js";
 import { fieldJsonName, findEnumSharedPrefix } from "./private/names.js";
-import { protoInt64 } from "./proto-int64.js";
+import {
+  parseTextFormatEnumValue,
+  parseTextFormatScalarValue,
+} from "./private/text-format.js";
 
 /**
  * Create a DescriptorSet, a convenient interface for working with a set of
@@ -92,17 +96,11 @@ interface Cart {
  */
 function newFile(proto: FileDescriptorProto, cart: Cart): DescFile {
   assert(proto.name, `invalid FileDescriptorProto: missing name`);
-  assert(
-    proto.syntax === undefined || proto.syntax === "proto3",
-    `invalid FileDescriptorProto: unsupported syntax: ${
-      proto.syntax ?? "undefined"
-    }`,
-  );
   const file: DescFile = {
     kind: "file",
     proto,
     deprecated: proto.options?.deprecated ?? false,
-    syntax: proto.syntax === "proto3" ? "proto3" : "proto2",
+    ...parseFileSyntax(proto.syntax, proto.edition),
     name: proto.name.replace(/\.proto/, ""),
     enums: [],
     messages: [],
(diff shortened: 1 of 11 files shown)


packages/protoc-gen-es/src/javascript.ts
```diff
@@ -43,7 +44,7 @@ export function generateJs(schema: Schema) {
 
 // prettier-ignore
 function generateEnum(schema: Schema, f: GeneratedFile, enumeration: DescEnum) {
-  const protoN = schema.runtime[enumeration.file.syntax];
+  const protoN = getNonEditionRuntime(schema, enumeration.file);
```
timostamm, line 47: The addition of `"editions"` to the `syntax` property will break code that only expects `"proto2" | "proto3"`, such as this expression. :disappointed:

packages/protobuf/src/create-descriptor-set.ts
```diff
@@ -793,36 +854,41 @@ function isOptionalField(
...
+          // are not packed by default, which makes sense because they are length-delimited.
+          return true;
+        default:
+          return false;
+      }
+    case "proto2":
+      return false;
+    case "editions":
+      // TODO support edition featureset
+      return true;
```
jhump, line 890: nit: I would probably unify this with "proto3" so it is at least closer to accurate. By default in editions (without element-level feature override), it behaves the same as proto3.

packages/protoc-gen-es/src/javascript.ts
jhump: By "will break", does that mean this function would throw an exception in the face of "editions" syntax? If not, what exactly will happen here?

jhump (approved):

packages/protoc-gen-es/src/javascript.ts
timostamm: In this specific code, nothing will happen. The `getNonEditionRuntime` will throw an error at runtime, but it never get's a chance to do so, because protoc-gen-es does not opt into edition support yet.

But the change I'm making here was necessary to avoid a compile-time error caused by `file.syntax` no longer being compatible as an index to `schema.runtime`.

Updated the PR description.

packages/protobuf/src/create-descriptor-set.ts
timostamm: In EDITION_2023 it does, but in EDITION_PROTO2 it does not 😛 Will update this logic in the next step, where this value will be plucked from FeatureSetDefaults.

jhump: But if the edition is `EDITION_PROTO2`, the syntax will _not_ be "editions"; it will be "proto2".

timostamm: You're right about "proto2" of course. 

Here is the [new implementation of isPackedFieldByDefault](https://github.com/bufbuild/protobuf-es/pull/627/files#diff-b8e9e2ab828f5f89cf78736298aafb3fb0727af2b53730fa9ef3ba3563025348R959). It uses the seeded default value of `repeated_field_encoding` for the edition, including `EDITION_PROTO2`, `EDITION_PROTO3`, `EDITION_2023`, and any future edition.

The implementation is shorter and seems less ambiguous. I guessed it would, which is why I wanted to hold off modifying it in this PR. I should have explained better.

PR #624, opened 2023-11-20
Add test-protoplugin-example to CI

smaye81:
This adds the `make test-protoplugin-example` target to CI and in doing so, fixes a few minor issues:

* Regenerates the code in that directory so that it passes `checkdiff`.
* In addition to adding to CI, this also adds a Node version matrix to the CI step so that the CI operations are run on both Node 18 and 20.
* Fixes an issue with running `tsx` in Node 20. Previously we were using `node --loader tsx ...` in the test step, but in Node 20.6 and above, the `--loader` flag was deprecated in favor of the `--import` flag. This then causes errors running the tests in that package (see this [issue](https://github.com/privatenumber/tsx/issues/349) for more context). However, `tsx` is simply a wrapper around Node, so we can just issue the command directly via `tsx --test ...` instead of using a loader/import flag.


packages/protoplugin-example/test/generated.ts
```diff
@@ -14,8 +14,8 @@
 
 import * as assert from "node:assert/strict";
 import { describe, it, mock } from "node:test";
-import { ElizaServiceClient } from "../src/gen/connectrpc/eliza/v1/eliza_twirp";
-import { SayRequest } from "../src/gen/connectrpc/eliza/v1/eliza_pb";
+import { ElizaServiceClient } from "../src/gen/connectrpc/eliza/v1/eliza_twirp.js";
```
smaye81, line 17: Driveby change I noticed that should probably be fixed.

timostamm (approved):

PR #640, opened 2023-12-04
Support proto2 groups

timostamm:
This PR implements support for the proto2 groups, which allows to declare a field and a message at the same time. 

Here is an example - message `M` has a group field `g`:

```protobuf
syntax="proto2";
message M {
  optional group G = 1 {
    string field = 1;
  }
}
```

In descriptors, groups are represented as messages, and like many other implementations, Protobuf-ES generates groups as messages, Unfortunately, it does not implement the group wire format, and fails to serialize / parse messages correctly from the binary format.

This PR fixes the issue. It adds the property `delimited` to the `FieldInfo` type, and sets it to `true` in generated code for group fields (it is `false` or `undefined` otherwise). Based on this information, group fields are serialized and parsed with proper the tag-delimited format.


packages/protobuf/src/private/binary-format-common.ts
```diff
@@ -91,14 +90,22 @@ export function makeBinaryFormatCommon(): Omit<BinaryFormat, "writeMessage"> {
...
+      lengthOrEndTagFieldNo: number,
       options: BinaryReadOptions,
+      delimitedMessageEncoding?: boolean,
     ): void {
       const type = message.getType();
-      const end = length === undefined ? reader.len : reader.pos + length;
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      const end = delimitedMessageEncoding
+        ? reader.len
+        : reader.pos + lengthOrEndTagFieldNo;
```
jhump, line 101: This is a semantic change right? Previously, an `undefined` length could be provided, but this new expression doesn't look like it will handle that the same way/correctly. Should the first part of the ternery instead be `delimitedMessageEncoding || length === undefined`?

packages/protobuf/src/private/binary-format-common.ts
```diff
@@ -345,15 +367,21 @@ export function writeMapEntry(
 export function writeMessageField(
   writer: IBinaryWriter,
   options: BinaryWriteOptions,
-  type: MessageType,
-  fieldNo: number,
+  field: FieldInfo & { kind: "message" },
   value: any,
```
jhump, line 371: Is this signature change not a compatibility issue? I guess I would have expected this to remain, for backwards-compatibility, and an alternate one introduced for group handling (with the shared logic factored into some shared function).

jhump (approved):
Couple of questions, but looks great to me!

packages/protobuf/src/private/binary-format-common.ts
timostamm: Luckily, the function is not part of the public API. It is exported from this file, but it's not exported from the package.

packages/protobuf/src/private/binary-format-common.ts
timostamm: This threw me off as well. The code used to handle length being `undefined`, but the signature does not allow `undefined` to be passed, and I couldn't find any indication it ever did. I think this must have been left over from initial development.

PR #644, opened 2023-12-05
Honor editions in the preamble generated by @bufbuild/protoplugin

timostamm:
A <sub><sup>tiny</sup></sub> feature: In generated code, we add the following information in a preamble:

```ts
// @generated from file x.proto (syntax proto3)
```

With this update, editions are supported as well. For example:

```ts
// @generated from file x.proto (edition 2023)
```

(Based on https://github.com/bufbuild/protobuf-es/pull/643)


packages/protoplugin-test/src/transpile.test.ts
```diff
@@ -107,120 +102,6 @@ describe("transpile", function () {
     ]);
   });
 
-  describe("file preamble", () => {
```
timostamm, line 110: Not sure why those tests where here. Moved to a different file and increased coverage.

smaye81 (approved):

Issue #587, opened 2023-10-12
Add CommonJS support as a plugin option


mfridman:
The current code generators (like `bufbuild/protobuf-es` and `connectrpc/es`) support ESM only. What do you think about adding an opt-in plugin option to modify the default behaviour?

As this project becomes more popular, it'd be nice for the generated code to "play nice" with the rest of the ecosystem. For example, some projects like NestJS do not work well with ESM-only. And the alternatives aren't great, such as importing the module asynchronously. 

Given that the Buf Schema Registry (BSR) hosts plugins with a pre-configured set of options, it means we'd need to publish a second plugin named based on the import. This isn't great, so something to think about.

https://buf.build/bufbuild/es vs https://buf.build/bufbuild/es-commonjs (does not exist)

Lastly, since we'll want this to be a "registry-compatible" plugin, we'll need a commonjs implementation for rewriting import paths.

therockstorm:
From [this Slack thread](https://bufbuild.slack.com/archives/CRZ680FUH/p1696444329276079), without BSR support for CommonJS (and, by extension, NestJS), the "best" way to handle this is to bypass BSR completely and to instead create a monorepo with CI builds to generate CommonJS NPM packages.

Is this true? If so, how likely is Buf to support CommonJS and would we be talking months or years? If this is farther away than a couple of months, we'll have to move ahead with doing all of this ourselves and removing BSR from our workflow. NestJS has [no plans](https://github.com/nestjs/nest/issues/7021#issuecomment-831799620) to update to ESM. Another example: [this almost two year old PR](https://github.com/nestjs/nest/pull/8736).

timostamm:
Hey Rocky, I am not sure that bypassing the BSR completely is the best workaround. NestJS does not support ESM-only npm packages, but it _does_ support TypeScript files with ESM imports.

For example, you might have an import like the following in your app:

```ts
// app.module.ts
import { AppController } from './app.controller';
```

This works in Nest, because all source files are transpiled to CJS.

If you generate code locally with protoc-gen-es, you can set the option `target=ts` to get the same result - TypeScript files with ESM imports. If you import them in your app, they will be transpiled to CJS like any other source file.

It isn't as neat as remote packages, but you can generate local code from remote protobuf files on the BSR. 

You need one configuration file:

```yaml
# buf.gen.yaml
version: v1
plugins:
  - plugin: buf.build/bufbuild/es
    out: src/gen
    opt: target=ts
[...]

PR #648, opened 2023-12-07
Add CommonJS support as a plugin option

timostamm:
This PR adds support for generating JavaScript for the [CommonJS](https://nodejs.org/api/modules.html#modules-commonjs-modules) module system.

When the plugin option `js_import_style` is set to `legacy_commonjs`, generated JavaScript will use `require()` to import, and the `exports` variable to export.

Here is a shortened example, generated from [example.proto](https://github.com/bufbuild/protobuf-es/blob/e7003368876017b1703cea141b541e70509354bf/packages/protobuf-test/extra/example.proto):

```js
// @generated by protoc-gen-es v1.5.1 with parameter "target=js+dts,js_import_style=legacy_commonjs"
// @generated from file extra/example.proto (package docs, syntax proto3)

"use strict";
Object.defineProperty(exports, "__esModule", { value: true });

const { proto3 } = require("@bufbuild/protobuf");

/**
 * @generated from message docs.User
 */
const User = ...

exports.User = User;
```

As you can see, [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) is enabled. For best interoperability, we define [__esModule](https://2ality.com/2017/01/babel-esm-spec-mode.html).

**For plugin authors:**
- To support CommonJS output, your code must use `exportDecl` instead of printing an `export` statement. For example, use `f.print(f.exportDecl("abstract class", "MyClass"), " {}")` instead of `f.print("export abstract class MyClass {}")`.
- In case you are [overriding the built-in transpilation](https://github.com/bufbuild/protobuf-es/blob/main/docs/writing_plugins.md#overriding-transpilation), your  `transpile` function needs to handle the new 4th argument `jsImportStyle` to support CommonJS output.

Closes https://github.com/bufbuild/protobuf-es/issues/587.


docs/writing_plugins.md
```diff
@@ -322,48 +327,44 @@ Note that some of the `ImportSymbol` types in the schema runtime (such as `JsonV
...
+- **Import styles**: If the plugin option `js_import_style=legacy_commonjs` is set, code is automatically generated
+  with `require()` calls instead of `import` statements.
   
 
 ### Exporting
 
-Working with exports is accomplished via the `export` function on the generated file object.  Let's walk through an example:
-
-Suppose you generate a validation function for every message.  If you have a nested message, such as:
+To export a declaration from you code, use `exportDecl`:
```
smaye81, line 340: nit: `your` code

timostamm:
> Do you also want to update the docs here? https://github.com/bufbuild/protobuf-es/blob/main/docs/generated_code.md?plain=1#L81-L86

Thanks! Updated in https://github.com/bufbuild/protobuf-es/pull/648/commits/526fe50f26eecbd9f59cb2148eddce3cd6c8729c.

mfridman (approved):
🎉

Issue #86, opened 2022-05-10
Extensions


bufdev:
The docs say:

> We implement all proto2 features, except for extensions and the text format.

I don't think we can ignore extensions - what does the core js do?

SyrieBianco:
Replaced by TCN-61

fubhy:
Any news on this? This is currently my main blocker for implementing message validation (using https://github.com/bufbuild/protoc-gen-validate)

https://github.com/bufbuild/connect-web/issues/270

Is this being worked on or can I lend a hand here? If so, I'd love some pointers (tried to read through the code generator request => descriptor set creation code but am a bit lost to be honest :-)

timostamm:
Hey Sebastian, `findCustomMessageOption()` is indeed the way to go for now!

fubhy:
Yeah, I built my own version of that before I found that in your code :-D. Sorry for the noise.

zfedoran:
@fubhy are you able to share what you've done to get message validation running using `findCustomMessageOption()`?

rofrankel:
Just commenting to +1 the issue - I'd benefit from extension support (specifically, message and field options).

timostamm:
Support for extensions was just released in [v1.7.0](https://github.com/bufbuild/protobuf-es/releases/tag/v1.7.0).

Extensions can be used to retrieve custom Protobuf options from descriptors, [for example in a plugin](https://github.com/bufbuild/protobuf-es/blob/main/docs/writing_plugins.md#using-custom-protobuf-options). But it's not possible yet to retrieve options at runtime from generated code. We do have concrete plans for this feature though, and we'll keep the issue https://github.com/bufbuild/protobuf-es/issues/397 updated.

PR #654, opened 2023-12-13
Improve errors for unsupported editions in createDescriptorSet

timostamm:
While adding support for editions in `@bufbuild/protoplugin`, I noticed two issues with `createDescriptorSet()` and editions:

1. Unsupported editions only raise an error when calling `getFeatures()` on an element. This behavior is a bit surprising in practice, and I don't think we really gain much from deferring the error. Using `createDescriptorSet()` with editions from the future is still possible by passing a `google.protobuf.FeatureSetDefaults` message. So this PR changes the behavior to raise an error immediately on `createDescriptorSet()` when an unsupported edition is used.

2. We resolve features using `DescFile.edition`, but the property is dropping the test-only editions. This PR uses the original field `google.protobuf.FileDescriptorProto.edition` instead so that we can use the test-only editions in tests.

### DIFF
--- a/packages/protobuf-test/src/descriptor-set.test.ts
+++ b/packages/protobuf-test/src/descriptor-set.test.ts
@@ -12,19 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { describe, expect, test } from "@jest/globals";
+import { beforeAll, describe, expect, test } from "@jest/globals";
 import { readFileSync } from "fs";
 import type { AnyDesc } from "@bufbuild/protobuf";
 import {
+  BinaryReader,
   createDescriptorSet,
   Edition,
   FeatureSet,
   FeatureSet_EnumType,
+  FeatureSet_FieldPresence,
   FeatureSet_JsonFormat,
   FeatureSet_MessageEncoding,
   FeatureSet_RepeatedFieldEncoding,
   FeatureSet_Utf8Validation,
-  FeatureSet_FieldPresence,
+  FeatureSetDefaults,
   proto3,
   ScalarType,
 } from "@bufbuild/protobuf";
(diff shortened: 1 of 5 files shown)


packages/protobuf-test/src/descriptor-set.test.ts
```diff
@@ -209,6 +215,132 @@ describe("DescriptorSet", () => {
...
+      const fileDescriptorSet = FileDescriptorSet.fromBinary(
+        await new UpstreamProtobuf().compileToDescriptorSet(`edition="2023";`),
+      );
+      fileDescriptorSet.file[0].edition = Edition.EDITION_99999_TEST_ONLY;
+      expect(() =>
+        createDescriptorSet(fileDescriptorSet, {
+          featureSetDefaults,
+        }),
+      ).toThrow(
+        /^Edition EDITION_99999_TEST_ONLY is later than the maximum supported edition EDITION_2023$/,
```
timostamm, line 264: Using the test-only edition to test.

packages/protobuf-test/src/descriptor-set.test.ts
```diff
@@ -209,6 +215,132 @@ describe("DescriptorSet", () => {
...
+        .filter((f) => f.no === 9999);
+      assert(uf.length >= 0);
+      const tf = new TestFeatures();
+      for (const { data } of uf) {
+        const bytes = new BinaryReader(data).bytes();
+        tf.fromBinary(bytes, { readUnknownFields: false });
+      }
+      return tf as TestFeatures & Required<TestFeatures>;
+    }
+  });
```
timostamm, line 343: Just adding test coverage for language-specific extensions to FeatureSet. They work as expected. getTestFeatures can be replaced when we support extensions (https://github.com/bufbuild/protobuf-es/issues/86).

packages/protobuf/src/private/feature-set.ts
```diff
@@ -61,42 +61,35 @@ export function createFeatureResolver(
...
-          `No valid default found for edition ${Edition[edition]}`,
-        );
-      }
-      defaultsBin = highestMatch.f.toBinary();
-      defaultsBinByEdition.set(edition, defaultsBin);
+  if (edition < min) {
+    throw new Error(
+      `Edition ${Edition[edition]} is earlier than the minimum supported edition ${Edition[min]}`,
+    );
+  }
```
timostamm, line 68: Throwing here right away instead of in the resolver function, the previous behavior.

smaye81 (approved):

jhump (approved):
Sorry for late review. I thought I looked at this last week. I think I just forgot to submit a review. 🤦.

LGTM.

Issue #397, opened 2023-02-23
Provide custom options at runtime


hugebdu:
`@protobuf-ts/runtime-rpc/reflection-info#MethodInfo` has `options`.
Any plans to have it in `protobuf-es` as well?

Same applies to `ServiceInfo`

Thanks

timostamm:
enum, enum fields, messages, message fields, and files can also have options, but I'd agree that service and rpc are some of the most interesting ones. 

We've punted on this feature so far because some users don't want to ship their options in a bundle, so this can become complicated. Can I ask what you want to use it for?

hugebdu:
we use our custom options for various needs.
 - message fields annotated with validation options (both for API documentation and runtime validations)
 - methods and services annotated with some runtime-related options (permissions etc)

and then there are google http annotations wich we also use.

hugebdu:
Also backtracking reference (from `MethodInfo` to `ServiceInfo`) is missing and it is really convenient, especially in `protobuf-ts/runtime-rpc` interceptors.

I'm trying to migrate gradually to `protobuf-es` + `@grpc/grpc-js` from `protobut-ts` stack.

smaye81:
Thanks for the info @hugebdu.  Appreciate the input and this is definitely something we have on our minds.  As @timostamm mentioned though, it becomes complicated with some users who don't want their option values shipped as part of the bundle.  So, making both camps happy is tricky and probably involves some sort of plugin option.  

It's something we'd like to think through and get right, so I don't see us implementing this anytime _soon_ but it's on our radar.  In the meantime, though, you could implement your own plugin using Protobuf-ES' plugin framework.  Custom options are available to you during code generation time ([docs](https://github.com/bufbuild/protobuf-es/blob/main/docs/writing_plugins.md#reading-custom-options)). Maybe that solution would work for you?

hugebdu:
@smaye81 I have a workaround for now - I generate a [binary descriptor](https://googleapis.dev/python/protobuf/latest/google/protobuf/descriptor.html) as well and use it at runtime for all the option-bound logic.

Thanks

timostamm:
[Protocol Buffers v22.0](https://github.com/protocolbuffers/protobuf/releases/tag/v22.0) adds the feature [option retention](https://protobuf.dev/programming-guides/proto3/#option-retention). This could potentially help us here to control which options should be included in generated code.

hugebdu:
@smaye81  @timostamm 

[...]

PR #666, opened 2024-01-17
Support extensions

timostamm:
This PR adds support for extensions, a Protobuf language feature that allows to extend a message with an additional field. 

Here is an example:

```proto
syntax = "proto2";
import "example.proto";
package extra;

extend example.User {
  optional uint32 age = 100;
}
```

The message we extend must declare a range of field numbers for extensions:

```proto
// example.proto
syntax = "proto2";
package example;

message User {
  optional string name = 1;
  extensions 100 to 200;
}
```

Generating code with @bufbuild/protoc-gen-es will add a new export with the name of the extension field:

```ts
/**
 * @generated from extension: optional uint32 age = 100;
 */
export declare const age: Extension<User, number>;
```

Using the new accessor functions for extensions introduced by this PR, the field can be set on the extended message in a type-safe way:

```ts
import { clearExtension, getExtension, hasExtension, setExtension } from "@bufbuild/protobuf";
import { User} from "./gen/example_pb.js";
import { age } from "./gen/extra_pb.js";

const user = new User();
setExtension(user, age, 123);
hasExtension(user, age); // true
getExtension(user, age); // 123
clearExtension(user, age);
```

Extensions are supported in the JSON format as well. Similar to `google.protobuf.Any`, you 
have to provide a type registry to let the serializer know about extension.

<details>
<summary>Example for parsing and serializing messages with extensions.</summary>

```ts
import {createRegistry, getExtension, hasExtension} from "@bufbuild/protobuf";
import {User} from "./gen/example_pb.js";
import {age} from "./gen/extra_pb.js";

const options = {
  typeRegistry: createRegistry(age),
};

const user = User.fromJsonString(`
  {
   "name": "Joe",  
   "[extra.age]": 77 
  }
`, options);

hasExtension(user, age); // true
getExtension(user, age); // 77
user.toJsonString(options); // {"name":"Joe","[extra.age]":77}
```
</details>

**Related changes:**
- `createRegistry` accepts the new `Extension` type.
- The registries created by `createRegistry` and `createRegistryFromDesc` implement the new interface `IExtensionRegistry` to look up extensions.
- The `typeRegistry` JSON serialization option can now implement `IExtensionRegistry`.
- @bufbuild/protoplugin accepts `DescExtension` in `GeneratedFile.import`, `GeneratedFile.print`, and related methods to allow custom plugins to refer to extensions easily.

**Regression checks:**
- Bundle size in our benchmark increases by ~4.7%. 
- There are no noticeable regressions in performance from what I could see running `make perf`.

**A note about custom options:**
You can learn more about extensions in the [language guide](https://protobuf.dev/programming-guides/proto2/#extensions). Note that extensions can only be used for custom options in proto3. 

Currently, extensions cannot be used to retrieve custom options from generated code in Protobuf-ES, but we plan to follow up with a solution. https://github.com/bufbuild/protobuf-es/issues/397 is tracking the feature. Custom options can be retrieved from descriptors, so it is possible to use them to read options in a custom plugin.

**Planned follow-ups:**

- Add documentation in this repository in docs/
- Deprecate the custom option accessor functions exported from @bufbuild/protoplugin/ecmascript and document how to use extensions instead.

Closes https://github.com/bufbuild/protobuf-es/issues/86.


packages/protobuf-test/extra/extensions-proto2.proto
```diff
@@ -0,0 +1,79 @@
...
+  optional docs.User message_ext_proto3 = 6002;
+
+  repeated Proto2ExtMessage repeated_message_ext = 7001;
+  repeated Proto2ExtEnum repeated_enum_ext = 7005;
+  repeated string repeated_string_ext = 7002;
+  repeated uint32 packed_uint32_ext = 7003 [packed = true];
+  repeated uint32 unpacked_uint32_ext = 7004; // unpacked by default in proto2
+
+  optional google.protobuf.UInt32Value wrapper_ext = 8001;
+}
```
jhump, line 67: Since groups are now ostensibly supported, let's also add a group extension, to make sure they work, too. (Maybe a repeated group extension, too?)

packages/protobuf/src/create-registry.ts
```diff
@@ -71,9 +48,55 @@ export function createRegistry(
...
+    findExtensionFor(typeName: string, no: number): Extension | undefined {
+      return extensionsByExtendee.get(typeName)?.get(no) ?? undefined;
+    },
+    findExtension(typeName: string): Extension | undefined {
+      return extensionsByName.get(typeName) ?? undefined;
+    },
   };
+
+  function addType(type: MessageType | EnumType | ServiceType | Extension) {
+    if ("fields" in type) {
```
jhump, line 60: I take it this cannot use `instanceof` because the types in the union are all interfaces (vs. classes or other nominal JS types)? Unfortunate...

packages/protobuf/src/create-registry.ts
```diff
@@ -71,9 +48,55 @@ export function createRegistry(
...
+      }
+    } else {
+      enums[type.typeName] = type;
+    }
+  }
+
+  function addField(field: FieldInfo) {
+    if (field.kind == "message") {
+      addType(field.T);
+    } else if (field.kind == "map" && field.V.kind == "message") {
```
jhump, line 92: I guess the `kind` for group fields will be `"message"`? Or does this need to be expanded to include groups?

packages/protobuf/src/extension-accessor.ts
```diff
@@ -0,0 +1,125 @@
...
+ * an empty message instance for message fields).
+ */
+export function getExtension<E extends Message<E>, V>(
+  message: E,
+  extension: Extension<E, V>,
+  options?: Partial<BinaryReadOptions>,
+): V {
+  const opt = extension.runtime.bin.makeReadOptions(options);
+  assert(!extension.field.oneof); // oneof is not allowed for extensions
+  assert(extension.field.kind != "map"); // maps are not allowed for extensions
```
jhump, line 40: Is it possible to also assert here that the type name of the message matches the extension's extendee? I'm thinking about possible dynamic message cases, where a concrete type `E` of the message is not sufficient to guarantee that the extension correctly applies to it.

packages/protobuf/src/extension-accessor.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Retrieve an extension value from a message.
+ *
+ * The function never returns undefined. Use hasExtension() to check whether an
+ * extension is set. If the extension is not set, this function returns the
+ * default value (if one was specified in the protobuf source), or the zero value
+ * (for example `0` for numeric types, `[]` for repeated extension fields, and
+ * an empty message instance for message fields).
+ */
+export function getExtension<E extends Message<E>, V>(
```
jhump, line 33: A consequential downside of this approach that may need considerable documentation is that extension values are effectively read-only. So even if this returns a mutable value, mutating it has no effect on the original extended message.  Instead, `setExtension` must always be called, even if the top-level element (like an array or object) is unchanged but its contents are being mutated. This may be counter-intuitive compared to all other kinds of field access on the message.

packages/protobuf/src/extension-accessor.ts
```diff
@@ -0,0 +1,125 @@
...
+        .getType()
+        .runtime.bin.onUnknownField(message, uf.no, uf.wireType, uf.data);
+    }
+  }
+  const writer = writeOpt.writerFactory();
+  extension.runtime.bin.writeField(extension.field, value, writer, writeOpt);
+  const reader = readOpt.readerFactory(writer.finish());
+  while (reader.pos < reader.len) {
+    const [no, wireType] = reader.tag();
+    const data = reader.skip(wireType);
```
jhump, line 87: If wire type is "start group", how does this know the tag number to look for in the closing "end group" tag?

jhump (approved):
Very nice.

packages/protobuf-test/extra/extensions-proto2.proto
timostamm: 👍 added in https://github.com/bufbuild/protobuf-es/pull/666/commits/97b9729195c57786f330d948ccd7caaf779db4c6.

packages/protobuf/src/create-registry.ts
timostamm: That's correct, and I agree 😢

packages/protobuf/src/create-registry.ts
timostamm: Yes, a group field is `kind: "message"` with `delimited: true`. Since we generate a message class for a group, it's a decent fit. This will also work nicely with the edition feature `message_encoding = DELIMITED`.

packages/protobuf/src/extension-accessor.ts
timostamm: Good call, added an assertion for the mutating functions, but left the behavior of `hasExtension` as is: https://github.com/bufbuild/protobuf-es/pull/666/commits/c1d5b71f7e7a885ed3dcfb5ec469d00c7eaaead7

packages/protobuf/src/extension-accessor.ts
timostamm: 👍 Added a comment in https://github.com/bufbuild/protobuf-es/pull/666/commits/5fd6e9635560735be3a8c8b0c783130f3ab6fa71.

I think the documentation (a follow-up to this PR) will make this clear via examples, and I'll make sure to call it out explicitly.

packages/protobuf/src/extension-accessor.ts
timostamm: It doesn't. Method `skip` returns the bytes until (and including) the "end group" tag. It recursively skips nested groups. We're iterating over tags we just created ourselves, so we should be fine in this case.

There is a check for matching end group field number when we parse a group [here](https://github.com/bufbuild/protobuf-es/blob/5ee26a1f465af8a2ea703a9aa9ac21758a87ccbb/packages/protobuf/src/private/binary-format-common.ts#L191). IMO it would be better to check in method `skip`, so that we verify unknown fields for matching group start/end tags as well, but I'm not sure it's worth it right now, as we're already moving a lot of parts.

jhump: Ack. Maybe a TODO in `reader.skip`?

packages/protobuf/src/extension-accessor.ts
```diff
@@ -29,15 +29,19 @@ import type { Extension } from "./extension.js";
...
  */
 export function getExtension<E extends Message<E>, V>(
   message: E,
   extension: Extension<E, V>,
   options?: Partial<BinaryReadOptions>,
 ): V {
+  assertExtendee(extension, message);
   const opt = extension.runtime.bin.makeReadOptions(options);
-  assert(!extension.field.oneof); // oneof is not allowed for extensions
-  assert(extension.field.kind != "map"); // maps are not allowed for extensions
```
jhump: Were these two checks not worth keeping anywhere?

jhump (approved):
LGTM!

packages/protobuf/src/extension-accessor.ts
timostamm: These were left over from development, I forgot to remove them earlier. I expect it to very rare that users create their own extension, so we can rely on documentation and typings (makeExtension() does not allow oneof or map) here to shave off non-essential bytes from the bundle size. The assertions for the matching extendee in the accessors is a different situation, because it's very possible to pass in the wrong message without nominal typing.

PR #668, opened 2024-01-18
Add additional context about JSON

smaye81:
This adds an FAQ about JSON to our FAQ page. It also adds some additional context about initializer objects when constructing messages.

Note this PR replaces https://github.com/bufbuild/protobuf-es/pull/467 because that PR is very old and contains numerous merge conflicts. The code from 467 as well as the feedback has been incorporated here.


docs/faq.md
```diff
@@ -115,6 +115,47 @@ In the above code, `plainFoo` will be a type with only its fields and `oneOf` gr
 
 Great segue!  Our [docs](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#advanced-typescript-types) provide a good explanation for their usage and example use cases.
 
+### Why doesn't Protobuf-ES simply generate interfaces for the JSON representation?
```
timostamm, line 118: 👍

docs/runtime_api.md
```diff
@@ -69,12 +71,46 @@ new User({
...
+```typescript
+const obj = {
+  firstName: "Homer",
+  active: true,
+  manager: {
+    lastName: "Burns",
+  },
+} as PartialMessage<User>;
+
+const user = new User(obj);
```
timostamm, line 108: We should not recommend to use `as`. This is that the `case` property of a literal is typed as `string`. `as const` is the better solution here:

```suggestion
If you are not using TypeScript 4.9 yet, use `as const` instead of `satisfies PartialMessage<User>`.
```

timostamm (approved):

Issue #673, opened 2024-01-20
[feat] Option to generate enum with full prefix


vctqs1:
For example, the code below is generated from `buf`:

```
/**
 * @generated from enum usermgmt.common.ImportUserEventStatus
 */
export enum ImportUserEventStatus {
  /**
   * @generated from enum value: IMPORT_USER_EVENT_STATUS_WAITING = 0;
   */
  WAITING = 0,

  /**
   * @generated from enum value: IMPORT_USER_EVENT_STATUS_FINISHED = 1;
   */
  FINISHED = 1,

  /**
   * @generated from enum value: IMPORT_USER_EVENT_STATUS_FAILED = 2;
   */
  FAILED = 2,
}
```

However, in the backend (BE) code, the definition is as follows:

```
enum ImportUserEventStatus{
  IMPORT_USER_EVENT_STATUS_WAITING = 0;
  IMPORT_USER_EVENT_STATUS_FINISHED = 1;
  IMPORT_USER_EVENT_STATUS_FAILED = 2;
}
```
The question arises: Instead of generating and stripping out the`prefix`, should we consider adding opts to generate with the full key, such as`IMPORT_USER_EVENT_STATUS_FAILED` instead of just`FAILED`? 

This consideration is important because when storing data in the database, it uses the full name (`IMPORT_USER_EVENT_STATUS_FAILED`) instead of the stripped-out version. 

Using the full key `ImportUserEventStatus[key]` in the code makes it easier for mapping between the number and key instead of using longer methods like `proto3.getEnumType(ImportUserEventStatus).findName(<value>)`.

I also post my question in Slack before https://bufbuild.slack.com/archives/CRZ680FUH/p1699259433760749

vctqs1:
I'm willing to make a change for this.

But it's preferable to align with the owner's perspective to ensure a change without any breaking in any section of the entire code generation process

vctqs1:
Already did an PR for this proposal feature https://github.com/bufbuild/protobuf-es/pull/675

vctqs1:
@timostamm ti
hi, sorry for bothering you, but could you help a look at this issue. Thank you

timostamm:
[...]

PR #675, opened 2024-01-20
[issue#673] Adding opts to keep enum prefix

vctqs1:
Issue https://github.com/bufbuild/protobuf-es/issues/673
by adding opts `--es_opt=keep_enum_prefix=true`. by default, it's `--es_opt=keep_enum_prefix=false`


timostamm:
Thank you for the contribution! Unfortunately, we cannot add this plugin option for the [reasons outlined in the FAQ](https://github.com/bufbuild/protobuf-es/blob/main/docs/faq.md#what-is-your-stance-on-adding-options-to-the-plugin).

If it turns out that we made a mistake with shorter enum value names, we prefer to change the behavior outright, and only use the original enum value names.

Issue #663, opened 2024-01-12
Runtime error on TextEncoder/TextDecoder


Autokaka:
I'm running @bufbuild/protobuf module on a JavaScript Runtime that could only accept ECMAScript (For example, QuickJS::Eval), and the runtime throw exception that said:

```log
Error message:TextDecoder is not defined
SourceCode:
    this.textDecoder = textDecoder !== null && textDecoder !== void 0 ? textDecoder : new TextDecoder();
                                                                                          ^
Stacktrace:
    at BinaryReader (oh_modules/.ohpm/@bufbuild+protobuf@1.6.0/oh_modules/@bufbuild/protobuf/dist/esm/binary-encoding.js:281:95)
    at readerFactory (oh_modules/.ohpm/@bufbuild+protobuf@1.6.0/oh_modules/@bufbuild/protobuf/dist/esm/private/binary-format-common.js:25:31)
    at fromBinary (oh_modules/.ohpm/@bufbuild+protobuf@1.6.0/oh_modules/@bufbuild/protobuf/dist/esm/message.js:45:34)
    at fromBinary (oh_modules/.ohpm/@bufbuild+protobuf@1.6.0/oh_modules/@bufbuild/protobuf/dist/esm/google/protobuf/descriptor_pb.js:1680:16)
    at func_main_0 (oh_modules/.ohpm/@bufbuild+protobuf@1.6.0/oh_modules/@bufbuild/protobuf/dist/esm/private/feature-set.js:19:35)
```

This issue occurs at 'feature-set.ts':
```typescript
export const featureSetDefaults = FeatureSetDefaults.fromBinary(protoBase64.dec(
/*upstream-inject-feature-defaults-start*/ "ChESDAgBEAIYAiABKAEwAhjmBwoREgwIAhABGAEgAigBMAEY5wcKERIMCAEQARgBIAIoATABGOgHIOYHKOgH" /*upstream-inject-feature-defaults-end*/)); // This method uses TextDecoder...
```
The TextEncoder/TextDecoder could only be used in browser/NodeJS platform.

According to the source code, there is no way to make configurations for TextEncoder/TextDecoder.

Should we implement the TextEncoder/TextDecoder using pure JavaScript since this library can run in any ECMAScript runtime? 

Or will there be a configuration for this?

jcready:
>According to the source code, there is no way to make configurations for TextEncoder/TextDecoder.

[...]

PR #681, opened 2024-01-29
fix: don't use .fromBinary on import

### LINKED ISSUE #680: Calling FeatureSetDefaults.fromBinary in a not nodejs runtime fails importing @bufbuild/protobuf
We recently bumped our @bufbuild/protobuf dependency version and have been experiencing crashes when importing it in non node runtimes (we are using Temporal workers, which are sandboxed and don't have access to the `TextEncoder` / `TextDecoder` node apis).
The issue is the same as #663. But I believe the solution of this issue is not complete, as yes we can `globalThis.TextEncoder`, but:
- it is not very elegant, and might have side-effects
- https://github.com/bufbuild/protobuf-es/blob/db0bd03b9056ed489cb6ed53fdf9f3624d77b2fb/packages/protobuf/src/private/feature-set.ts#L25 will call .fromBinary right when importing @bufbuild/protobuf, which means the polyfill has to be applied before this import, which is not easy when using transpilled / bundled code.

I've made a fix using yarn patch for my codebase, which consists in replacing https://github.com/bufbuild/protobuf-es/blob/db0bd03b9056ed489cb6ed53fdf9f3624d77b2fb/packages/protobuf/src/private/feature-set.ts#L25 by:
```ts
export const featureSetDefaults = FeatureSetDefaults.fromJsonString(
  '{"defaults":[{"edition":"EDITION_PROTO2","features":{"fieldPresence":"EXPLICIT","enumType":"CLOSED","repeatedFieldEncoding":"EXPANDED","utf8Validation":"NONE","messageEncoding":"LENGTH_PREFIXED","jsonFormat":"LEGACY_BEST_EFFORT"}},{"edition":"EDITION_PROTO3","features":{"fieldPresence":"IMPLICIT","enumType":"OPEN","repeatedFieldEncoding":"PACKED","utf8Validation":"VERIFY","messageEncoding":"LENGTH_PREFIXED","jsonFormat":"ALLOW"}},{"edition":"EDITION_2023","features":{"fieldPresence":"EXPLICIT","enumType":"OPEN","repeatedFieldEncoding":"PACKED","utf8Validation":"VERIFY","messageEncoding":"LENGTH_PREFIXED","jsonFormat":"ALLOW"}}],"minimumEdition":"EDITION_PROTO2","maximumEdition":"EDITION_2023"}'
);
```
I obtained the JSON string by running 
```ts
FeatureSetDefaults.fromBinary(
  protoBase64.dec(
    /*upstream-inject-feature-defaults-start*/ "ChESDAgBEAIYAiABKAEwAhjmBwoREgwIAhABGAEgAigBMAEY5wcKERIMCAEQARgBIAIoATABGOgHIOYHKOgH" /*upstream-inject-feature-defaults-end*/,
  ),
).toJsonString()
```

I believe non runtime agnostic operations such as `fromBinary` & `toBinary` (and probably some more) shouldn't be called at import.

I'm not sure about why `FeatureSetDefaults` is needed and why / where from this b64 string injected to the code?

Would injecting a JSON string instead would be an acceptable solution?

This could be done by creating the FeatureSetDefaults in https://github.com/bufbuild/protobuf-es/blob/246e6df360edc3f201a7a1eb9c6a95b2327f7cae/packages/upstream-protobuf/bin/upstream-inject-feature-defaults.mjs#L24 and serializing it to a json string, then injecting this string.

Thank you :)

nohehf:
This PR fixes #680, by transforming the b64 feature set to a JSON string describing the message, at buildtime (bootstrap step / `npm run -w packages/protobuf bootstrap:featureset-defaults`).
This allows to import `@bufbuild/protobuf` in a context that doesn't provide a `TextEncoder` globally.
Please let me know if you have any feedback :)

### DIFF
--- a/packages/upstream-protobuf/bin/upstream-inject-feature-defaults.mjs
+++ b/packages/upstream-protobuf/bin/upstream-inject-feature-defaults.mjs
@@ -18,6 +18,7 @@ import { argv, exit, stdout, stderr } from "node:process";
 import { parseArgs } from "node:util";
 import { readFileSync, writeFileSync } from "node:fs";
 import { UpstreamProtobuf } from "../index.mjs";
+import { FeatureSetDefaults, protoBase64 } from "@bufbuild/protobuf";
 
 void main(argv.slice(2));
 
@@ -41,14 +42,18 @@ async function main(args) {
   const upstream = new UpstreamProtobuf();
   const defaults = await upstream.getFeatureSetDefaults(min, max);
   stdout.write(
-    `Injecting google.protobuf.FeatureSetDefaults into ${positionals.length} files...\n`,
+    `Injecting google.protobuf.FeatureSetDefaults into ${positionals.length} files...\n`
   );
   for (const path of positionals) {
     const content = readFileSync(path, "utf-8");
-    const r = inject(content, ` "${defaults.toString("base64url")}" `);
+    const base64String = defaults.toString("base64url");
+    const buffer = protoBase64.decode(base64String);
+    const featureSetDefaults = FeatureSetDefaults.fromBinary(buffer);
+    const jsonString = featureSetDefaults.toJsonString();
+    const r = inject(content, ` "${jsonString}" `);
     if (!r.ok) {
       stderr.write(`Error injecting into ${path}: ${r.message}\n`, () =>
-        exit(1),
+        exit(1)
       );
       return;
     }
@@ -99,6 +104,6 @@ function inject(content, contentToInject) {
 function exitUsage() {
   stderr.write(
     `USAGE: upstream-inject-feature-defaults [--min <mininum supported edition>] [--max <maximum supported edition>] <file-to-inject-into>\n`,
-    () => exit(1),
+    () => exit(1)
   );
 }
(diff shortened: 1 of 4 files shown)


packages/upstream-protobuf/bin/upstream-inject-feature-defaults.mjs
```diff
@@ -45,7 +46,11 @@ async function main(args) {
   );
   for (const path of positionals) {
     const content = readFileSync(path, "utf-8");
-    const r = inject(content, ` "${defaults.toString("base64url")}" `);
+    const base64String = defaults.toString("base64url");
+    const buffer = protoBase64.dec(base64String);
+    const featureSetDefaults = FeatureSetDefaults.fromBinary(buffer);
```
timostamm, line 51: Depending on our own runtime can be an issue here, because we need this package to bootstrap it 😓 

I'm not completely sure it'll be problem in practice now or in the future, but I'm sure that not using our own runtime within this helper package avoids the problem 100% 🙂

timostamm:
So I realized this was an issue when I saw https://github.com/bufbuild/protobuf-es/issues/663 last week and wrote a fix, but didn't get to pushing it up until today. You beat me to it - but https://github.com/bufbuild/protobuf-es/pull/683 avoids the dependency cycle. Could you confirm that it works for you?

packages/upstream-protobuf/bin/upstream-inject-feature-defaults.mjs
nohehf: Yep I did think about it, but as we only depend on some google types, it didn't seem to be an issue. However I do agree it's best to avoid it :)

timostamm:
Superseded by https://github.com/bufbuild/protobuf-es/pull/683, but thanks for the PR ❤️

PR #683, opened 2024-01-29
Avoid parsing feature-set defaults at module initialization

timostamm:
In [v1.5.0](https://github.com/bufbuild/protobuf-es/releases/tag/v1.5.0), we added a constant that was initialized by parsing protobuf data (see PR https://github.com/bufbuild/protobuf-es/pull/627).

This caused problems for users without a text-encoding API, because it becomes much more difficult to bring your own implementation. For context, see https://github.com/bufbuild/protobuf-es/issues/663#issuecomment-1908474133

This change avoids the init at module time by deferring to a function call. The function `createDescriptorSet` receives a new option `serializationOptions` that can be used to bring your own [binary encoding](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#binary-encoding) (including text-encoding implementation).

Fixes https://github.com/bufbuild/protobuf-es/issues/680.

### DIFF
--- a/packages/protobuf/src/create-descriptor-set.ts
+++ b/packages/protobuf/src/create-descriptor-set.ts
@@ -52,11 +52,9 @@ import {
   parseTextFormatEnumValue,
   parseTextFormatScalarValue,
 } from "./private/text-format.js";
+import type { BinaryReadOptions, BinaryWriteOptions } from "./binary-format.js";
 import type { FeatureResolverFn } from "./private/feature-set.js";
-import {
-  createFeatureResolver,
-  featureSetDefaults,
-} from "./private/feature-set.js";
+import { createFeatureResolver } from "./private/feature-set.js";
 
 /**
  * Create a DescriptorSet, a convenient interface for working with a set of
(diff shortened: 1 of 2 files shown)


nohehf:
Hey @timostamm thanks for the reactivity here :) Not sure (and curious to learn?) about what `createDescriptorSet` does / impacts, but this should work indeed. Looking through the codebase it seems that `createDescriptorSet` doesn't impact the features we use, so it should be fine anyway. We do use registries but via types (`createRegistry`) so I believe it should be fine. 
Thanks for the fix :)

timostamm:
> Not sure (and curious to learn?) about what `createDescriptorSet` does / impacts, but this should work indeed. Looking through the codebase it seems that `createDescriptorSet` doesn't impact the features we use, so it should be fine anyway. We do use registries but via types (`createRegistry`) so I believe it should be fine. 

Yes, you should be fine @nohehf, `createRegistry` does not depend on `createDescriptorSet` at all. 

`createDescriptorSet` provides a small wrapper around the descriptor Protobuf messages that are used throughout the ecosystem. You typically only interact with them/use them if you write your own plugin or create types dynamically. We just updated the documentation to explain them a bit better [here](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#reflection). 

The feature-set defaults are for the upcoming language feature ["Editions"](https://protobuf.dev/editions/overview/). It's still experimental and not fully supported in protobuf-es, but the defaults will drive the runtime behavior.

nohehf:
Thanks for the clarification & the fix @timostamm :))

PR #688, opened 2024-01-29
Fix name collision with import in generated code

timostamm:


packages/protobuf-test/extra/name-clash.proto
```diff
@@ -15,6 +15,13 @@
 syntax = "proto3";
 package spec;
 
+import "extra/example.proto";
+
+// This message class will clash with the imported message class
+message User {
+  docs.User u = 1; // Reference the import with the same name to trigger a clash
+}
```
timostamm, line 23: We trigger the issue by declaring a message, and importing a message with the same name from a different namespace.

packages/protobuf-test/src/gen/js/extra/name-clash_pb.js
```diff
@@ -17,6 +17,19 @@
...
 
 import { proto3 } from "@bufbuild/protobuf";
+import { User } from "./example_pb.js";
+
+/**
+ * This message class will clash with the imported message class
+ *
+ * @generated from message spec.User
+ */
+export const User = proto3.makeMessageType(
```
timostamm, line 27: As a result, we end up with two conflicting identifiers - the import, and the declaration.

packages/protoplugin-test/src/file-export-decl.test.ts
```diff
@@ -48,6 +48,19 @@ describe("file exportDecl", () => {
...
+      f.print(f.import("Foo", "pkg"));
+      f.print(f.exportDecl("const", "Foo"), " = 123;");
+    });
+    expect(lines).toStrictEqual([
+      `import { Foo as Foo$1 } from "pkg";`,
+      ``,
+      `Foo$1`,
+      `export const Foo = 123;`,
+    ]);
+  });
```
timostamm, line 62: Fails with:

```
    - Expected  - 2
    + Received  + 2

      Array [
    -   "import { Foo as Foo$1 } from \"pkg\";",
    +   "import { Foo } from \"pkg\";",
        "",
    -   "Foo$1",
    +   "Foo",
        "export const Foo = 123;",
      ]
```

packages/protobuf-test/src/gen/js/extra/name-clash_pb.js
```diff
@@ -17,6 +17,19 @@
...
 
 import { proto3 } from "@bufbuild/protobuf";
+import { User as User$1 } from "./example_pb.js";
+
+/**
+ * This message class will clash with the imported message class
+ *
+ * @generated from message spec.User
+ */
+export const User = proto3.makeMessageType(
```
timostamm, line 27: With the fix https://github.com/bufbuild/protobuf-es/pull/688/commits/729b9f3ac4eb79c74f934f3cda462c27fbbcac35, the import is aliased as expected.

smaye81 (approved):

PR #689, opened 2024-01-30
Fix proto3 extension presence for singular, non-optional scalar

timostamm:
When setting a proto3 extension, we apply explicit field presence for `optional` extension fields, and implicit presence otherwise. The behavior for non-optional singular scalar fields is incorrect.

For example, given the following definition:

```proto
syntax = "proto3";
import "google/protobuf/descriptor.proto";
extend google.protobuf.FileOptions {
  uint32 uint32_ext = 1001;
}
```

Setting a zero value is effectively a no-op:

```ts
setExtension(msg, uint32_ext, 0);
hasExtension(msg, uint32_ext); // false
```

This PR changes the behavior, treating the non-optional field as if it was `optional`, so that `hasExtension` in the example above returns `true`. 

For context, see https://github.com/protocolbuffers/protobuf/issues/8234

### DIFF
--- a/packages/protobuf-bench/README.md
+++ b/packages/protobuf-bench/README.md
@@ -10,5 +10,5 @@ server would usually do.
 
 | code generator      | bundle size             | minified               | compressed         |
 |---------------------|------------------------:|-----------------------:|-------------------:|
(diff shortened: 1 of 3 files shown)


jhump (approved):
LGTM

Issue #599, opened 2023-10-27
Build ESM artifacts with Node16 module resolution


timostamm:
We currently build ESM artifacts with module resolution "Node", which is the Node.js v10 resolution algorithm. 

We should switch to "Node16", which would have caught https://github.com/bufbuild/protobuf-es/issues/586.

timostamm:
Another extension-less import slipped through: https://github.com/bufbuild/protobuf-es/pull/690

PR #690, opened 2024-01-31
Add missing extension to import of descriptor-set for module resolution NodeNext

tehwalris:
This single `import` was missing the `.js` extension that all the others seem to have ([example](https://github.com/bufbuild/protobuf-es/blob/11e0f56f213870ba0c330a1cb844ca9f70cca12a/packages/protobuf/src/extension-accessor.ts#L15)). This causes TypeScript errors with `"moduleResolution": "NodeNext"`.

### DIFF
--- a/packages/protobuf/src/codegen-info.ts
+++ b/packages/protobuf/src/codegen-info.ts
@@ -29,7 +29,7 @@ import type {
   DescMethod,
   DescOneof,
   DescService,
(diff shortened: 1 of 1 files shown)


timostamm (approved):
Thanks!

Hope we get to switching our own config soon (tracked in https://github.com/bufbuild/protobuf-es/issues/599), which would have caught this.

PR #712, opened 2024-02-12
Deprecate field helper functions from @bufbuild/protoplugin

timostamm:
To support editions, and to improve support for proto2, we will have to make changes to field types in generated code, and how they are determined. 

This means that the following functions from `@bufbuild/protoplugin/ecmascript` would need to change:

- `getFieldIntrinsicDefaultValue`
- `getFieldExplicitDefaultValue`
- `getFieldTyping`

This PR deprecates the functions, and switches `protoc-gen-es` to an internal implementation that will receive future changes. The existing functionality of the deprecated functions will remain for now.


packages/protoc-gen-es/src/util.ts
```diff
@@ -0,0 +1,315 @@
...
+      return value;
+    case ScalarType.STRING:
+      if (typeof value != "string") {
+        throw new Error(
+          `Unexpected value for ${ScalarType[field.scalar]} ${field.toString()}: ${String(value)}`,
+        );
+      }
+      return literalString(value);
+    case ScalarType.BYTES:
+      if (!(value instanceof Uint8Array)) {
```
smaye81, line 235: Is it worth avoiding `instanceof` here if possible in light of https://github.com/bufbuild/protobuf-es/issues/713?

smaye81 (approved):

packages/protoc-gen-es/src/util.ts
timostamm: Uint8Array is a built-in global. It isn't ever imported from a package, and it's not subject to the dual package hazard.

PR #716, opened 2024-02-14
Support proto2 default values and field presence

timostamm:
This PR adds support for default values in proto2. For example, consider this simple Protobuf definition:

```proto
syntax="proto2";

message Example {
  optional int32 int32_field = 1;
  optional string string_field = 2 [default = "hello"];
  required bool bool_field = 3;
}
```

Note that `string_field` defines a default value, which isn't available in generated code (except through [reflection](https://github.com/bufbuild/protobuf-es/blob/0f3299629afcbc558b59cc2a2e0b31cc38ba9d3a/docs/runtime_api.md#reflection-at-runtime)).

If you have previously used a "required" field like `bool_field` in the example above, you might also have wondered why we generate an optional property `boolField?: boolean`. 

This PR changes the generated code as follows:

```diff
export class Example extends Message<Example> {
  /**
   * @generated from field: optional int32 int32_field = 1;
   */
- int32Field?: number;
+ declare int32Field: number;

  /**
   * @generated from field: optional string string_field = 2 [default = "hello"];
   */
- stringField?: string;
+ declare stringField: string;

  /**
   * @generated from field: required bool bool_field = 3;
   */
- boolField?: boolean;
+ declare boolField: boolean;
  // ...
}

+ Example.prototype.int32Field = 0;
+ Example.prototype.stringField = "hello";
+ Example.prototype.boolField = false;
```

#### Default values

If you access `string_field` from a fresh message, the property will have the default value now. If no default value is specified, the property is initialized with the zero-value for the field type - an empty string in this case, `false` for a boolean field, and so on:

```ts
const e = new Example();
e.int32Field // 0
e.stringField // "hello"
e.boolField // false
```

#### Field presence

The reason proto2 required fields were previously typed as option in generated code is [field presence](https://protobuf.dev/programming-guides/field_presence/). Proto2 fields use explicit presence tracking, which means we need to distinguish between the initial value, and a value that is explicitly set. Many Protobuf implementations use getters and setters for this purpose. We use the [prototype chain](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_prototypes) instead. 

To determine whether a field is present, we provide the function `isFieldSet`, and the function `clearField` to reset a field to its initial value. Here is a practical example:

```ts
import { isFieldSet, clearField } from "@bufbuild/protobuf";

const e = new Example();
isFieldSet(e, "int32Field"); // false

e.int32Field = 123; // setting a value makes the field present
isFieldSet(e, "int32Field"); // true

clearField(e, "int32Field"); // reset to initial value
e.int32Field // 0
```

These new functions provide a unified API to determine field presence for proto2, proto3, and - in the future - [editions](https://protobuf.dev/editions/overview/).

#### Notes

- An [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#description) will not include initial values, but [toPlainMessage](https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#plainmessage) does. Be careful when passing a plain message to a constructor – any default values in the plain message will become present.
- Protobuf `bytes` fields can have default values. You must be careful not to modify this default value, as it would affect every other instance of the message. This only applies to explicitly set default values, not to zero-values, since a `Uint8Array` with zero length cannot be mutated.
- The change for property typings only applies to scalar and enum fields, not to message fields. To do the same for message fields, we would need immutable messages, so that instances for default values can be shared. This may or may not be an option going forward, but we think that the change for primitive fields stands on its own.

### DIFF
--- a/packages/protobuf/src/private/reflect.ts
+++ b/packages/protobuf/src/private/reflect.ts
@@ -78,11 +78,7 @@ export function clearField(
         if (implicitPresence) {
           target[localName] = field.T.values[0].no;
         } else {
-          if (fieldUsesPrototype(field)) {
-            delete target[localName];
-          } else {
-            target[localName] = undefined;
-          }
+          delete target[localName];
         }
         break;
       case "scalar":
@@ -90,11 +86,7 @@ export function clearField(
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
           target[localName] = scalarZeroValue(field.T, field.L);
         } else {
-          if (fieldUsesPrototype(field)) {
-            delete target[localName];
-          } else {
-            target[localName] = undefined;
-          }
+          delete target[localName];
         }
         break;
       case "message":
@@ -102,18 +94,4 @@ export function clearField(
         break;
     }
   }
-}
-
-/**
- * Tells whether a field uses the prototype chain for field presence.
- * Behavior must match with the counterpart in @bufbuild/protoc-gen-es.
- */
-export function fieldUsesPrototype(field: FieldInfo): boolean {
-  if (field.repeated || field.oneof) {
-    return false;
-  }
-  if (field.kind != "scalar" && field.kind != "enum") {
-    return false;
-  }
-  return field.req;
 }
--- a/packages/protobuf/src/private/reflect.ts
+++ b/packages/protobuf/src/private/reflect.ts
@@ -80,3 +80,17 @@ export function clearField(
     }
   }
 }
+
+/**
+ * Tells whether a field uses the prototype chain for field presence.
+ * Behavior must match with the counterpart in @bufbuild/protoc-gen-es.
+ */
+export function fieldUsesPrototype(field: FieldInfo): boolean {
+  if (field.repeated || field.oneof) {
+    return false;
+  }
+  if (field.kind != "scalar" && field.kind != "enum") {
+    return false;
+  }
+  return field.req;
+}
--- a/packages/protobuf/src/private/reflect.ts
+++ b/packages/protobuf/src/private/reflect.ts
@@ -34,7 +34,10 @@ export function isFieldSet(
     case "scalar":
       if (field.opt || field.req) {
         // explicit presence
-        return target[localName] !== undefined;
+        return (
+          Object.prototype.hasOwnProperty.call(target, localName) &&
+          target[localName] !== undefined
+        );
       }
       // implicit presence
       if (field.kind == "enum") {
--- a/packages/protobuf/src/private/reflect.ts
+++ b/packages/protobuf/src/private/reflect.ts
@@ -70,12 +70,27 @@ export function clearField(
         target[localName] = {};
         break;
       case "enum":
-        target[localName] = implicitPresence ? field.T.values[0].no : undefined;
+        if (implicitPresence) {
+          target[localName] = field.T.values[0].no;
+        } else {
+          if (fieldUsesPrototype(field)) {
+            delete target[localName];
+          } else {
+            target[localName] = undefined;
+          }
+        }
         break;
       case "scalar":
-        target[localName] = implicitPresence
-          ? scalarZeroValue(field.T, field.L)
-          : undefined;
+        if (implicitPresence) {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+          target[localName] = scalarZeroValue(field.T, field.L);
+        } else {
+          if (fieldUsesPrototype(field)) {
+            delete target[localName];
+          } else {
+            target[localName] = undefined;
+          }
+        }
         break;
       case "message":
         target[localName] = undefined;
--- a/docs/generated_code.md
+++ b/docs/generated_code.md
@@ -134,8 +134,8 @@ foo = "";
 bar?: string;
 ```
 
-Note that all scalar fields have an intrinsic default value in proto3 syntax, unless they are marked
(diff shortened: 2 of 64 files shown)


packages/protobuf/src/private/reflect.ts
```diff
@@ -67,16 +70,45 @@ export function clearField(
         target[localName] = {};
         break;
       case "enum":
-        target[localName] = implicitPresence ? field.T.values[0].no : undefined;
+        if (implicitPresence) {
+          target[localName] = field.T.values[0].no;
```
jhump, line 74: No way to have `scalarZeroValue` compute this for enums? It would be nice to consolidate this case with the one below given how much repetition there is. (Just a thought; not a blocking a comment.)

jhump (commented):
Should there be something in the generated code that can fail-fast if the new generated code is inadvertently combined with the old runtime library?

What about the reverse case? What happens if someone inadvertently upgrades to the new runtime library but forgets to re-generate code? It seems like there needs to be _some_ affordance for this since any breakage would probably be pretty subtle and potentially inexplicable to users.

> Protobuf bytes fields can have default values. You must be careful not to modify this default value, as it would affect every other instance of the message.

Could the default value be frozen with `Object.seal()`? Or perhaps bytes fields could be updated to use a union type that allows `Uint8Array` but also some read-only type (with similar read interface and a way to convert to `Uint8Array`)?

> The change for property typings only applies to scalar and enum fields, not to message fields.

I think this is probably fine, perhaps even expected. (It is certainly similar to message types in protobuf-go, which are very different from other field types since they are pointers.)

packages/protobuf/src/private/reflect.ts
```diff
@@ -34,7 +34,10 @@ export function isFieldSet(
     case "scalar":
       if (field.opt || field.req) {
         // explicit presence
-        return target[localName] !== undefined;
+        return (
+          Object.prototype.hasOwnProperty.call(target, localName) &&
```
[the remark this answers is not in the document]
timostamm: It's best practice. eslint (most popular linter in the ecosystem) considers a direct call as an error. 

ES2022 adds the static method `Object.hasOwn` as a replacement, but we target ES2017 and can't use it yet, most likely similar to many users.

packages/protobuf/src/private/reflect.ts
timostamm: I agree that it would be nice to avoid the repetition here, but it would add enough complication for `scalarZeroValue` and the [ScalarValue](https://github.com/bufbuild/protobuf-es/blob/89777bb0a6e84c8440be0f23bc62a27071683fc9/packages/protobuf/src/scalar.ts#L86) type (both parts of the public API), that it wouldn't be a net improvement overall.

timostamm:
> Should there be something in the generated code that can fail-fast if the new generated code is inadvertently combined with the old runtime library?
> 
> What about the reverse case? What happens if someone inadvertently upgrades to the new runtime library but forgets to re-generate code? It seems like there needs to be _some_ affordance for this since any breakage would probably be pretty subtle and potentially inexplicable to users.

That would definitely be nice to have, but it's quite a bit more complex to implement than in Go. A simple function call can throw off tree-shakers, which may either remove the call, or consider it an important side-effect and opt out of tree-shaking for the entire module. 


> Could the default value be frozen with `Object.seal()`? Or perhaps bytes fields could be updated to use a union type that allows `Uint8Array` but also some read-only type (with similar read interface and a way to convert to `Uint8Array`)?

`Object.seal()` still allows to change properties (including elements in the array). `Object.freeze()` would prevent mutation, but its not supported on typed arrays. We could roll our own, but it's far from trivial because typed arrays provide access to the underlying `ArrayBuffer`, and we have to be extremely cautious to still pass `instanceof Uint8Array`. Since typed arrays are built-in types, it's quite easy to run into issues on different runtime implementations and constraints they may put on built-ins. It may be possible to pull it off, but it's a significant piece of work that'll require tests and ongoing maintenance. 

A union type has the disadvantage that it'll apply to every bytes field, and users will have to inspect and convert to `Uint8Array` every time they pass the value somewhere. To illustrate:

```ts
declare const bytesOrReadonlyBytes: Uint8Array | ReadonlyArray<number>;
const bytes: Uint8Array = "buffer" in bytesOrReadonlyBytes ? bytesOrReadonlyBytes : new Uint8Array(bytesOrReadonlyBytes);
```

This could be improved with a bespoke class for readonly bytes, or helper function that converts if necessary, but there are pitfalls around assignability that are difficult to predict with TypeScripts complex type system, the dual package hazard for a bespoke class, and it would require significant work to get right, not much different to the runtime-based sealing.

If the ecosystem or we come up with a solution for immutable typed arrays, we can bolt it on later. I do not think we should invest the necessary time into this edge case now.

> > The change for property typings only applies to scalar and enum fields, not to message fields.
> 
> I think this is probably fine, perhaps even expected. (It is certainly similar to message types in protobuf-go, which are very different from other field types since they are pointers.)

The advantage with protobuf-go is that it's possible to call methods on null pointers. `msg.GetMsgField().GetBoolField()` will return `false` even if `Msg msg_field` is unset. Since RPC response messages often contain a message field with an entity to return, this requires optional chaining in JS, as in `resp.msgField?.boolField`, and possibly more boilerplate when the value is passed somewhere that doesn't accept undefined. It's a bit of a pain point for users, especially on the fronted with functional paradigms that don't let you exit a code path.

jhump (approved):
LGTM!

PR #717, opened 2024-02-15
Add functions for field presence

timostamm:
Protobuf fields can have [default values](https://github.com/bufbuild/protobuf-es/blob/89777bb0a6e84c8440be0f23bc62a27071683fc9/docs/runtime_api.md#default-field-values). To easily determine whether a field has an actual value, this PR adds the function `isFieldSet`, and the function `clearField` to reset a field to its initial value. 

For example:

```typescript
import { isFieldSet, clearField } from "@bufbuild/protobuf";

const user = new User({
  active: true,
});

isFieldSet(user, "active"); // true
isFieldSet(user, "firstName"); // false

clearField(user, "active");
isFieldSet(user, "active"); // false
```

### DIFF
--- a/packages/protobuf/src/private/reflect.ts
+++ b/packages/protobuf/src/private/reflect.ts
@@ -60,7 +60,12 @@ export function clearField(
   if (field.repeated) {
     target[localName] = [];
   } else if (field.oneof) {
-    target[field.oneof.localName] = { case: undefined };
+    if (
+      (target[field.oneof.localName] as { case: string }).case ===
+      field.localName
+    ) {
+      target[field.oneof.localName] = { case: undefined };
+    }
   } else {
     switch (field.kind) {
       case "map":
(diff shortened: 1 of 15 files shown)


packages/protobuf/src/private/reflect.ts
```diff
@@ -60,7 +60,12 @@ export function clearField(
   if (field.repeated) {
     target[localName] = [];
   } else if (field.oneof) {
-    target[field.oneof.localName] = { case: undefined };
+    if (
+      (target[field.oneof.localName] as { case: string }).case ===
+      field.localName
+    ) {
+      target[field.oneof.localName] = { case: undefined };
+    }
```
timostamm, line 68: This internal function was added in https://github.com/bufbuild/protobuf-es/pull/702. It used to always reset a `oneof`. This wasn't an issue with how we used it, but now that it's a public API, it seems to make much more sense to only reset a `oneof` if the given field was actually selected.

jhump (approved):

PR #720, opened 2024-02-16
Apply new proto2 field presence and default values for optional fields as well

timostamm:
This widens https://github.com/bufbuild/protobuf-es/pull/716 to support default values for proto2 optional fields as well, and applies the same technique to support field presence as for proto2 required fields.

Since the well-known types in `google/protobuf/descriptor.proto` and `google/protobuf/plugin.proto` use proto2, call sites in our descriptor wrappers and plugin-framework that distinguish between set/unset fields need to be updated to use the functions for field presence added in https://github.com/bufbuild/protobuf-es/pull/717.


jhump (approved):
LGTM!

Issue #713, opened 2024-02-12
Avoid `instanceof` and remove the "node" export condition


timostamm:
In https://github.com/connectrpc/connect-es/pull/974, we override `ConnectError`'s implementation of `instanceof` to avoid the dual package hazard, which allows us to remove the "node" export condition, which causes problems with some bundlers.

We need to take a similar approach here: Avoid the dual package hazard with `instanceof`, and remove the "node" export condition. 

There are many more uses of `instanceof` here that in Connect-ES, and we will need to take a more cautious approach, and also make sure that we do not regress on performance.

PR #728, opened 2024-02-26
Introduce type guard isMessage

timostamm:
This PR introduces the function `isMesssage`:

This function is equivalent to the `instanceof` operator. For example, `isMessage(foo, MyMessage)` is the same as `foo instanceof MyMessage`, and `isMessage(foo)` is the same as `foo instanceof Message`.

Just like `instanceof`, `isMessage` narrows the type. The advantage of `isMessage` is that it compares identity by the message type name, not by class identity. This makes it robust against the dual package hazard and similar situations, where the same message is duplicated.

The next step is to replace all usages of `instanceof Message` (with the base class `Message`, but also subclasses) with the new function. While we replace the existing `instanceof` operators, we should take a look at how the change affects performance. Afterwards, we can wrap up https://github.com/bufbuild/protobuf-es/issues/713.

### DIFF
--- a/packages/protobuf/src/is-message.ts
+++ b/packages/protobuf/src/is-message.ts
@@ -1,0 +1,56 @@
+// Copyright 2021-2024 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { MessageType } from "./message-type.js";
+import type { AnyMessage } from "./message.js";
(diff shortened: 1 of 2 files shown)


packages/protobuf/src/is-message.ts
```diff
@@ -0,0 +1,56 @@
...
+ * This function is equivalent to the `instanceof` operator. For example,
+ * `isMessage(foo, MyMessage)` is the same as `foo instanceof MyMessage`, and
+ * `isMessage(foo)` is the same as `foo instanceof Message`.
+ *
+ * Just like `instanceof`, `isMessage` narrows the type. The advantage of
+ * `isMessage` is that it compares identity by the message type name, not by
+ * class identity. This makes it robust against the dual package hazard and
+ * similar situations, where the same message is duplicated.
+ */
+export function isMessage<T extends Message<T> = AnyMessage>(
```
smaye81, line 31: No idea if it's possible/feasible and I'm sure you've thought of it but: Is it possible to somehow tie this logic in with the whole `[Symbol.hasInstance]` so that we can just continue to use `instanceof`?

smaye81 (approved):

packages/protobuf/src/is-message.ts
timostamm: It's possible, but I think exporting this function is better than overriding operators.

PR #729, opened 2024-02-28
Replace instanceof Message usages 

smaye81:
This replaces the usage of `instanceof Message` (and `Message` subclasses) with the new `isMessage` function. In addition, it adds some tests to verify behavior.

`isMessage` can be used to determine whether a given object is any subtype of `Message` or is a specific
`Message` by passing the type. It is recommended to use this instead of `instanceof Message`.

**Example usage**
```typescript
import { isMessage } from "@bufbuild/protobuf";

const user = new User({
    firstName: "Homer",
});

isMessage(user);                    // true
isMessage(user, User);              // true
isMessage(user, OtherMessageType);  // false
```

View the docs [here](./docs/runtime-api.md#identifying-messages).

**Perf benchmarks:**
```
// isMessage
large google.protobuf.FileDescriptorSet (1061190 bytes) x 60.52 ops/sec ±1.13% (63 runs sampled)
tiny docs.User (4 bytes) x 3,386,238 ops/sec ±0.19% (100 runs sampled)
scalar values (102 bytes) x 630,178 ops/sec ±0.19% (96 runs sampled)
repeated scalar fields (195 bytes) x 401,749 ops/sec ±0.19% (98 runs sampled)
map with scalar keys and values (186 bytes) x 225,122 ops/sec ±0.21% (100 runs sampled)
repeated field with 1000 messages (2000 bytes) x 7,154 ops/sec ±0.42% (96 runs sampled)
map field with 1000 messages (8890 bytes) x 3,594 ops/sec ±0.21% (98 runs sampled)
```

```
// main
large google.protobuf.FileDescriptorSet (1061190 bytes) x 60.74 ops/sec ±1.27% (64 runs sampled)
tiny docs.User (4 bytes) x 3,416,634 ops/sec ±0.24% (97 runs sampled)
scalar values (102 bytes) x 627,019 ops/sec ±0.23% (99 runs sampled)
repeated scalar fields (195 bytes) x 397,073 ops/sec ±1.53% (97 runs sampled)
map with scalar keys and values (186 bytes) x 224,147 ops/sec ±0.42% (98 runs sampled)
repeated field with 1000 messages (2000 bytes) x 7,162 ops/sec ±0.39% (95 runs sampled)
map field with 1000 messages (8890 bytes) x 3,307 ops/sec ±0.20% (97 runs sampled)
```

### DIFF
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -13,6 +13,7 @@ provided by the library.
   - [Cloning messages](#cloning-messages)
   - [Comparing messages](#comparing-messages)
   - [Serializing messages](#serializing-messages)
+  - [Identifying messages](#identifying-messages)
 - [Enumerations](#enumerations)
 - [Extensions](#extensions)
   - [Extensions and JSON](#extensions-and-json)
@@ -314,6 +315,34 @@ JSON.
 To learn about serialization options and other details related to serialization, 
 see the section about [advanced serialization](#advanced-serialization).
 
+### Identifying messages
+
+To check whether a given object is any subtype of `Message` or is a specific
+type of `Message`, we recommend using the exported function `isMessage`.
+
+`isMessage` is _mostly_ equivalent to the `instanceof` operator. For
+example, `isMessage(foo, MyMessage)` is the same as `foo instanceof MyMessage`,
+and `isMessage(foo)` is the same as `foo instanceof Message`. 
+
+The advantage of `isMessage` is that it compares identity by the message type 
+name, not by class identity. This makes it robust against the dual package 
+hazard and similar situations, where the same message is duplicated.
+
+To determine if an object is any subtype of `Message`, pass that object to the
+function. To determine if an object is a specific type of `Message`, pass the 
+object as well as the type.
+
+```typescript
+import { isMessage } from "@bufbuild/protobuf";
+
+const user = new User({
+    firstName: "Homer",
+});
+
+isMessage(user);                    // true
+isMessage(user, User);              // true
+isMessage(user, OtherMessageType);  // false
+```
 
 ## Enumerations
 
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -317,8 +317,7 @@ see the section about [advanced serialization](#advanced-serialization).
 
 ### Identifying messages
 
-To check whether a given object is any subtype of `Message` or is a specific
-type of `Message`, we recommend using the exported function `isMessage`.
+To check whether a given object is a message, use the function `isMessage`.
 
 `isMessage` is _mostly_ equivalent to the `instanceof` operator. For
 example, `isMessage(foo, MyMessage)` is the same as `foo instanceof MyMessage`,
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -1100,7 +1100,7 @@ function to your users that processes this message:
 ```ts
 export function sendUser(user: PartialMessage<User>) {
   // convert partial messages into their full representation if necessary
-  const u = user instanceof User ? user : new User(user);
+  const u = isMessage(user, User) ? user : new User(user);
   // process further...
   const bytes = u.toBinary();
 }
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -317,7 +317,7 @@ see the section about [advanced serialization](#advanced-serialization).
 
 ### Identifying messages
 
-To check whether a given object is a message, use the function `isMessage`.
+To check whether a given object is a message, use the function [`isMessage`][src-isMessage].
 
 `isMessage` is _mostly_ equivalent to the `instanceof` operator. For
 example, `isMessage(foo, MyMessage)` is the same as `foo instanceof MyMessage`,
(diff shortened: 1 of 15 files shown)


smaye81:
@timostamm A few questions:

- I wasn't sure if we should change the docs in `runtime_api.md` or `migrating.md`. They make a few mentions of `instanceof` but figured I'd check to see if we wanted to even document/advertise the usage of `isMessage` there.
- There's now a [failing test](https://github.com/bufbuild/protobuf-es/pull/729/files#diff-8980637697c9046fdade2b3fb63a2a73afce62c137ad688019f688061ebe1ea9R47) in `packages/protobuf-test/src/mixing-instances.test.ts`. This is because if we swap out `instanceof` [here](https://github.com/bufbuild/protobuf-es/pull/729/files#diff-efd29ad699c25633e971135335a86cf3aae21372ff4a6a8d65b613ddcf2cf20eR122), it returns `true` now where `instanceof` returned `false` due to the mixed instances. As a result, the test now fails because a new, normalized instance is no longer created.
- Calling `isMessage` with an actual instance of `Message` returns `false` right now, which is most likely not what we want. However, I wasn't sure of the best way to determine the explicit type of `Message` in the `isMessage` function. We could maybe check that `arg` _only_ contains the properties returned by `Object.getOwnPropertyNames(Message.prototype)`?

Let me know what you think.

srikrsna-buf:
> * Calling `isMessage` with an actual instance of `Message` returns `false` right now, which is most likely not what we want. However, I wasn't sure of the best way to determine the explicit type of `Message` in the `isMessage` function. We could maybe check that `arg` _only_ contains the properties returned by `Object.getOwnPropertyNames(Message.prototype)`?

I think we can let it be as is, because `new Message()` is not usable. It expects a subclass in every one of the methods.

> * There's now a [failing test](https://github.com/bufbuild/protobuf-es/pull/729/files#diff-8980637697c9046fdade2b3fb63a2a73afce62c137ad688019f688061ebe1ea9R47) in `packages/protobuf-test/src/mixing-instances.test.ts`. This is because if we swap out `instanceof` [here](https://github.com/bufbuild/protobuf-es/pull/729/files#diff-efd29ad699c25633e971135335a86cf3aae21372ff4a6a8d65b613ddcf2cf20eR122), it returns `true` now where `instanceof` returned `false` due to the mixed instances. As a result, the test now fails because a new, normalized instance is no longer created.

I think we should remove/switch the test to use `isMessage`. We should ask users to not use `instanceOf` and depend on  `isMessage` moving forward.

srikrsna-buf:
> I think in this case we'd have to remove it entirely because right now it's testing

But we can test that the JS generated one will match the TS generated one using `isMessage` which I think was the point of the 

> That's fair, but if we do that, we should change the messaging/comments around `isMessage`

Agreed. We should just say "use to check if this is any subtype of Message or a specific Message by passing the type"

smaye81:
@srikrsna-buf updated the test and docs. We should probably still document `isMessage` somewhere, but not entirely sure of the best place. Let me know what you think so far.

srikrsna-buf:
> We should probably still document `isMessage` somewhere, but not entirely sure of the best place. Let me know what you think so far.

I think we can document it in the runtime api, is there any other place you want to put it?

smaye81:
> > We should probably still document `isMessage` somewhere, but not entirely sure of the best place. Let me know what you think so far.
> 
> I think we can document it in the runtime api, is there any other place you want to put it?

Nope, that works for me. I updated the PR with the docs.

srikrsna-buf (approved):

docs/runtime_api.md
```diff
@@ -314,6 +315,34 @@ JSON.
 To learn about serialization options and other details related to serialization, 
 see the section about [advanced serialization](#advanced-serialization).
 
+### Identifying messages
+
+To check whether a given object is any subtype of `Message` or is a specific
+type of `Message`, we recommend using the exported function `isMessage`.
```
timostamm, line 321: Just to shorten it a bit:

```suggestion
To check whether a given object is a message, use the function `isMessage`.
```

Can you link `isMessage` to the source, same as others?

packages/protobuf-test/src/mixing-instances.test.ts
```diff
@@ -37,14 +35,12 @@ describe("mixing message instances", () => {
...
+  test("matches expected message", () => {
     const test = new JS_MessageFieldMessage_TestMessage({ name: "foo" });
     const message = new TS_MessageFieldMessage({
       messageField: test,
     });
     expect(message.messageField?.name).toBe("foo");
-    expect(message.messageField).toBeInstanceOf(
-      TS_MessageFieldMessage_TestMessage,
-    );
+    expect(isMessage(message.messageField, JS_MessageFieldMessage_TestMessage));
```
timostamm, line 44: We test what happens when a message field is passed to a constructor. Is it kept as is if it's the same type?

Our understanding of "same type" changed slightly. To test more precisely, let's not use `isMessage`, but an identity check instead: 

```suggestion
    expect(message.messageField).toBe(test);
```

packages/protobuf-test/src/is-message.test.ts
```diff
@@ -0,0 +1,56 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe, expect, test } from "@jest/globals";
+import { User as TS_User } from "./gen/ts/extra/example_pb.js";
+import { User as JS_User } from "./gen/js/extra/example_pb.js";
+import { isMessage } from "@bufbuild/protobuf";
+
+describe("isMessage", () => {
```
timostamm, line 20: We're missing a test in the type system. isMessage is a type guard, and we should test whether it works.

This should cover the two use cases:

```ts
const user: unknown = new TS_User();
if (isMessage(user)) {
  expect(user.toJsonString).toBeDefined();
}
if (isMessage(user, TS_User)) {
  expect(user.firstName).toBeDefined();
}
```

packages/protobuf-test/src/is-message.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      firstName: "Homer",
+      lastName: "Simpson",
+    });
+
+    expect(isMessage(user, JS_User)).toBeTruthy();
+  });
+  test("type guard works as expected", () => {
+    const user: unknown = new TS_User();
+    if (isMessage(user)) {
+      // @ts-ignore
```
timostamm: There shouldn't be an annotation to ignore type errors here, otherwise the test is useless. Same below.

smaye81: Yeah I removed it. For some reason, vim was erroring out for me.

timostamm: You can change `user.toJsonString` to `user.toJsonString()` without issues if that's necessary for TS or the linter.

smaye81: I added an `eslint-disable` directive for unbound methods since that was failing lint. It's weird -- the vim error states that `user` is `unknown`, but compilation works fine. 🤷

timostamm: It would be much more clear to just call toJsonString and remove the linter ignore. 

`user.firstName` isn't even a method, so it should need an ignore or anything else.

timostamm (approved):
Nice, LGTM!

Can you update the PR description with an example and a link to the docs?

PR #731, opened 2024-02-29
Run CI processes against v2 branch also

smaye81:
This adds the `v2` branch to our CI workflows so that the CI processes (and the approval bypass) Github actions are run when pull requests and merges happen with that branch.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,10 +1,10 @@
 name: ci
 on:
   push:
-    branches: [main]
+    branches: [main, v2]
     tags: ['v*']
   pull_request:
-    branches: [main]
+    branches: [main, v2]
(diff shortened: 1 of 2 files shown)


timostamm (approved):

rubensf (approved):
LGTM
Do you not also want to add branch protection to the `v2` branch?
https://github.com/bufbuild/protobuf-es/settings/branches

smaye81:
> LGTM Do you not also want to add branch protection to the `v2` branch? https://github.com/bufbuild/protobuf-es/settings/branches

I can't see that link :(.  I'm assuming, though, that we would also want the same protections as `main`, but I'll let @timostamm make the final decision.

timostamm:
> Do you not also want to add branch protection to the v2 branch?

Done, TY.

PR #742, opened 2024-03-08
Add type guard for PluginOptionError

smaye81:
This adds a type guard function for `PluginOptionError` so that `instanceof` is no longer used. For additional details, see #713

### DIFF
--- a/packages/protoplugin/src/error.ts
+++ b/packages/protoplugin/src/error.ts
@@ -39,8 +39,5 @@ export function isPluginOptionError(arg: unknown): arg is PluginOptionError {
   if (!(arg instanceof Error)) {
     return false;
   }
-  if (Object.getPrototypeOf(arg) === PluginOptionError.prototype) {
-    return true;
-  }
   return arg.name === "PluginOptionError";
 }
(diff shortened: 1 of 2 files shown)


smaye81:
@timostamm couldn't write any tests for this since these aren't exported by `protoplugin` and I'm guessing we don't want to just for the sake of testing.

Also, lmk if there's additional checks for the `PluginOptionError` type.

packages/protoplugin/src/error.ts
```diff
@@ -32,3 +34,13 @@ export function reasonToString(reason: unknown): string {
...
   return String(reason);
 }
+
+export function isPluginOptionError(arg: unknown): arg is PluginOptionError {
+  if (!(arg instanceof Error)) {
+    return false;
+  }
+  if (Object.getPrototypeOf(arg) === PluginOptionError.prototype) {
+    return true;
+  }
```
timostamm, line 44: This is equivalent to `instanceof PluginOptionError`, so we cannot use it.

It's fine to simply check for `instanceof Error` and `name === "PluginOptionError"`.

timostamm (approved):

PR #744, opened 2024-03-11
Sayers/remove node export

### LINKED ISSUE #713: Avoid `instanceof` and remove the "node" export condition

smaye81:
Fixes #713

In #728, #729, and #742, we made changes that remove a common failure point for the dual package hazard -- usage of `instanceof` with custom types . As a result, the `node` exports condition for our packages should not be necessary and is actually problematic for some bundlers.

### DIFF
--- a/packages/protobuf/package.json
+++ b/packages/protobuf/package.json
@@ -10,10 +10,9 @@
   },
   "scripts": {
     "clean": "rm -rf ./dist/*",
-    "build": "npm run build:cjs && npm run build:esm && npm run build:proxy",
+    "build": "npm run build:cjs && npm run build:esm",
     "build:cjs": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
-    "build:proxy": "node ../../scripts/gen-esm-proxy.mjs .",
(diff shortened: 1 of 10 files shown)


smaye81:
@timostamm I'd love to be able to get the [import/extensions](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md) ESLint rule working so that we could catch these missing extensions. I tried numerous configurations though and can't get it to flag `.ts` files that are missing `.js` extensions in their imports. Have you looked into this in the past?

timostamm:
I didn't look into eslint. https://github.com/bufbuild/protobuf-es/issues/599 seems like the best way to ensure extensions. But we need to look whether it comes with other unexpected constraints (need to look whether `dist` changes with the new setting).

I suggest we do that as a follow up. To check for missing extensions, checking with `from "\.[a-zA-Z0-9-\/]+";$` is better than nothing.

packages/protobuf/package.json
```diff
@@ -10,10 +10,9 @@
   },
   "scripts": {
     "clean": "rm -rf ./dist/*",
-    "build": "npm run build:cjs && npm run build:esm && npm run build:proxy",
+    "build": "npm run build:cjs && npm run build:esm",
```
timostamm, line 13: Can you also delete scripts/gen-esm-proxy.mjs ?

smaye81: Omg, I swear I removed this. I must not have pushed it.

timostamm (approved):

PR #745, opened 2024-03-12
Add the DescSet type

timostamm:
Introduce the type `DescSet`, a set of descriptors with convenient accessors to look up descriptors by name.
This type is the successor to `DescriptorSet` and registries, and combines the features of both.

For example, if generated files exports the messages `User` and `Foo`, a set of the descriptors can be created with:

```ts
// before:
const reg = createRegistry( User, Foo );

// now:
const set = createDescSet( UserDesc, FooDesc );
```

To create a set of descriptors from a `google.protobuf.FileDescriptorSet` message:

```ts
declare const fileDescriptorSet: FileDescriptorSet;

// before:
const set = createDescriptorSet( fileDescriptorSet );

// now:
const set = createDescFileSet( fileDescriptorSet );
```

This creates a `DescSet` that also includes file descriptors, a `DescFileSet`. The old function `createDescriptorSet()` and the interface `DescriptorSet` are not removed by this PR, and continue to work as expected.

Descriptor sets can be merged by passing multiple input sets to the variadic function `createDescSet()` or `createDescFileSet()`:

```ts
const a = createDescSet( UserDesc, FooDesc );
const b = createDescriptorSet( fileDescriptorSet );
const set = createDescSet( a, b );
```

This feature makes it trivial to create "registries" from multiple generated file descriptors. If a descriptor is given multiple times (identified by the same type name), the one given last wins.


timostamm:
(This is based on https://github.com/bufbuild/protobuf-es/pull/743, which needs to be merged first.)

srikrsna-buf (approved):

srikrsna-buf (approved):

PR #746, opened 2024-03-13
Expose OneofInfo type

q42jaap:
`OneofInfo` is a return type of `FieldList.byMember`. To avoid constructions like `Extract<ReturnType<FieldList['byMember']>[0], { kind: 'oneof' }>;`exposing the type is harmless.


timostamm (approved):
We're usually hesitant with adding exports, but it makes sense to export `OneofInfo` along with the other `...Info` types.

PR #751, opened 2024-03-18
Add `toBinary` and `fromBinary` functions

srikrsna-buf:
Add `toBinary` and `fromBinary` functions to serialize to and from protobuf encoded bytes. Added compatibility tests with v1 for now.


packages/protobuf/src/next/from-binary.ts
```diff
@@ -0,0 +1,274 @@
...
+      break;
+  }
+}
+
+// Read a map field, expecting key field = 1, value field = 2
+function readMapEntry(
+  field: DescField & { fieldKind: "map" },
+  reader: IBinaryReader,
+  options: BinaryReadOptions,
+): [string | number, ScalarValue | MessageShape<DescMessage>] {
```
timostamm, line 129: We have to update this logic to support edition features. They apply to synthetic map entries as well. 

Can you add a TODO?

packages/protobuf/src/next/from-binary.ts
```diff
@@ -0,0 +1,274 @@
...
+  );
+}
+
+function readMessage<Desc extends DescMessage>(
+  desc: Desc,
+  reader: IBinaryReader,
+  lengthOrEndTagFieldNo: number,
+  options: BinaryReadOptions,
+  delimitedMessageEncoding?: boolean,
+): MessageShape<Desc> {
```
timostamm, line 66: We need a better function signature. 

The signature is part of the v1 public API (see [here](https://github.com/bufbuild/protobuf-es/blob/c0d0553d24a5df462c3b2c775f9b86b50a22bbe8/packages/protobuf/src/binary-format.ts#L43-L61)), and is a bit odd because it needed to be backwards compatible.

Can you add a TODO?

packages/protobuf/src/next/to-binary.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Get information for writing a scalar value.
+ *
+ * Returns tuple:
+ * [0]: appropriate WireType
+ * [1]: name of the appropriate method of IBinaryWriter
+ */
+function scalarTypeInfo(
```
timostamm, line 205: Calling the writer's method with the bracket notation is awfully slow.  The v1 has a TODO comment to remove this function. 

Can you add it here as well, or do the change right away?

packages/protobuf/src/next/to-binary.ts
```diff
@@ -0,0 +1,236 @@
...
+  message: Message,
+) {
+  const rm = reflectToBinary(reflect(message), opts);
+  if (
+    field.proto.type === FieldDescriptorProto_Type.GROUP ||
+    field.getFeatures().messageEncoding === FeatureSet_MessageEncoding.DELIMITED
+  ) {
+    writer
+      .tag(field.number, WireType.StartGroup)
+      .raw(rm)
```
timostamm, line 115: We're creating a new writer here for a a message field that uses the delimited encoding. The advantage of the delimited encoding is that we don't need to do that, because we don't need to write a length prefix. 

We can optimize this once we have more test coverage. Can you add a TODO?

timostamm (approved):

PR #753, opened 2024-03-19
Support unknown fields

srikrsna-buf:
Support unknown fields in binary serialization and reflections.


timostamm (approved):
LGTM, but please address comment below before merging.

PR #754, opened 2024-03-19
Build ESM artifacts with Node16 moduleResolution

### LINKED ISSUE #599: Build ESM artifacts with Node16 module resolution

smaye81:
Fixes #599 

This switches all packages to build ESM artifacts with Node16 module resolution. In doing so, it turns on Node16 for both `moduleResolution` and `module` by default throughout the repo. It also turns on `verbatimModuleSyntax` by default (which is why the need for all the `import type {...}` changes).

For CJS artifacts, the exceptions for `moduleResolution node10` and `verbatimModuleSyntax false` are now specified in the command script.

### DIFF
--- a/packages/protobuf-conformance/package.json
+++ b/packages/protobuf-conformance/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "clean": "rm -rf ./dist/cjs/* ./dist/types/*",
     "build": "npm run build:esm+types",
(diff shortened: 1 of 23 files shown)


timostamm (approved):
LGTM!

We can do the same for Connect, see https://github.com/connectrpc/connect-es/issues/887

PR #755, opened 2024-03-19
Fix typo in error output

stefanvanburen:
Just noticed this debugging some remote plugin output :).


smaye81 (approved):
Sweet, thank you!

PR #759, opened 2024-03-21
Add `google.protobuf.Any` helpers

srikrsna-buf:
Add `google.protobuf.Any` helpers. Ported the non-json tests from v1, will add the json ones once we add json support


timostamm (approved):
LGTM after fixing the CI failure (still one or more inline type specifiers in import statements left).

PR #760, opened 2024-03-22
Change `clone` to create a new array for unknown fields

srikrsna-buf:
Change `clone` to create a new array for unknown fields. Also adds a type for unknown fields.


timostamm (approved):
LGTM, thanks for catching this.

PR #761, opened 2024-03-22
Add `toJson` and `toJsonString` functions

srikrsna-buf:
Add `toJson` and `toJsonString` functions to serialize messages as `JsonValue`. Doesn't support extensions. Added compat tests with v1.

### DIFF
--- a/packages/protobuf/src/next/to-json.ts
+++ b/packages/protobuf/src/next/to-json.ts
@@ -73,6 +73,13 @@ export interface JsonWriteOptions {
   descSet?: DescSet;
 }
 
+/**
+ * Options for serializing to JSON.
+ */
+export interface JsonWriteStringOptions extends JsonWriteOptions {
+  prettySpaces: number;
+}
+
 // Default options for serializing to JSON.
 const jsonWriteDefaults: Readonly<JsonWriteOptions> = {
   emitDefaultValues: false,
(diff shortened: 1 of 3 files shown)


packages/protobuf/src/next/to-json.ts
```diff
@@ -0,0 +1,486 @@
...
+            p +
+            '" is irreversible',
+        );
+      }
+      return protoCamelCase(p);
+    })
+    .join(",");
+}
+
+function structToJson(val: Struct, opts: JsonWriteOptions) {
```
timostamm, line 425: It looks like we can remove `opts` from here, and also from `valueToJson` and from `listValueToJson`.

packages/protobuf/src/next/to-json.ts
```diff
@@ -0,0 +1,486 @@
...
+      return structToJson(val.kind.value, opts);
+    case "listValue":
+      return listValueToJson(val.kind.value, opts);
+    default:
+      throw new Error(`${val.$typeName} must have a value`);
+  }
+}
+
+function listValueToJson(val: ListValue, opts: JsonWriteOptions): JsonValue[] {
+  return val.values.map((v) => valueToJson(v, opts));
```
timostamm, line 456: If I'm not mistaken about `opts`:

```suggestion
  return val.values.map(valueToJson);
```

packages/protobuf/src/next/to-json.ts
```diff
@@ -0,0 +1,486 @@
...
+  Any,
+  Duration,
+  FieldMask,
+  ListValue,
+  Struct,
+  Timestamp,
+  Value,
+} from "./wkt/index.js";
+import { anyUnpack } from "./wkt/index.js";
+import { isWktWrapperDesc } from "./reflect/wkt.js";
```
timostamm, line 38: We have WKT-specific code here, in wkt/, and in reflect/.

I don't think we should everything over to wkt/ (import cycle on `JsonWriteOptions`, still need `"google.protobuf.NullValue"` here, and more exports to deal with). 

But I think it makes sense to move reflect/wkt.ts to wkt/wrappers.ts.

Not in this PR, just thinking out load. If you want to do it, go for it. If you disagree, let me know.

timostamm (commented):
Nice, just one comment, see below.

timostamm (approved):

PR #768, opened 2024-04-02
Add `{get|set|has|clear}Extension` functions

srikrsna-buf:
Add `{get|set|has|clear}Extension` functions. They behave exactly like in v1.


timostamm (commented):
Nice! 

I think we can simplify the extension container, and possibly remove it. We should definitely look into this before cutting a stable release, but this works for now.

Just two comments, see below:

PR #774, opened 2024-04-05
Support extensions in json encoding

srikrsna-buf:
Support extensions in json encoding. Uses the extension container, if we refactor the fromJson, fromBinary and toBinary we might be able to get rid of it.

### DIFF
--- a/packages/protobuf-test/src/next/json.test.ts
+++ b/packages/protobuf-test/src/next/json.test.ts
@@ -18,6 +18,8 @@ import {
   toJson,
   fromJson,
   mergeFromJson,
+  setExtension,
+  getExtension,
 } from "@bufbuild/protobuf/next";
 import type {
   MessageInitShape,
@@ -55,6 +57,10 @@ import type {
 } from "@bufbuild/protobuf";
 import { createDescSet } from "@bufbuild/protobuf/next/reflect";
 import { equals } from "@bufbuild/protobuf/next";
+import {
+  Proto2ExtendeeDesc,
+  string_ext,
+} from "../gen/ts/extra/extensions-proto2_pbv2.js";
 
 describe("json serialization", () => {
   describe("should be identical to v1", () => {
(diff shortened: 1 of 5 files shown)


packages/protobuf-test/src/next/json.test.ts
```diff
@@ -130,6 +136,23 @@ describe("json serialization", () => {
       });
     });
   });
+  describe.only("extensions", () => {
```
timostamm, line 139:
```suggestion
  describe("extensions", () => {
```

:sweat_smile:

packages/protobuf/src/next/to-json.ts
```diff
@@ -151,7 +155,25 @@ function reflectToJson(msg: ReflectMessage, opts: JsonWriteOptions): JsonValue {
       json[jsonName(f, opts)] = jsonValue;
     }
   }
-  // TODO: Add support for extensions.
+  const descSet = opts.descSet;
+  if (descSet?.getExtensionFor) {
```
timostamm, line 159: Shouldn't this be sufficient?

```suggestion
  if (descSet) {
```

packages/protobuf/src/next/to-json.ts
```diff
@@ -151,7 +155,25 @@ function reflectToJson(msg: ReflectMessage, opts: JsonWriteOptions): JsonValue {
...
+    const tagSeen = new Set<number>();
+    for (const uf of msg.getUnknown() ?? []) {
+      if (tagSeen.has(uf.no)) {
```
timostamm, line 162: Can you add a comment that explains that tags can appear multiple times for a single extension field, and this saves us from evaluating them multiple times?

packages/protobuf/src/next/index.ts
```diff
@@ -24,4 +24,9 @@ export { fromBinary, mergeFromBinary } from "./from-binary.js";
 export type { BinaryReadOptions } from "./from-binary.js";
 export * from "./to-json.js";
 export * from "./from-json.js";
-export * from "./extension-accessor.js";
+export {
+  hasExtension,
+  getExtension,
+  setExtension,
+  clearExtension,
+} from "./extension-accessor.js";
```
timostamm, line 32: Nit: Can you rename `extension-accessor.ts` to `extensions.ts`, for consistency with `fields.ts`?

timostamm (commented):
> Uses the extension container, if we refactor the fromJson, fromBinary and toBinary we might be able to get rid of it.

That would be good. Right now, edition features are resolved by walking through ancestors. This becomes very muddy with the container. 

Let's get this merged first and increase test coverage, though.

Left a few comments:

timostamm (approved):
LGTM

(unless CI fails)

PR #779, opened 2024-04-12
Port json tests

srikrsna-buf:
Port json tests from v1. Covers most of the cases. Need to add tests for groups and merge.

### DIFF
--- a/packages/protobuf-test/src/next/json.test.ts
+++ b/packages/protobuf-test/src/next/json.test.ts
@@ -52,232 +52,216 @@ import { JSTypeProto2NormalMessageDesc } from "../gen/ts/extra/jstype-proto2_pbv
 import { fromJsonString } from "@bufbuild/protobuf/next";
 
 describe(`json serialization`, () => {
-  test("scalar fields", () => {
-    assertJson(
-      ScalarValuesMessageDesc,
-      {
-        doubleField: 0.75,
-        floatField: -0.75,
-        int64Field: protoInt64.parse(-1),
-        uint64Field: protoInt64.uParse(1),
-        int32Field: -123,
-        fixed64Field: protoInt64.uParse(1),
-        fixed32Field: 123,
-        boolField: true,
-        stringField: "hello world",
-        bytesField: new Uint8Array([
+  testJson(
+    ScalarValuesMessageDesc,
+    {
+      doubleField: 0.75,
+      floatField: -0.75,
+      int64Field: protoInt64.parse(-1),
+      uint64Field: protoInt64.uParse(1),
+      int32Field: -123,
+      fixed64Field: protoInt64.uParse(1),
+      fixed32Field: 123,
+      boolField: true,
+      stringField: "hello world",
+      bytesField: new Uint8Array([
+        104, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100,
+      ]),
+      uint32Field: 123,
+      sfixed32Field: -123,
+      sfixed64Field: protoInt64.parse(-1),
+      sint32Field: -1,
+      sint64Field: protoInt64.parse(-1),
+    },
+    {
+      doubleField: 0.75,
+      floatField: -0.75,
+      int64Field: "-1",
+      uint64Field: "1",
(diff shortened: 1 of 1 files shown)


packages/protobuf-test/src/next/json.test.ts
```diff
@@ -129,13 +303,34 @@ describe("json serialization", () => {
...
+        });
+        expect(got).toStrictEqual({
+          "@type": "type.googleapis.com/spec.JsonNamesMessage",
+          "": "b",
+          a: "a",
         });
       });
     });
   });
+  // TODO: Add tests for remaining WKT.
```
timostamm, line 333: Sounds good to me to add JSON tests for WKT here, and keep other WKT-specific tests in `wkt` (v1 has them in `google/protobuf`).

packages/protobuf-test/src/next/json.test.ts
```diff
@@ -155,24 +350,14 @@ describe("json serialization", () => {
...
 });
 
-function testV1Compat<T extends Message<T>, Desc extends DescMessage>(
-  type: MessageType<T>,
-  desc: Desc,
-  init: PartialMessage<T> & MessageInitShape<Desc>,
-  options?: Partial<
-    JsonWriteOptionsV1 & JsonWriteOptions & JsonReadOptionsV1 & JsonReadOptions
-  >,
+function testJson<Desc extends DescMessage>(
```
timostamm, line 353: This works well and looks good. But ion general, let's be cautious adding functions that create tests, and prefer explicit `describe` / `test` for consistency and readability.

packages/protobuf-test/src/next/json.test.ts
```diff
@@ -17,100 +17,274 @@ import {
...
+        floatValueField: 1.3,
+        int64ValueField: protoInt64.parse(4),
+        uint64ValueField: protoInt64.parse(5),
+        int32ValueField: 6,
+        uint32ValueField: 7,
+        stringValueField: "a",
+        // TODO: uncomment this once it is fixed.
+        // bytesValueField: new Uint8Array([
+        //   104, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100,
+        // ]),
```
timostamm, line 272: LGTM after merging in b721c26 and uncommenting this.

packages/protobuf-test/src/next/json.test.ts
srikrsna-buf: Removed the test call and renamed the function.

timostamm: Unfortunately, rolling your own assertion functions with Jest has problems. See here:

```
Summary of all failing tests
 FAIL  next/json.test.js
  ● Test suite failed to run

    expect(received).toStrictEqual(expected) // deep equality

    Expected: true
    Received: {"boolValueField": true, "bytesValueField": "aGVsbG8gd29ybGQ=", "doubleValueField": 1.2, "floatValueField": 1.3, "int32ValueField": 6, "int64ValueField": "4", "stringValueField": "a", "uint32ValueField": 7, "uint64ValueField": "5"}

      462 | function assertJson(desc, init, json) {
      463 |     const msg = create(desc, init);
    > 464 |     expect(toJson(desc, msg)).toStrictEqual(true);
          |                               ^
      465 |     expect(fromJson(desc, json)).toStrictEqual(msg);
      466 | }
      467 |

      at assertJson (next/json.test.js:464:31)
```

What we want to see is the test "json serialization > scalar fields" failing, but instead the test suite errors, and does not show which test failed, and since line numbers point to the transpiled .js files, this becomes difficult to debug.

I _think_ that the previous version that created the test did not have this issue. If it doesn't it's better to revert.

Issues like this are the reason some existing tests are so verbose. V1 has `testMT()`, but I wish it didn't. It's best to stick with what Jest suggests in their docs, avoid test helpers, and use variables scoped to `describe`, with `beforeEach`, if necessary.

Can you revise?

srikrsna-buf: Done

timostamm (approved):

PR #781, opened 2024-04-16
chore: add turborepo

yordis:
## Context

Using `turbo` repo makes the entire build step much faster, going from multiple seconds to under 200ms responses with caching


timostamm:
Replacing `make` with turbo is something we want 👍 

But it will need a much detailed turbo config (see [here](https://github.com/connectrpc/connect-playwright-es/blob/main/turbo.json) for the basics). There are some challenging bits to get right with code generation and bootstrapping. Turbo is great, but it's quite easy to cache the wrong things, and end up with unreliable CI.

timostamm:
I realize you are targeting the v2 branch. TBH, this is not the right time for this change. We cannot do this before the dust has settled a bit. Closing this, but thank you for the initiative!

PR #788, opened 2024-04-18
Remove tests for V1

timostamm:
This is a purely mechanical change. We're simply deleting unused tests, which have been ported to V2.


packages/protobuf-test/src/next/perf.ts
```diff
@@ -29,15 +28,15 @@ import {
...
-    return {
-      name: `large google.protobuf.FileDescriptorSet (${data.byteLength} bytes)`,
-      fn: () => {
-        fromBinary(FileDescriptorSetDesc, data);
-      },
-    };
-  },
+  // function () {
+  //   const data = readFileSync("./descriptorset.binpb");
+  //   return {
```
srikrsna-buf, line 33: Was this intentional?

timostamm: Yes. Testing a real-world case is good, but the size of this set changes depending on the version of protoc and the proto files, so it's not a good test fixture. We can look into a better replacement that allows for a reliable comparison between v1 and v2.

srikrsna-buf (approved):

PR #790, opened 2024-04-18
Remove protobuf V1

timostamm:
There are no dependencies on the V1 runtime left. We can remove it completely.


packages/protobuf-test/src/next/reflect/names.test.ts
```diff
@@ -188,16 +185,13 @@ describe("protoCamelCase", () => {
...
-    const bytes = await upstream.compileToDescriptorSet({
-      "i.proto": `
+    const field = await compileField(`
       syntax="proto3";
       message M { 
         int32 ${name} = 1;
-      }`,
-    });
-    const protocJsonName =
-      FileDescriptorSet.fromBinary(bytes).file[0].messageType[0].field[0]
```
timostamm, line 199: Oops, that won't work without v1. The `compileField` helper achieves the same.

packages/protobuf/src/index.ts
```diff
@@ -12,93 +12,6 @@
...
-export * from "./google/protobuf/empty_pb.js";
-export * from "./google/protobuf/field_mask_pb.js";
-export * from "./google/protobuf/source_context_pb.js";
-export * from "./google/protobuf/struct_pb.js";
-export * from "./google/protobuf/timestamp_pb.js";
-export * from "./google/protobuf/type_pb.js";
-export * from "./google/protobuf/wrappers_pb.js";
+export * from "./desc-types.js";
+export * from "./next/proto-int64.js";
+export type { JsonValue } from "./next/json-value.js";
```
timostamm, line 17: There are imports for these symbols from `@bufbuild/protobuf` instead of `@bufbuild/protobuf/next`. Since we're going to remove "next" in a follow-up, there's no need to change this around.

PR #792, opened 2024-04-18
Remove the "v2" suffix from generated files

timostamm:
We use the `_pbv2` suffix for files generated by v2. This is no longer needed, and we can go back to using just the `_pb` suffix.

This removes the last remaining step we took to run both V1 and V2 in the same branch.


packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
```diff
@@ -52,7 +52,7 @@ const pure = "/*@__PURE__*/";
 // prettier-ignore
 function generateTs(schema: Schema) {
   for (const file of schema.files) {
-    const f = schema.generateFile(file.name + "_pbv2.ts");
+    const f = schema.generateFile(file.name + "_pb.ts");
```
timostamm, line 55: This controls the suffix.

packages/protoplugin/src/ecmascript/names.ts
```diff
@@ -43,7 +43,7 @@ export function generateFilePath(
   ) {
     return wktFrom;
   }
-  return "./" + file.name + "_pbv2.js";
+  return "./" + file.name + "_pb.js";
```
timostamm, line 46: This must match the suffix.

Everything else is renaming files and updating imports.

srikrsna-buf (approved):

PR #794, opened 2024-04-18
V2: Revise map field serialization with file option message_encoding = DELIMITED

timostamm:
With this change, delimited encoding applies to the map entry message, and - if the map uses a message type for values - to map values.

Whether this is the "correct" behavior is unclear as of now.


timostamm:
Superseded by https://github.com/bufbuild/protobuf-es/pull/829

PR #796, opened 2024-04-19
Automatically publish with the proper dist-tag

timostamm:
npm uses dist-tags to distinguish between the "latest" version, and other distributions. When publishing a package, the tag can be given with the flag `--tag`. When installing a package, the version tagged "latest" will be installed by default. When a tag is specified (for example `npm install foo@alpha`), the tagged version is installed.

Since we follow semver, we can trivially deduce the proper dist-tag from our workspace version: When running the target `release`, we call `npm publish --tag latest` for stable versions such as 3.1.4. For alpha versions such as 2.0.0-alpha.8, we call `npm publish --tag alpha`. Similar for beta versions and release candidates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,6 +176,7 @@ setversion: ## Set a new version in for the project, i.e. make setversion SET_VE
 release: all ## Release @bufbuild/protobuf
 	@[ -z "$(shell git status --short)" ] || (echo "Uncommitted changes found." && exit 1);
 	npm publish \
+		--tag $(shell node scripts/get-workspace-publish-tag.js) \
 		--workspace packages/protobuf \
 		--workspace packages/protoplugin \
 		--workspace packages/protoc-gen-es
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ setversion: ## Set a new version in for the project, i.e. make setversion SET_VE
 release: all ## Release @bufbuild/protobuf
 	@[ -z "$(shell git status --short)" ] || (echo "Uncommitted changes found." && exit 1);
 	npm publish \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -176,6 +176,7 @@ setversion: ## Set a new version in for the project, i.e. make setversion SET_VE
 release: all ## Release @bufbuild/protobuf
 	@[ -z "$(shell git status --short)" ] || (echo "Uncommitted changes found." && exit 1);
 	npm publish \
+		--tag $(shell node scripts/get-workspace-publish-tag.js || kill $$PPID;) \
```
timostamm, line 179: The `kill` makes sure we don't run the publish command. I've tested it, and it shows the error, and exists with a non-zero code.

scripts/get-workspace-publish-tag.js
```diff
@@ -0,0 +1,75 @@
...
+
+/*
+
+USAGE: node get-workspace-publish-tag.js
+
+Looks at the version used in the workspace, and returns one of:
+ - latest - for versions such as 3.1.4, 2.0.0, but also 0.1.2
+ - alpha - for versions such as 1.0.0-alpha.8
+ - beta - for versions such as 1.0.0-beta.8
+ - rc - for versions such as 1.0.0-rc.8
```
timostamm, line 27: This needs to be merged into v2, but then cutting an alpha will simply be `make setversion SER_VERSION=2.0.0-alpha.1 && make release`.

timostamm: Merged in https://github.com/bufbuild/protobuf-es/commit/9cf3b0bf8abe9f8abf15ffb9a18c9a1f632ba175

PR #805, opened 2024-04-22
V2: Bump supported version of TypeScript to 4.9.5

timostamm:
v4.9 was released well over two years ago.

When we first released protobuf-es, the minimum supported version of TypeScript - v4.1 - was about as old. So it seems reasonable to support a similar range of older versions with v2 of protobuf-es.


packages/protobuf-test/src/create.test.ts
```diff
@@ -671,15 +671,14 @@ describe("create()", () => {
           // @ts-expect-error expected type error
           repeatedInt64JsStringField: [protoInt64.parse(6)],
           either: {
-            // @ts-ignore -- required for older TS
             case: "oneofInt64Field",
-            // @ts-ignore -- required for older TS
+            // @ts-expect-error expected type error
```
timostamm, line 675: In tests, we can replace some `@ts-ignore` annotations with `@ts-expect-error`, since all supported versions of TS produce an error here.

smaye81 (approved):

PR #807, opened 2024-04-23
V2: Strip only last .proto suffix for DescFile.name

timostamm:


packages/protobuf/src/reflect/registry.ts
```diff
@@ -443,15 +443,15 @@ function addFile(proto: FileDescriptorProto, reg: MutableRegistry): void {
     proto,
     deprecated: proto.options?.deprecated ?? false,
     edition: getFileEdition(proto),
-    name: proto.name.replace(/\.proto/, ""),
+    name: proto.name.replace(/\.proto$/, ""),
```
timostamm, line 446: Previously, this removed the first sub-string.

packages/protobuf-test/src/reflect/registry.test.ts
```diff
@@ -542,6 +528,19 @@ describe("DescFile", () => {
...
+    test("is proto file name without .proto suffix", async () => {
+      const file = await compileFile(`syntax="proto3";`, "foo/bar/baz.proto");
+      expect(file.name).toBe("foo/bar/baz");
+    });
+    test("strips only last .proto", async () => {
+      const file = await compileFile(
+        `syntax="proto3";`,
+        "foo.proto/baz.proto.proto",
+      );
+      expect(file.name).toBe("foo.proto/baz.proto");
```
timostamm, line 541: Without the change above, `name` would be `"foo/baz.proto.proto"`.

packages/protobuf/src/reflect/registry.ts
```diff
@@ -829,7 +829,7 @@ function newField(
           keyField.scalar != ScalarType.FLOAT &&
           keyField.scalar != ScalarType.DOUBLE,
       );
-      const valueField = mapEntry.fields.find((f) => f.proto.number === 2);
+      const valueField = mapEntry.fields.find((f) => f.number === 2);
```
timostamm, line 832: We can use `DescField.number` here, no need to go through `FieldDescriptorProto`.

srikrsna-buf (approved):

PR #810, opened 2024-04-24
V2: Check for matching field numbers in StartGroup / EndGroup tags

timostamm:
When an unknown field is encountered while parsing binary, the tag is skipped. This adds a check for skipping groups, verifying that the field number of StartGroup and EndGroup matches.

### DIFF
--- a/packages/protobuf/src/from-binary.ts
+++ b/packages/protobuf/src/from-binary.ts
@@ -108,12 +108,12 @@ function readMessage(
   const unknownFields = message.getUnknown() ?? [];
   while (reader.pos < end) {
     [fieldNo, wireType] = reader.tag();
-    if (wireType == WireType.EndGroup) {
+    if (delimited && wireType == WireType.EndGroup) {
       break;
     }
     const field = message.findNumber(fieldNo);
     if (!field) {
-      const data = reader.skip(wireType);
+      const data = reader.skip(wireType, fieldNo);
       if (options.readUnknownFields) {
         unknownFields.push({ no: fieldNo, wireType, data });
       }
--- a/packages/protobuf/src/wire/binary-encoding.ts
+++ b/packages/protobuf/src/wire/binary-encoding.ts
@@ -374,10 +374,12 @@ export class BinaryReader {
   }
 
   /**
-   * Skip one element on the wire and return the skipped data.
-   * Supports WireType.StartGroup since v2.0.0-alpha.23.
+   * Skip one element and return the skipped data.
+   *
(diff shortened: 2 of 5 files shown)


packages/protobuf/src/from-binary.ts
```diff
@@ -108,12 +108,12 @@ function readMessage(
   const unknownFields = message.getUnknown() ?? [];
   while (reader.pos < end) {
     [fieldNo, wireType] = reader.tag();
-    if (wireType == WireType.EndGroup) {
+    if (delimited && wireType == WireType.EndGroup) {
```
timostamm, line 111: We should only break the loop for EndGroup if we're currently parsing a group (a.k.a. delimited message encoding).

Previously, this would behave incorrectly for a rogue EndGroup tag.

packages/protobuf/src/from-binary.ts
```diff
@@ -108,12 +108,12 @@ function readMessage(
...
       break;
     }
     const field = message.findNumber(fieldNo);
     if (!field) {
-      const data = reader.skip(wireType);
+      const data = reader.skip(wireType, fieldNo);
```
timostamm, line 116: We have to pass the field number, so that the EndGroup tag's field number can be checked.

timostamm (commented):
This should be ported to main / v1.

packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -398,10 +400,15 @@ export class BinaryReader {
         this.pos += len;
         break;
       case WireType.StartGroup:
-        // TODO check for matching field numbers in StartGroup / EndGroup tags
-        let t: WireType;
-        while ((t = this.tag()[1]) !== WireType.EndGroup) {
-          this.skip(t);
+        for (;;) {
```
jhump, line 403: Interesting syntax. Is this idiomatic for JS/TS? In languages with `while` keywords, I am used to instead seeing `while (true)`.

jhump (approved):
LGTM!

packages/protobuf/src/wire/binary-encoding.ts
timostamm: It would be idiomatic to avoid loops, which is not feasible here. I would prefer `while (true)` as well, but the linter objects the constant condition. 

It's possible that the linter can be configured to accept it, but touching the linter config is always very involved. Since we have to change the linter config to replace `make` with `turborepo`, it's better to punt on it for now. Our usage of `make` in this repo is much more unidiomatic than this loop construct.

PR #811, opened 2024-04-24
V2: Fix perf

timostamm:


packages/protobuf-test/src/perf.ts
```diff
@@ -38,19 +37,21 @@ run("Parsing binary", [
...
       active: false,
       manager: { active: true },
     });
-    const data = toBinary(UserDesc, tinyUser);
+    const data = toBinary(desc, tinyUser);
     return {
       name: `tiny docs.User (${data.byteLength} bytes)`,
       fn: () => {
-        fromBinary(FileDescriptorSetDesc, data);
+        fromBinary(desc, data);
```
timostamm, line 49: I was wondering why BinaryReader.skip was so hot 🙈

smaye81 (approved):

PR #813, opened 2024-04-25
V2: Return zero values from ReflectMessage.get

timostamm:
This changes the behavior of `ReflectMessage` to never return `undefined` with `get()`. For unset fields, we return the zero value instead. 

This simplifies call-sites, since they do not have to check for `undefined` anymore, even if the code path is guarded by a call to `isSet()`.

Besides this change of behavior, this also adds JSDoc for `ReflectMessage`, `ReflectList`, `ReflectMap`, `isFieldSet`, and a few others.

### DIFF
--- a/packages/protobuf/src/reflect/reflect-types.ts
+++ b/packages/protobuf/src/reflect/reflect-types.ts
@@ -18,48 +18,172 @@ import { unsafeLocal } from "./unsafe.js";
 import type { Message, UnknownField } from "../types.js";
 import type { LongType, ScalarValue } from "./scalar.js";
 
+/**
+ * ReflectMessage provides dynamic access and manipulation of a message.
+ */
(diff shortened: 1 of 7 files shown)


packages/protobuf/src/reflect/reflect-types.ts
```diff
@@ -102,8 +226,22 @@ export interface ReflectList<V = unknown> extends Iterable<V> {
   [unsafeLocal]: unknown[];
 }
 
+/**
+ * ReflectMap provides dynamic access and manipulation of a map field on a
+ * message.
+ *
+ * ReflectMap is iterable - you can loop throw all entries with a for...of loop.
```
smaye81, line 233: nit: `through`

smaye81 (approved):

PR #814, opened 2024-04-25
V2: Update the test file for deprecation

timostamm:
To make sure we generate the appropriate JSDoc `@deprecated` tags for the Protobuf option `deprecated`, this extends the test protos with more cases.

### DIFF
--- a/packages/protobuf-test/src/gen/js/extra/deprecation-implicit_pb.d.ts
+++ b/packages/protobuf-test/src/gen/js/extra/deprecation-implicit_pb.d.ts
@@ -16,9 +16,9 @@
 // @generated from file extra/deprecation-implicit.proto (package spec, syntax proto3)
 /* eslint-disable */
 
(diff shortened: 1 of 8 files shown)


packages/protobuf-test/src/gen/js/extra/deprecation-implicit_pb.d.ts
```diff
@@ -26,8 +27,47 @@ export declare const fileDesc_extra_deprecation_implicit: GenDescFile;
...
+    I: Empty;
+    O: Empty;
+  },
+}
+>;
+
+/**
+ * @generated from extension: int32 implicitly_deprecated_option = 2001;
+ */
+export declare const implicitly_deprecated_option: GenDescExtension<FieldOptions, number>;
```
timostamm, line 72: We're missing a `@deprecated` tag here, and also for message, enum, and file descriptors.

Improving this will be a follow-up.

smaye81 (approved):

PR #817, opened 2024-04-26
V2: Add function parentTypes() to reflect

timostamm:
```
/**
 * Returns the ancestors of a given Protobuf element, up to the file.
 */
function parentTypes(desc: AnyDesc): Parent[] 
```

### DIFF
--- a/packages/protobuf/src/reflect/nested-types.ts
+++ b/packages/protobuf/src/reflect/nested-types.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import type {
+  AnyDesc,
   DescEnum,
   DescExtension,
   DescFile,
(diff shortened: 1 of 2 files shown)


packages/protobuf/src/reflect/nested-types.ts
```diff
@@ -43,3 +48,34 @@ export function* nestedTypes(
...
+    case "field":
+    case "oneof":
+    case "rpc":
+      return desc.parent;
+    case "service":
+      return desc.file;
+    case "extension":
+    case "enum":
+    case "message":
+      return desc.parent ?? desc.file;
```
timostamm, line 79: I wonder whether we should look into using `parent: DescMessage | DescFile` instead.

srikrsna (approved):

srikrsna-buf (approved):

packages/protobuf/src/reflect/nested-types.ts
timostamm: I took a note to consider this.

PR #818, opened 2024-04-26
V2: For deprecated elements, generate a @deprecated tag on descriptors

timostamm:
Following up on https://github.com/bufbuild/protobuf-es/pull/814#discussion_r1579712319, this adds a `@deprecated` tag to generated descriptors. For example:

```proto
syntax="proto3";

message Example {
  option deprecated = true;
}
```

Already generates: 

```ts
/**
 * @generated from message Example
 * @deprecated
 */
export type DeprecatedMessage = Message<"Example"> & {};
```

With this change, the descriptor is also marked deprecated:

```ts
/**
 * Describes the message Example.
 * Use `create(ExampleDesc)` to create a new message.
 * @deprecated
 */
export const ExampleDesc: GenDescMessage<Example> = /*@__PURE__*/
  messageDesc(fileDesc_example, 0);
```

Previously, we generated line comments for descriptors, but since the `@deprecated` tag is ignored in line comments, we're switching to block comments.

This also improves implicit deprecation: If a message, extension, or enum is nested in a deprecated message, it is also generated with a `@deprecated` tag. Similar for files that are marked deprecated.


packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
```diff
@@ -233,6 +232,36 @@ function generateDts(schema: Schema) {
   }
 }
 
+function generateDescDoc(
```
timostamm, line 235: Generates the block comments for descriptors.

packages/protoplugin/src/ecmascript/jsdoc.ts
```diff
@@ -67,14 +68,16 @@ export function createJsDocTextFromDesc(desc: Exclude<AnyDesc, DescFile>) {
...
-    case "service":
-      deprecated = deprecated || desc.file.deprecated;
+    case "field":
+    case "enum_value":
+    case "rpc":
+      deprecated = desc.deprecated;
       break;
     default:
+      deprecated =
+        desc.deprecated || parentTypes(desc).some((d) => d.deprecated);
```
timostamm, line 80: Improves existing behavior for implicit deprecation.

srikrsna-buf (approved):

PR #819, opened 2024-04-26
Remove extension container

srikrsna-buf:
Remove extension container.

### DIFF
--- a/packages/protobuf/src/from-binary.ts
+++ b/packages/protobuf/src/from-binary.ts
@@ -12,13 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { DescField, DescMessage } from "./desc-types.js";
-import type { MessageShape } from "./types.js";
+import type { DescExtension, DescField, DescMessage } from "./desc-types.js";
+import type { MessageShape, UnknownField } from "./types.js";
 import type { MapEntryKey, ReflectMessage } from "./reflect/index.js";
 import { LongType, ScalarType, scalarZeroValue } from "./reflect/scalar.js";
 import type { ScalarValue } from "./reflect/scalar.js";
 import { reflect } from "./reflect/reflect.js";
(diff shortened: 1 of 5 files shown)


packages/protobuf/src/from-json.ts
```diff
@@ -233,14 +240,54 @@ function readField(
...
+        const list = {
+          add: (...items) => void listValue.push(...items),
+        } satisfies Pick<ReflectList, "add">;
+        if (field.listKind === "message") {
+          list.add = (...items) =>
+            void listValue.push(
+              items.map((v) => (v as ReflectMessage).message),
+            );
+        }
+        readListField(list, field, json, opts);
```
srikrsna-buf, line 279: We are able to reuse this be tweaking the function a little, I copied the logic in binary, not sure what works better.

packages/protobuf/src/from-binary.ts
```diff
@@ -131,9 +132,6 @@ function readMessage(
   }
 }
 
-/**
- * @private
- */
```
timostamm, line 136: Can we also unexport the function?

packages/protobuf/src/from-binary.ts
```diff
@@ -165,6 +163,67 @@ export function readField(
   }
 }
 
+/**
+ * @private
+ */
+export function readExtension(ufs: UnknownField[], field: DescExtension) {
```
timostamm, line 169: This function is private, but WDYT about applying `filterUnknownFields` within?

packages/protobuf/src/from-json.ts
timostamm: It's probably also an option to tweak reflectList a bit to accept extensions, but I'm not sure that works better either.

Let me check our test coverage and dig into this a bit.

timostamm (commented):
I'm curious what you think. To me, it doesn't look like this approach is working out in our favor 😞 Let's chat offline and regroup. The least we need to do before we can continue is to add more test coverage, IMO.

PR #822, opened 2024-04-30
Make DescField.presence available for all fields

timostamm:
The property `DescField.presence` tells whether a field tracks presence, but only for singular fields. This change makes it available for all fields. The values match with the [documentation](https://protobuf.dev/programming-guides/field_presence/), which should be helpful for users who want to take a closer look at the schema.

The property is also available for extension descriptors. It's not well documented, but implicit presence (e.g. a proto3 singular scalar field) does not apply to extensions. Otherwise, setting an extension value `0` would be ambigous.

### DIFF
--- a/packages/protobuf-test/src/json.test.ts
+++ b/packages/protobuf-test/src/json.test.ts
@@ -48,10 +48,12 @@ import {
 } from "@bufbuild/protobuf/wkt";
 import * as ext_proto2 from "./gen/ts/extra/extensions-proto2_pb.js";
 import * as ext_proto3 from "./gen/ts/extra/extensions-proto3_pb.js";
+import * as proto3_ts from "./gen/ts/extra/proto3_pb.js";
 import { OneofMessageDesc } from "./gen/ts/extra/msg-oneof_pb.js";
 import { JsonNamesMessageDesc } from "./gen/ts/extra/msg-json-names_pb.js";
 import { JSTypeProto2NormalMessageDesc } from "./gen/ts/extra/jstype-proto2_pb.js";
 import { TestAllTypesProto3Desc } from "./gen/ts/google/protobuf/test_messages_proto3_pb.js";
+import { compileMessage } from "./helpers.js";
 
 describe("JSON serialization", () => {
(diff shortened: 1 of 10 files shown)


packages/protobuf-test/src/json.test.ts
```diff
@@ -702,6 +704,116 @@ describe("extensions in JSON", () => {
   });
 });
 
+describe("JsonWriteOptions", () => {
+  describe("emitDefaultValues", () => {
```
timostamm, line 708: Adding coverage because the logic for deciding when to emit changed.

packages/protobuf/src/to-json.ts
```diff
@@ -135,19 +134,13 @@ function reflectToJson(msg: ReflectMessage, opts: JsonWriteOptions): JsonValue {
...
         throw new Error(
           `cannot encode field ${msg.desc.typeName}.${f.name} to binary: required field not set`,
         );
       }
-      if (!opts.emitDefaultValues) {
-        continue;
-      }
-      if (!canEmitFieldDefaultValue(f)) {
+      if (!opts.emitDefaultValues || f.presence !== IMPLICIT) {
+        // Fields with implicit presence omit zero values (e.g. empty string) by default
```
timostamm, line 143: The function `canEmitFieldDefaultValue` is replaced by this simple check, matching the [specified behavior](https://protobuf.dev/programming-guides/proto3/#json-options):

> Fields that don’t support presence and that have their default value are omitted by default in JSON output (for example, an implicit presence integer with a 0 value, implicit presence string fields that are empty strings, and empty repeated and map fields). 

We might want to rename the serialization option `emitDefaultValues` to `emitFieldsWithoutPresence` as a follow-up.

timostamm:
@srikrsna-buf, the `presence` and other properties of DescField/DescExtension are computed eagerly (that wasn't the case some time ago), and I can't think of a way that the kludge we use in the extension container should cause any more problems. So I think we should keep the extension container.

srikrsna-buf (approved):

PR #827, opened 2024-05-01
V2: Support option retention

timostamm:
This PR adds support for [option retention](https://protobuf.dev/programming-guides/proto3/#option-retention): Options can specify whether they should be included in generated code (making them accessible at runtime), or be only available for plugins, while generating code.

With this change, options that specify `retention = RETENTION_SOURCE` are available in plugins written on top of [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin). But they will not be included in the code generated by [protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es).


packages/protoplugin/src/ecmascript/schema.ts
```diff
@@ -216,9 +216,19 @@ function getFilesToGenerate(
...
+  // Our goal is to provide options with source retention to plugin authors.
+  // CodeGeneratorRequest.proto_file elides options with source retention for
+  // files to generate. For these files, we take the file from source_file_descriptors,
+  // which does include options with source retention.
+  const allProtoWithSourceOptions = request.protoFile.map((protoFile) => {
+    const sourceFile = request.sourceFileDescriptors.find(
+      (s) => s.name == protoFile.name,
+    );
+    return sourceFile ?? protoFile;
+  });
```
timostamm, line 228: This is backwards compatible. Old compilers will not populate `source_file_descriptors`, and we keep using `proto_file`.

packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
```diff
@@ -263,8 +263,10 @@ function generateDescDoc(
 }
 
 // prettier-ignore
-function getFileDescCall(f: GeneratedFile, file: DescFile) {
-  const info = embedFileDesc(file);
+function getFileDescCall(f: GeneratedFile, file: DescFile, schema: Schema) {
+  const sourceFile = file.proto;
+  const runtimeFile = schema.proto.protoFile.find(f => f.name == sourceFile.name);
+  const info = embedFileDesc(runtimeFile ?? sourceFile);
```
timostamm, line 269: This removes source retention options from embedded file descriptors generated by protoc-gen-es. We simply do the reverse of packages/protoplugin/src/ecmascript/schema.ts:223.

jhump: Might be worth a comment in here about why we want the version in `schema.proto.protoFile` instead of the provided `file.proto`, because it would already have omitted source-only options.

jhump (approved):
LGTM

packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
timostamm: Good call, added in https://github.com/bufbuild/protobuf-es/pull/827/commits/2ca5ebdfe34f78cab813ee7daa2be0a79bbd4a85

PR #834, opened 2024-05-09
V2: Add localName property to DescField and others

timostamm:
This replaces the function `localName` from `@bufbuild/protobuf/reflect` with properties on `DescField`, `DescOneof`, `DescEnum`, `DescEnumValue`, and `DescMethod`.

The result is a significant performance boost:

```diff
- fromBinary perf-payload.bin x 1,161 ops/sec ±0.45% (97 runs sampled)
+ fromBinary perf-payload.bin x 4,548 ops/sec ±0.54% (93 runs sampled)
```

PR #836, opened 2024-05-10
V2: Avoid JIT deoptimizations

timostamm:
Following https://github.com/bufbuild/protobuf-es/pull/834, we can make another - but smaller - performance improvement by avoiding a couple of deopts in v8:

```diff
- fromBinary perf-payload.bin x 4,548 ops/sec ±0.54% (93 runs sampled)
+ fromBinary perf-payload.bin x 5,094 ops/sec ±0.51% (97 runs sampled)
```

### DIFF
--- a/packages/protobuf-test/package.json
+++ b/packages/protobuf-test/package.json
@@ -12,7 +12,7 @@
     "generate:ts": "protoc --es_out=src/gen/ts --es_opt=ts_nocheck=false,target=ts,import_extension=.js     --proto_path=. $(buf ls-files extra) --proto_path=$(upstream-include test) $(upstream-files test) google/protobuf/type.proto",
     "generate:js": "protoc --es_out=src/gen/js --es_opt=ts_nocheck=false,target=js+dts,import_extension=.js --proto_path=. $(buf ls-files extra) --proto_path=$(upstream-include test) $(upstream-files test) google/protobuf/type.proto",
     "postgenerate": "license-header src/gen",
-    "perf": "tsx src/perf.ts benchmark 'fromBinary perf-payload.bin'",
+    "perf": "tsx src/perf.ts benchmark '.*'",
     "profile": "dexnode dist/esm/perf.js run 'fromBinary perf-payload.bin' 10000",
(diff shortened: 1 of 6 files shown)


packages/protobuf-test/package.json
```diff
@@ -13,6 +13,7 @@
     "generate:js": "protoc --es_out=src/gen/js --es_opt=ts_nocheck=false,target=js+dts,import_extension=.js --proto_path=. $(buf ls-files extra) --proto_path=$(upstream-include test) $(upstream-files test) google/protobuf/type.proto",
     "postgenerate": "license-header src/gen",
     "perf": "tsx src/perf.ts benchmark '.*'",
+    "profile": "dexnode dist/esm/perf.js run 'fromBinary perf-payload.bin' 10000",
```
timostamm, line 16: The resulting logs can be inspected with the [deopt explorer](https://github.com/microsoft/deoptexplorer-vscode). Seems worth keeping this script and dep.

packages/protobuf/src/from-binary.ts
```diff
@@ -158,19 +158,18 @@ export function readField(
       readListField(message, reader, options, field, wireType);
       break;
     case "map":
-      // eslint-disable-next-line no-case-declarations
-      const [key, val] = readMapEntry(field, reader, options);
-      message.setMapEntry(field, key, val);
+      readMapEntry(message, field, reader, options);
```
timostamm, line 161: Avoiding an eager bailout here.

packages/protobuf/src/reflect/registry.ts
```diff
@@ -792,6 +793,13 @@ function newField(
     message: undefined,
     enum: undefined,
     presence: getFieldPresence(proto, oneof, isExtension, parentOrFile),
+    listKind: undefined,
+    mapKind: undefined,
+    mapKey: undefined,
+    delimitedEncoding: undefined,
```
timostamm, line 799: Adding these properties means that the JIT can use the same map. It adds some bytes to the bundle size, but that's always the trade-off. I think it's worth it here for the >10% gain.

srikrsna (approved):

srikrsna-buf (approved):

PR #837, opened 2024-05-10
V2: Speed up reflect with classes

timostamm:
Following https://github.com/bufbuild/protobuf-es/pull/836, we can make another performance improvement by using classes for `ReflectMessage`, `ReflectList` and `ReflectMap`:

```diff
- fromBinary perf-payload.bin x 5,094 ops/sec ±0.51% (97 runs sampled)
+ fromBinary perf-payload.bin x 7,040 ops/sec ±0.67% (98 runs sampled)
```

### DIFF
--- a/packages/protobuf/src/reflect/reflect-types.ts
+++ b/packages/protobuf/src/reflect/reflect-types.ts
@@ -144,7 +144,7 @@ export interface ReflectMessage {
    */
   addListItem<Field extends DescField & { fieldKind: "list" }>(
     field: Field,
-    value: NewListItem<Field>,
+    value: ReflectAddListItemValue<Field>,
   ): FieldError | undefined;
 
   /**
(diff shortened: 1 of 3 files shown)


packages/protobuf/src/reflect/reflect-types.ts
```diff
@@ -264,12 +264,18 @@ export interface ReflectMap<K extends MapEntryKey = MapEntryKey, V = unknown>
...
 export type MapEntryKey = string | number | bigint | boolean;
 
 type enumVal = number;
 
+/**
+ * The return type of ReflectMessage.get()
+ */
 // prettier-ignore
-type ReflectGetValue<Field extends DescField = DescField> = (
+export type ReflectGetValue<Field extends DescField = DescField> = (
```
timostamm, line 278: We can't infer all types with a class implementation, so we have to export this type. Another disadvantage is that classes are more verbose, hence the small bump in bundle size.

packages/protobuf/src/reflect/reflect.ts
```diff
@@ -224,64 +278,88 @@ function assertOwn(owner: Message, member: DescField | DescOneof) {
 export function reflectList<V>(
   field: DescField & { fieldKind: "list" },
   unsafeInput?: unknown[],
+  check = true,
```
timostamm, line 281: This is new, and not final yet, see existing comment on line 51. We're not using it yet.

srikrsna-buf (approved):

PR #839, opened 2024-05-13
V2: Introduce DescMessage.field

timostamm:
In some situations, it's useful to be able to refer to a field by name. 

`DescMessage` provides all fields in the property `fields: DescField[]` - that's convenient to loop through all fields, but it's awkward to pick a field by name.

This PR adds another property to `DescMessage`:

```ts
/**
 * All fields of this message by their "localName".
 */
readonly field: Record<string, DescField>;
```

In generated code, the property is a type-safe record. You get autocomplete for field names, and a typo is a compile-time error:

```ts
import { UserDesc } from "./gen/example_pb.js";

UserDesc.field.firstName; // DescField
UserDesc.field.firstNam; // type error
```

This PR also changes the signatures of `isFieldSet` and `clearField` to make use of this new feature. Instead of having to pass in the message descriptor, the message, and the field name, both functions only require two arguments now: The message, and the field:

```ts
import { create, clearField } from "@bufbuild/protobuf";
import { UserDesc } from "./gen/example_pb.js";

const user = create(UserDesc);
clearField(user, UserDesc.field.firstName);
```

### DIFF
--- a/packages/protobuf/src/reflect/registry.ts
+++ b/packages/protobuf/src/reflect/registry.ts
@@ -543,6 +543,7 @@ function addFields(
     const oneof = findOneof(proto, allOneofs);
     const field = newField(proto, message, reg, oneof, mapEntries);
     message.fields.push(field);
+    message.field[field.localName] = field;
     if (oneof === undefined) {
       message.members.push(field);
     } else {
@@ -629,6 +630,7 @@ function addMessage(
     name: proto.name,
     typeName: makeTypeName(proto, parent, file),
     fields: [],
+    field: {},
     oneofs: [],
     members: [],
     nestedEnums: [],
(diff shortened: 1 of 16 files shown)


packages/protobuf/src/reflect/registry.ts
```diff
@@ -543,6 +543,7 @@ function addFields(
     const oneof = findOneof(proto, allOneofs);
     const field = newField(proto, message, reg, oneof, mapEntries);
     message.fields.push(field);
+    message.field[field.localName] = field;
```
timostamm, line 546: We're eagerly populating the record - it's quite simple and cheap.

srikrsna-buf (approved):

PR #841, opened 2024-05-14
Move Registry to the top level exports

timostamm:
Registries are first-class primitives, so we're moving the type `Registry` from the subpath `@bufbuild/protobuf/reflect` to the top level `@bufbuild/protobuf`.

### DIFF
--- a/packages/protobuf/src/descriptors.ts
+++ b/packages/protobuf/src/descriptors.ts
@@ -25,7 +25,7 @@ import type {
   OneofDescriptorProto,
   ServiceDescriptorProto,
 } from "./wkt/gen/google/protobuf/descriptor_pb.js";
-import type { LongType, ScalarType, ScalarValue } from "./reflect/scalar.js";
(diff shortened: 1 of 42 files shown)


packages/protobuf/src/descriptors.ts
```diff
@@ -39,6 +39,73 @@ type SupportedFieldPresence = Extract<
   | FeatureSet_FieldPresence.LEGACY_REQUIRED
 >;
 
+/**
+ * Scalar value types. This is a subset of field types declared by protobuf
+ * enum google.protobuf.FieldDescriptorProto.Type The types GROUP and MESSAGE
+ * are omitted, but the numerical values are identical.
+ */
+export enum ScalarType {
```
timostamm, line 47: `ScalarType` and `LongType` move as well - they are closely tied to `DescField`, so they now live next to the definition of `DescField`, along with similar types.

srikrsna-buf (approved):

PR #845, opened 2024-05-17
Update bundle-size benchmark

timostamm:
This updates the code size comparison from a simple table to a chart. This lets us show how bundle size grows with multiple files.

The package is renamed from `protobuf-bench` to `bundle-size`. The module used in the comparison is changed from [buf](https://buf.build/bufbuild/buf) to [googleapis](https://buf.build/googleapis/googleapis). We use a small plugin to generate the entry points to make sure we compare the same data.


packages/bundle-size/buf.gen.yaml
```diff
@@ -10,3 +10,8 @@ plugins:
   - plugin: buf.build/protocolbuffers/js:v3.21.2
     out: src/gen/google-protobuf
     opt: import_style=commonjs
+  - plugin: entrypoints
+    path: ["tsx", "./src/protoc-gen-entrypoints.ts"]
```
smaye81, line 14: Whoa 🧙‍♂️

smaye81 (approved):

PR #848, opened 2024-05-17
Prevent toBinary error with undefined fields

mnito:
I noticed that there was a subtle behavioral change introduced in version 1.8.0 that causes an error in `toBinary` when a field was set to `undefined` explicitly on a `Message` instance. At runtime this can be fairly common when setting proto fields based on the result of less type-safe functions.

I am not sure if this behavioral change was intentional or not. The proposed fix is a one-line change to use the custom `writeField` implementation that handles the `undefined` case explicitly.

### DIFF
--- a/packages/bundle-size/README.md
+++ b/packages/bundle-size/README.md
@@ -15,11 +15,11 @@ usually do. We repeat this for an increasing number of files.
 <!--- TABLE-START -->
 | code generator  | files    | bundle size             | minified               | compressed         |
 |-----------------|----------|------------------------:|-----------------------:|-------------------:|
-| protobuf-es | 1 | 79,466 b | 34,303 b | 9,800 b |
-| protobuf-es | 4 | 92,563 b | 37,277 b | 10,113 b |
-| protobuf-es | 8 | 101,904 b | 41,775 b | 10,824 b |
-| protobuf-es | 16 | 165,584 b | 67,020 b | 13,312 b |
-| protobuf-es | 32 | 343,363 b | 147,051 b | 20,063 b |
+| protobuf-es | 1 | 79,470 b | 34,316 b | 9,803 b |
+| protobuf-es | 4 | 92,567 b | 37,290 b | 10,117 b |
+| protobuf-es | 8 | 101,908 b | 41,788 b | 10,817 b |
+| protobuf-es | 16 | 165,588 b | 67,033 b | 13,328 b |
+| protobuf-es | 32 | 343,367 b | 147,064 b | 20,048 b |
 | protobuf-javascript | 1 | 339,613 b | 255,820 b | 42,481 b |
 | protobuf-javascript | 4 | 366,281 b | 271,092 b | 43,912 b |
 | protobuf-javascript | 8 | 388,324 b | 283,409 b | 45,038 b |
--- a/packages/bundle-size/chart.svg
+++ b/packages/bundle-size/chart.svg
@@ -43,14 +43,14 @@
 <text x="-10" y="294" text-anchor="end">0 KiB</text>
 </g>
 <g transform="translate(110, 20)">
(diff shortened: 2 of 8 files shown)


timostamm:
Hey Michael, thanks for the contribution. v1.8.0 added a couple of additional checks to avoid undefined behavior that can possibly lead to data corruption. 

To clarify, you have a definition like this:

```protobuf
syntax = "proto3";
message User {
  string first_name = 1;
}
```

Which generates: 

```ts
export declare class User extends Message<User> {
  firstName: string;
  // ...
```

And toBinary errors when setting the field to undefined:

```ts
const user = new User();
user.firstName = undefined as unknown as string;
user.toBinary(); // error
```

I see your point that it can be useful to skip `undefined` values in `toBinary`, but that only helps if you immediate discard the message after serialization. If the message is passed elsewhere, all application logic still has to check for `undefined`, even though the types say otherwise. So even if we supported this for `toBinary`, `toJson`, `equals`, and probably several other places, it still wouldn't be a good solution.

In my opinion, it's better to let strongly-typed messages be strongly typed, and have a separate layer to convert from unknown input. 

For the example above, you could write a function that's safe:

```ts
export function createUser(firstName: unknown) {
  const user = new User();
  if (typeof firstName == "string") {
    user.firstName = firstName;
  }
  return user;
}
```

If you only need to handle `undefined`, it's even simpler:

```ts
declare const firstName: string | undefined;
new User({
  firstName, // the constructor skips fields with value undefined
})
```

Because of the issues described above, we do not want to support setting field values that do not match typings, and recommend to validate inputs before assigning to a message.

mnito:
> Hey Michael, thanks for the contribution. v1.8.0 added a couple of additional checks to avoid undefined behavior that can possibly lead to data corruption.
> 
> To clarify, you have a definition like this:
> 
> ```proto
> syntax = "proto3";
> message User {
>   string first_name = 1;
> }
> ```
> 
> Which generates:
> 
> ```ts
> export declare class User extends Message<User> {
>   firstName: string;
>   // ...
> ```
> 
> And toBinary errors when setting the field to undefined:
> 
> ```ts
> const user = new User();
> user.firstName = undefined as unknown as string;
> user.toBinary(); // error
> ```
> 
> I see your point that it can be useful to skip `undefined` values in `toBinary`, but that only helps if you immediate discard the message after serialization. If the message is passed elsewhere, all application logic still has to check for `undefined`, even though the types say otherwise. So even if we supported this for `toBinary`, `toJson`, `equals`, and probably several other places, it still wouldn't be a good solution.
> 
> In my opinion, it's better to let strongly-typed messages be strongly typed, and have a separate layer to convert from unknown input.
> 
> For the example above, you could write a function that's safe:
> 
> ```ts
> export function createUser(firstName: unknown) {
>   const user = new User();
>   if (typeof firstName == "string") {
>     user.firstName = firstName;
>   }
>   return user;
> }
> ```
> 
> If you only need to handle `undefined`, it's even simpler:
> 
> ```ts
> declare const firstName: string | undefined;
> new User({
>   firstName, // the constructor skips fields with value undefined
> })
> ```
> 
> Because of the issues described above, we do not want to support setting field values that do not match typings, and recommend to validate inputs before assigning to a message.

That's fair -- thanks for the detailed response! I implemented a workaround on our end for one of our shared libraries that uses Protobuf ES for serialization -- upgrading to 1.8.0 made some of our tests fail. We did a somewhat automated refactor from our former JS proto library which used setters and getters to Protobuf ES, so we have a lot of cases where we are not using the constructor and working with some less type-safe code in a large codebase.

I think a similar change may be worth considering a la Postel's law which I think is particularly important for libraries. Our use case does pretty much always discard messages after using them for serialization, so a change in the core library may help some folks avoid production issues after upgrading past 1.8.0. Fortunately our tests caught it, so the workaround I implemented in our wrapper library gives me peace of mind.

This library is awesome and really helped us address some tech debt -- appreciate all the good work from you and the team on Protobuf ES! It really was a game changer.

PR #849, opened 2024-05-22
V2: Fix JSON error message

timostamm:


packages/protobuf/src/to-json.ts
```diff
@@ -140,7 +140,7 @@ function reflectToJson(msg: ReflectMessage, opts: JsonWriteOptions): JsonValue {
     if (!msg.isSet(f)) {
       if (f.presence == LEGACY_REQUIRED) {
         throw new Error(
-          `cannot encode field ${msg.desc.typeName}.${f.name} to binary: required field not set`,
+          `cannot encode field ${msg.desc.typeName}.${f.name} to JSON: required field not set`,
```
timostamm, line 143: The fix.

packages/protobuf-test/extra/msg-json-names.proto
```diff
@@ -17,12 +17,10 @@ package spec;
...
-    string a = 3;
-    string b = 4 [json_name = ""];
-    string c = 5 [json_name = "@type"];
-
+  string scalar_field = 1 [json_name = "scalarFieldJsonName"];
+  repeated string repeated_scalar_field = 2 [json_name = "repeatedScalarFieldJsonName"];
+  string a = 3;
+  string b = 4 [json_name = ""];
+  string c = 5 [json_name = "@type"];
+  string d = 6 [json_name = "1d"];
```
timostamm, line 25: This is an unrelated change. Proto field names cannot start with a digit (it's an error in the compiler), but there's no such restriction on `json_name`. JSON itself doesn't have that restriction either, but something like generated JSON types will have to quote the property. I'm adding this case here in advance.

srikrsna-buf (approved):

PR #850, opened 2024-05-22
V2: Use the toString method in JSON error messages

timostamm:
We're constructing full names for fields when we create error messages. We can simply use `DescField.toString` instead.

### DIFF
--- a/packages/protobuf/src/from-json.ts
+++ b/packages/protobuf/src/from-json.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/* eslint-disable no-case-declarations */
+/* eslint-disable no-case-declarations,@typescript-eslint/restrict-template-expressions */
 
 import {
   type DescEnum,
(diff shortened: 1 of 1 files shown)


packages/protobuf/src/from-json.ts
```diff
@@ -702,7 +690,7 @@ function timestampFromJson(timestamp: Timestamp, json: JsonValue) {
     ms > Date.parse("9999-12-31T23:59:59Z")
   ) {
     throw new Error(
-      `cannot decode message ${timestamp.$typeName} from JSON: must be from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z inclusive`,
+      `cannot decode ${timestamp.$typeName} from JSON: must be from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z inclusive`,
```
timostamm, line 693: We're omitting `message` in all other places for the wkt. Changing it here for consistency.

srikrsna-buf (approved):

PR #861, opened 2024-05-29
Co-locate assertInt32 and related functions

timostamm:
As a follow-up to https://github.com/bufbuild/protobuf-es/pull/859, this PR co-locates `assertInt32`, `assertUInt32`, and `assertFloat32`. The functions move to packages/protobuf/src/wire/binary-encoding.ts, the only call site.

As an aside, this PR also improves `toJson` error messages by using reflect's `checkField` instead of calls to `assert`. This allows us to co-locate the `assert` function as well, and avoid several odd cross-subpath imports.


packages/protobuf/src/codegenv1/scalar.ts
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ScalarType } from "../descriptors.js";
+
+/**
+ * Return the TypeScript type (as a string) for the given scalar type.
+ */
+export function scalarTypeScriptType(
```
timostamm, line 20: This function moves from `@bufbuild/protobuf/reflect` to `@bufbuild/protobuf/codegenv1`. IMO it's a much better fit because the function is intended to be used for generating code.

packages/protobuf-test/src/json.test.ts
```diff
@@ -1100,6 +1100,58 @@ describe("JSON parse errors", () => {
...
 
+describe("JSON serialize errors", () => {
+  const desc = proto3_ts.Proto3MessageDesc;
+  test("enum field not a number throws", () => {
+    const msg = create(desc);
+    // @ts-expect-error TS2322
+    msg.singularEnumField = "abc";
+    expect(() => toJson(desc, msg)).toThrow(
+      /^cannot encode enum spec.Proto3Enum to JSON: expected number, got "abc"$/,
+    );
```
timostamm, line 1111: Before, this would raise an error with an empty message. Since users have run into this case a couple of times, it seems worth to have better error messages for this and similar cases.

packages/protobuf/src/reflect/reflect-check.ts
```diff
@@ -106,6 +121,94 @@ function checkSingular(
   return isReflectMessage(value, field.message);
 }
 
+type InvalidScalarValueErr = false | "invalid UTF8" | `${string} out of range`;
+
+function checkScalarValue(
```
timostamm, line 126: Previously, this function was exported from packages/protobuf/src/codegenv1/scalar.ts. This module seems to be a better fit for this function, since it is the only call-site. Same for function scalarTypeDescription.

packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -71,6 +70,31 @@ export enum WireType {
   Bit32 = 5,
 }
 
+/**
+ * Maximum value for a 32-bit floating point value (Protobuf FLOAT).
+ */
+export const FLOAT32_MAX = 3.4028234663852886e38;
```
timostamm, line 76: This is now the only place where these constants are defined.

srikrsna-buf (approved):

PR #863, opened 2024-05-29
V2: Rename JSON serialization option emitDefaultValues

timostamm:
The [documentation for the suggested JSON options](https://protobuf.dev/programming-guides/proto3/#json-options) has been updated recently. The documentation originally only applied to proto3, where the name `emitDefaultValues` fits. With Editions, this name has become ambiguous: The option does not actually emit field default values (which would change data with a serialization roundtrip), it only emits empty maps and lists, and zero values for proto3 singular scalar fields (fields with implicit presence). 

This PR renames the option to `alwaysEmitImplicit`, and updates the documentation accordingly:

```ts
/**
 * By default, fields with implicit presence are not serialized if they are
 * unset. For example, an empty list field or a proto3 int32 field with 0 is
 * not serialized. With this option enabled, such fields are included in the 
 * output.
 */
alwaysEmitImplicit: boolean;
```

### DIFF
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -987,7 +987,7 @@ Options for `Message.toJson` and `Message.toJsonString`:
 
 - `alwaysEmitImplicit?: boolean`<br/>
   By default, fields with implicit presence are not serialized if they are
-  unset. For example, an empty list field or a proto3 in32 field with 0 is
+  unset. For example, an empty list field or a proto3 int32 field with 0 is
   not serialized. With this option enabled, such fields are included in the 
   output.
 - `enumAsInteger?: boolean`<br/>
--- a/docs/runtime_api.md
+++ b/docs/runtime_api.md
@@ -988,7 +988,8 @@ Options for `Message.toJson` and `Message.toJsonString`:
 - `alwaysEmitImplicit?: boolean`<br/>
   By default, fields with implicit presence are not serialized if they are
   unset. For example, an empty list field or a proto3 in32 field with 0 is
-  not serialized. With this option, such fields are included in the output.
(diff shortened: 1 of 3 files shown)


docs/runtime_api.md
```diff
@@ -985,10 +985,11 @@ Options for `Message.fromJson` and `Message.fromJsonString`:
 
 Options for `Message.toJson` and `Message.toJsonString`:
 
-- `emitDefaultValues?: boolean`<br/>
-  Fields with default values are omitted by default in JSON output.
-  This option overrides this behavior and outputs fields with
-  their default values.
+- `alwaysEmitImplicit?: boolean`<br/>
+  By default, fields with implicit presence are not serialized if they are
+  unset. For example, an empty list field or a proto3 in32 field with 0 is
```
smaye81, line 990:
```suggestion
  unset. For example, an empty list field or a proto3 int32 field with 0 is
```

packages/protobuf/src/to-json.ts
```diff
@@ -55,11 +55,12 @@ const IMPLICIT: FeatureSet_FieldPresence.IMPLICIT = 2;
  */
 export interface JsonWriteOptions {
   /**
-   * Emit fields with default values: Fields with default values are omitted
-   * by default in proto3 JSON output. This option overrides this behavior
-   * and outputs fields with their default values.
+   * By default, fields with implicit presence are not serialized if they are
+   * unset. For example, an empty list field or a proto3 in32 field with 0 is
```
smaye81, line 59:
```suggestion
   * unset. For example, an empty list field or a proto3 int32 field with 0 is
```

smaye81 (approved):

PR #866, opened 2024-05-30
V2: JSON Types

timostamm:
This PR adds an advanced feature with the plugin option `json_types=true`. If enabled, `protoc-gen-es` generates a JSON type for every Protobuf message.

Given this definition:

```proto
syntax = "proto3";

message Example {
  int32 amount = 1;
  bytes data = 2;
}
```

The following additional export is generated:

```ts
/**
 * JSON type for the message Example.
 */
export type ExampleJson = {
  /**
   * @generated from field: int32 amount = 1;
   */
  amount?: number;

  /**
   * @generated from field: bytes data = 2;
   */
  data?: string;
};
```

The JSON type matches exactly what `toJson()` will emit with standard serialization options, and `toJson()` will actually automatically return the JSON type if available:

```ts
const example = create(ExampleSchema, { amount: 123 });
const json: ExampleJson = toJson(ExampleSchema, example);

// Without json_types=true, the following would be a type error:
json.amount // number | undefined
json.data // string | undefined
```

For enumerations, a similar mechanism applies. We generate a union type with all JSON string values for the enum:

```proto
syntax = "proto3";

enum Format {
  FORMAT_UNSPECIFIED = 0;
  FORMAT_BINARY = 1;
  FORMAT_JSON = 2;
}
```

```ts
/**
 * JSON type for the enum Format.
 */
export type FormatJson = "FORMAT_UNSPECIFIED" | "FORMAT_BINARY" | "FORMAT_JSON";
```

With the functions `enumToJson()` and `enumFromJson()`, values can be converted between both representations. With `isEnumJson()`, unknown input can be narrowed down to known values. If JSON types are available, the functions are type safe:

```ts
const strVal: FormatJson = enumToJson(FormatSchema, Format.BINARY);

const enumVal: Format = enumFromJson(FormatSchema, strVal);

const someString: string = "FORMAT_BINARY";
if (isEnumJson(FormatSchema, someString)) {
  someString // FormatJson
}
```


packages/protobuf-test/src/gen/ts/extra/example_pb.ts
```diff
@@ -61,10 +61,45 @@ export type User = Message<"docs.User"> & {
...
 };
 
+/**
+ * JSON type for the message docs.User.
+ */
+export type UserJson = {
+  /**
+   * @generated from field: string first_name = 1;
+   */
+  firstName?: string;
```
timostamm, line 71: A simple example for a JSON type for a message.

mikeki:
Will there be something similar for v3 as well?

timostamm:
@mikeki, V2 refers to version 2 of this project. You can try it out here: https://github.com/bufbuild/protobuf-es/releases/tag/v2.0.0-alpha.1

srikrsna-buf (approved):

PR #874, opened 2024-06-03
V2: Add `method` field to `DescMethod`; Improve generated service types

srikrsna-buf:
Add `method` to `DescMethod` similar to `field` in `DescMessage`. This gives users easy access to method descriptors using their `localName`. 

This also aligns the generated types to match the value of this new field, the resulting type a is a strongly typed method descriptor.

### DIFF
--- a/packages/protobuf/src/codegenv1/types.ts
+++ b/packages/protobuf/src/codegenv1/types.ts
@@ -75,7 +75,7 @@ export type GenDescExtension<
  *
  * @private
  */
-export type GenDescService<RuntimeShape extends GenDescServiceShape> = Omit<
+export type GenDescService<RuntimeShape extends GenDescServiceMethods> = Omit<
   DescService,
   "method"
 > & {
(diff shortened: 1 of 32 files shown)


packages/protobuf/src/codegenv1/types.ts
```diff
@@ -74,17 +75,21 @@ export type GenDescExtension<
...
  * @private
  */
 export type GenDescServiceShape = {
   [localName: string]: {
-    kind: "unary" | "server_streaming" | "client_streaming" | "bidi_streaming";
-    I: DescMessage;
-    O: DescMessage;
+    input: DescMessage;
+    output: DescMessage;
+    methodKind: DescMethod["methodKind"];
```
timostamm, line 92: Since `GenDescServiceShape` really only defines methods, WDYT about renaming it, for example:

```ts
export type GenDescServiceMethods = Record<string, Pick<DescMethod, "input" | "output" | "methodKind">>;
```

BTW: I think we can remove the TODOs on line 83 and 84 in packages/protobuf/src/types.ts now.

timostamm (commented):
This seems to work out well!

timostamm (approved):
LGTM!

Issue #763, opened 2024-03-27
Upgrade TypeScript to v5 in `@bufbuild/protoplugin` to reduce install size for users


mcmxcdev:
`typescript@4.5.2` clocks in at 62 MB disk size after installation (biggest dependency in our `node_modules`) and leads to our project having both TypeScript v5 and v4 installed in parallel.

Please upgrade the dependency to reduce load on the end users, see: https://github.com/bufbuild/protobuf-es/blob/main/packages/protoplugin/package.json#L41

timostamm:
The dependency is used to transpile TypeScript code to the targets `js` and `dts`. This mechanism is used when a plugin only provides a generator function for TypeScript.

We're pinned to this specific version of TypeScript because TypeScript makes breaking changes in minor versions. Updating is likely to break users relying on the mechanism, and it would not solve the issue since we still have to pin a version, and end users will still end up with two versions of TypeScript installed if the versions don't match. We cannot relax the constraint because this means code generation results become unstable, depending on the version of TypeScript installed by the end user.

Going forward, we will change the behavior, and either 1) let plugin authors opt in to transpilation by bringing their own version of TypeScript, or 2) update to a later and smaller version of TypeScript, possibly bundled. Since this is a breaking change, it will need to wait for v2 (which we are already working on).

mcmxcdev:
Thanks for the detailed explanation, it makes sense!

Both proposed options sound reasonable, glad to hear that I asked for this in time for v2.

timostamm:
We've looked into making the `typescript` dependency optional, but couldn't find a good way to do that without hurting the experience for users who want to write their own plugin. 

We've update to TypeScript 5.4.5 instead in #879, which has a much smaller install size ([32.4MB](https://www.npmjs.com/package/typescript/v/5.4.5?activeTab=code) instead of 62).

This change applies to the [upcoming v2](https://github.com/bufbuild/protobuf-es/releases/tag/v2.0.0-alpha.1) because it is a breaking change.

PR #879, opened 2024-06-05
V2: Bump `typescript` version in protoplugin

srikrsna-buf:
Bump the `typescript` version for protoplugin. This reduces the download size, addressing #763.

### DIFF
--- a/packages/protoplugin/src/transpile.ts
+++ b/packages/protoplugin/src/transpile.ts
@@ -45,7 +45,7 @@ const defaultOptions: ts.CompilerOptions = {
 
   // Language and Environment
   lib: [],
-  moduleDetection: "force",
+  moduleDetection: ts.ModuleDetectionKind.Force,
   target: ts.ScriptTarget.ES2017,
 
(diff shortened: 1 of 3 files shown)


packages/protoplugin/src/transpile.ts
```diff
@@ -45,7 +45,7 @@ const defaultOptions: ts.CompilerOptions = {
 
   // Language and Environment
   lib: [],
-  moduleDetection: "force",
+  moduleDetection: ts.ModuleDetectionKind.Force,
```
timostamm, line 48: We refer to enum value `ts.ModuleResolutionKind.NodeJs` on line 32. The option was renamed to `Node10` and `NodeJs` is deprecated. Can you change it to `Node10`?

srikrsna-buf: Odd, my typings don't show that option or mark this as deprecated

timostamm (approved):

timostamm (approved):

PR #880, opened 2024-06-05
V2: Improve parameter parsing

srikrsna-buf:
Improve parameter parsing by passing the parsed options to the generate functions. This also provides the EcmaScript options in generate.

### DIFF
--- a/packages/protoplugin/src/create-es-plugin.ts
+++ b/packages/protoplugin/src/create-es-plugin.ts
@@ -29,6 +29,7 @@ import type { FileInfo } from "./generated-file.js";
 import type { Plugin } from "./plugin.js";
 import { transpile } from "./transpile.js";
 import { parseParameter } from "./parameter.js";
+import type { RawPluginOptions } from "./parameter.js";
 
 interface PluginInit<Options extends object> {
   /**
@@ -45,7 +46,7 @@ interface PluginInit<Options extends object> {
    * An optional parsing function which can be used to parse your own plugin
    * options.
    */
-  parseOptions?: (rawOptions: { key: string; value: string }[]) => Options;
+  parseOptions?: (rawOptions: RawPluginOptions) => Options;
 
   /**
    * The earliest edition supported by this plugin. Defaults to the minimum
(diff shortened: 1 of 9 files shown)


packages/protoplugin/src/schema.ts
```diff
@@ -61,6 +64,11 @@ export interface Schema {
    */
   readonly targets: readonly Target[];
 
+  /**
+   * Parsed plugin options. They include the default options available by all ecmascript plugins.
+   */
+  readonly options: Options & EcmaScriptPluginParameters;
```
timostamm, line 70:
```suggestion
  /**
   * Parsed plugin options. They include the standard options for all
   * plugins, and options parsed by your plugin.
   */
  readonly options: Options & EcmaScriptPluginParameters;
```

packages/protoplugin/src/create-es-plugin.ts
```diff
@@ -45,7 +45,7 @@ interface PluginInit {
    * An optional parsing function which can be used to parse your own plugin
    * options.
    */
-  parseOption?: (key: string, value: string) => void;
+  parseOptions?: (rawOptions: { key: string; value: string }[]) => Options;
```
timostamm, line 48: What do you think about adding an optional second argument with the standard options to this function?

```ts
(rawOptions: { key: string; value: string }[], standardOptions?: EcmaScriptPluginOptions) => Options
```

To be fair, I can't think of a good use case right now. We can add it once someone requests it.

---

It would be nice to document what `key` and `value` are, and - most importantly - that a `parseOption` function should raise an error when it does not recognize an option.

We could export a type for the raw options in parameter.ts:

```ts
/**
 * Raw options to parse.
 *
 * For example, if a plugin is run with the options foo=123,bar,baz=a,baz=b
 * the raw options are:
 *
 * ```ts
 * [
 *   { key: "foo", value: "123" },
 *   { key: "bar", value: "" },
 *   { key: "baz", value: "a" },
 *   { key: "baz", value: "b" },
 * ]
 * ```
 *
 * If your plugin does not recognize an option, it must throw an Error in
 * parseOptions.
 */
export type RawPluginOptions = {
  key: string;
  value: string;
}[];
```

WDYT?

timostamm (commented):
Very nice, just left some comments regarding documenting the feature.

packages/protoplugin/src/create-es-plugin.ts
srikrsna-buf: Added the type, I think we can defer the standardOptions parameter.

timostamm (approved):
LGTM!

PR #882, opened 2024-06-06
V2: Throw errors in ReflectMessage instead of returning them

timostamm:
The reflection API provided by `ReflectMessage`, `ReflectList`, and `ReflectMap` validates a value before setting it. The methods return an error for an invalid value. For example: 

```ts
import { UserDesc } from "./gen/ts/extra/example_pb.js";
import { create } from "@bufbuild/protobuf";
import { reflect } from "@bufbuild/protobuf/reflect";

const user = create(UserDesc);
const msg = reflect(UserDesc, user);

const err = msg.set(UserDesc.field.firstName, 123);
console.log(err.message); // expected string, got 123
```

The `firstName` field is a string, and calling `msg.set` returns an error.

While this pattern works well, returning errors is very uncommon. This PR changes the behavior to throw errors instead. To handle the error, the example above changes to:

```ts
const user = create(UserDesc);
const msg = reflect(UserDesc, user);

try {
  msg.set(UserDesc.field.firstName, 123);
} catch (e) {
  if (isFieldError(e)) {
    console.log(e.message); // expected string, got 123
  }
}
```

### DIFF
--- a/packages/protobuf/src/reflect/reflect.ts
+++ b/packages/protobuf/src/reflect/reflect.ts
@@ -189,12 +189,12 @@ class ReflectMessageImpl<Desc extends DescMessage> implements ReflectMessage {
   set<Field extends DescField>(
     field: Field,
     value: ReflectSetValue<Field>,
-  ): FieldError | undefined {
+  ): void {
     assertOwn(this.message, field);
     if (this.check) {
       const err = checkField(field, value);
       if (err) {
-        return err;
+        throw err;
       }
     }
     let local: unknown;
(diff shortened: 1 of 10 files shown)


packages/protobuf/src/reflect/reflect.ts
```diff
@@ -189,12 +189,12 @@ class ReflectMessageImpl<Desc extends DescMessage> implements ReflectMessage {
...
     field: Field,
     value: ReflectSetValue<Field>,
-  ): FieldError | undefined {
+  ): void {
     assertOwn(this.message, field);
     if (this.check) {
       const err = checkField(field, value);
       if (err) {
-        return err;
+        throw err;
```
timostamm, line 197: We could throw in `checkField`, but the function is internal, so we can punt on it until there is a really good reason do change the function.

srikrsna-buf (approved):

PR #885, opened 2024-06-12
Backport binary encoding from "wire" subpath

timostamm:
In the upcoming [version 2](https://github.com/bufbuild/protobuf-es/releases/tag/v2.0.0-alpha.1), we relaxed validation in BinaryWriter (see https://github.com/bufbuild/protobuf-es/pull/877), and provided a better way for users to bring their own UTF-8 encoder (see https://github.com/bufbuild/protobuf-es/commit/70805568da3058ba2454c614d864ad1266362074).

For better interoperability between v1 and v2, this PR backports the relevant parts of the "wire" subpath.

### DIFF
--- a/packages/protobuf/tsconfig.json
+++ b/packages/protobuf/tsconfig.json
@@ -1,8 +1,5 @@
 {
-  "files": [
-    "src/index.ts",
-    "src/private/options-map.ts" // not yet exported, added here to satisfy the linter
-  ],
+  "files": ["src/index.ts", "src/wire/index.ts"],
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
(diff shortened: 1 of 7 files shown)


packages/protobuf/tsconfig.json
```diff
@@ -1,8 +1,5 @@
 {
-  "files": [
-    "src/index.ts",
-    "src/private/options-map.ts" // not yet exported, added here to satisfy the linter
```
timostamm, line 4: Removed options-map.ts - v2 provides options via descriptors.

timostamm:
This isn't working out well. We can't make use of the code in v1, so it's dead weight.

PR #896, opened 2024-06-18
V2: Add example for a default registry with a custom toJson() function

timostamm:
If a project frequently uses `google.protobuf.Any` and calls `toJson()` in many places, developers have to take care to always provide a registry. 

Since `toJson` is just a function, it's trivial to add a custom `toJson` function to a project that supplies a default registry:

```ts
import { createRegistry, toJson as toJsonOriginal } from "@bufbuild/protobuf";

// Define your own default registry
const registry = createRegistry(
  // You can register as many types as you need. In this example, we only
  // register google.protobuf.Duration.
  DurationSchema,
);

// Define your own toJson function that uses your default registry
export const toJson: typeof toJsonOriginal = (schema, message, options) => {
  return toJsonOriginal(schema, message, options ?? { registry }) as any;
};
```

Until we add a default registry as a feature (which is not without caveats), this seems like a acceptable workaround for users who need it. We should have this in our documentation for v2.

### DIFF
--- a/packages/protobuf-test/src/to-json-default-registry.test.ts
+++ b/packages/protobuf-test/src/to-json-default-registry.test.ts
@@ -1,0 +1,55 @@
+// Copyright 2021-2024 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


srikrsna-buf (approved):
The `toJson` function can also get the root message from the registry, but probably not a good idea.

PR #904, opened 2024-06-26
V2: Avoid Array.flatMap

timostamm:
Array.flatMap is part of ES2019, but we target ES2017. This replaces the call with an alternative.

This slipped through because `tsc` automatically picks up `@types/node`, which includes a declaration for `flatMap`. We're setting the compiler option `"types": []` to prevent this, and add "node" where necessary.

### DIFF
--- a/packages/protobuf/src/registry.ts
+++ b/packages/protobuf/src/registry.ts
@@ -248,7 +248,7 @@ export function createFileRegistry(
           deps.push(dep);
         }
       }
-      return [...deps, ...deps.flatMap((dep) => recurseDeps(dep))];
+      return deps.concat(...deps.map(recurseDeps));
     }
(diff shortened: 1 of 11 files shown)


packages/protobuf/src/registry.ts
```diff
@@ -248,7 +248,7 @@ export function createFileRegistry(
           deps.push(dep);
         }
       }
-      return [...deps, ...deps.flatMap((dep) => recurseDeps(dep))];
+      return deps.concat(...deps.map(recurseDeps));
```
timostamm, line 251: Looks more concise to me :shrug:

srikrsna (approved):

srikrsna-buf (approved):

PR #910, opened 2024-06-28
V2: Add package.json keywords

timostamm:


packages/protoplugin/package.json
```diff
@@ -1,8 +1,15 @@
 {
   "name": "@bufbuild/protoplugin",
   "version": "2.0.0-beta.2",
-  "license": "(Apache-2.0 AND BSD-3-Clause)",
+  "license": "Apache-2.0",
```
timostamm, line 4: As the README says, packages/protobuf/src/google/varint.ts is BSD-3-Clause, so `@bufbuild/protobuf` needs both licenses, but `@bufbuild/protoplugin` does not.

srikrsna-buf (approved):

PR #923, opened 2024-07-03
V2: Add $typeName first

timostamm:
In `create()`, add the `$typeName` property first. When logging objects with `console`, `$typeName` currently comes last. With this change, it comes first:

```
{
  '$typeName': 'example.User',
  firstName: 'Homer',
  lastName: '',
  active: true,
  locations: [],
  projects: {},
}
```

### DIFF
--- a/packages/bundle-size/package.json
+++ b/packages/bundle-size/package.json
@@ -2,7 +2,7 @@
   "name": "@bufbuild/bundle-size",
   "private": true,
   "scripts": {
-    "report": "tsx src/report.ts",
+    "bundle-size": "tsx src/report.ts",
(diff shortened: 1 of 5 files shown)


packages/bundle-size/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@bufbuild/bundle-size",
   "private": true,
   "scripts": {
-    "report": "tsx src/report.ts",
+    "bundle-size": "tsx src/report.ts",
```
timostamm, line 5: This was silently broken.

srikrsna-buf (approved):

PR #924, opened 2024-07-04
V2: Remove unused npm scripts

timostamm:


packages/protobuf/package.json
```diff
@@ -20,9 +20,7 @@
     "build:cjs": "../../node_modules/typescript/bin/tsc --project tsconfig.json --module commonjs --verbatimModuleSyntax false --moduleResolution node10 --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm": "../../node_modules/typescript/bin/tsc --project tsconfig.json --outDir ./dist/esm",
     "bootstrap:inject": "node scripts/bootstrap-inject.mjs src",
-    "prebootstrap:wkt": "rm -rf .tmp && mkdir -p .tmp && cp -rp src/wkt/gen/* .tmp",
     "bootstrap:wkt": "protoc --es_out=src/wkt/gen --es_opt=bootstrap_wkt=true,target=ts,import_extension=js,json_types=true --proto_path $(upstream-include wkt) $(upstream-files wkt) && license-header src/wkt/gen",
-    "postbootstrap:wkt": "diff >/dev/null -r .tmp src/wkt/gen && cp -rp .tmp/* src/wkt/gen || true",
```
timostamm, line 25: With `make`, we had to generate into a temp directory, and sync changes to src, preserving timestamps. Otherwise, `make` would consider the files changed, and be stuck in a loop.

Turbo uses hashes instead of file modification timestamps, so we don't need the workaround anymore.

srikrsna-buf (approved):

PR #926, opened 2024-07-05
V2: Documentation

timostamm:

PR #930, opened 2024-07-10
Copyedit docs and migrate content from v1

rakuista:


MANUAL.md
```diff
@@ -46,34 +46,37 @@ Protobuf-ES consists of three npm packages:
...
-# Learn more: https://buf.build/docs/configuration/v2/buf-gen-yaml
-version: v2
-inputs:
-  - directory: proto
-plugins:
-  - local: protoc-gen-es
-    out: src/gen
-    opt: target=ts
-```
+2.  In a new folder, create a `buf.gen.yaml` file that looks like this:
```
timostamm, line 58: This file shouldn't be created in a new folder. The `npm install` commands imply that we're already in a new blank project folder, and `buf.gen.yaml` should sit at the root of that project folder.

The complete implied setup would be something like:

```shellsession
mkdir example
cd example
npm init -y
npm install typescript
npx tsc --init
```

Then the install commands.

Should we include this step for users who aren't familiar with the ecosystem? I tend to think so.

Tangential: For shell examples, do we want to include prompts? Then it would be:

```shellsession
$ mkdir example
$ cd example
$ npm init -y
$ npm install typescript
$ npx tsc --init
```

The advantage of leaving _out_ prompts is that the snippet can be copied, pasted, and run. It'll run all commands sequentially.

MANUAL.md
```diff
@@ -128,10 +131,60 @@ plugins:
       - import_extension=js
 ```
 
-With [`protoc`][gh-protoc], multiple options are specified with multiple `--es_opt` flags. Alternatively, both compilers
-allow to specify multiple options as a single comma-separated value like `target=ts,import_extension=js`.
+With [`protoc`][gh-protoc], you specify multiple options with multiple `--es_opt` flags. Alternatively, both compilers
+allow you to specify multiple options as a single comma-separated value like `target=ts,import_extension=js`.
+
+## Plugin options
```
timostamm, line 137:
```suggestion
```

MANUAL.md
rakuista: I removed the prompts—on buf.build, we have some code that causes them not to be copied but GitHub apprarently doesn't. I agree on adding the steps too, so put those in.

MANUAL.md
```diff
@@ -1426,15 +1507,15 @@ The companion function `hasOption` returns true if an option is present. The fun
...
+> Custom options can be read from generated code, or from the schema passed to a plugin. See the example in
 > our [guide for writing plugins](#writing-plugins).
-
-> [!TIP]
 >
 > To learn more about custom options in Protobuf, see the [language guide][protobuf.dev/customoptions].
 
 ### Reflection API
 
+TODO: Can the API be used without the BSR? We should note requirements and link to the BSR docs if not.
```
timostamm, line 1517: This is independent of the BSR. 

With the features listed here, it's possible to take full advantage of the BSR reflection API. It would be possible to write an SDK like [prototransform for Go](https://github.com/bufbuild/prototransform), or a command like `buf curl`. But doing that here in an example is a heavy lift.

I suggest we add an introductory paragraph right at the top `## Reflection` section:

> Reflection allows to inspect a schema and to dynamically manipulate data. In the following sections, we'll explain the core primitives, and write a function to redact sensitive information from messages.

Additionally, we can link to the BSR API, for example at the bottom of the `### Descriptors` section:
 
> [!TIP]
> 
> You can also fetch descriptors [from the Buf Schema Registry](https://buf.build/docs/bsr/reflection/overview).

MANUAL.md
```diff
@@ -128,80 +136,69 @@ plugins:
...
 
-/**
- * @generated from message example.User
- */
-export declare type User = Message<"example.User"> & {
-  /**
-   * @generated from field: string first_name = 1;
-   */
-  firstName: string;
+By default, [protoc-gen-es] doesn't add file extensions to import paths. However, some environments require an import
```
timostamm, line 158: We don't have a link reference for `[protoc-gen-es]`. Should we change this to `[@bufbuild/protoc-gen-es]`?

It's the npm package name for the same thing.

MANUAL.md
```diff
@@ -128,80 +136,69 @@ plugins:
...
+
+[protoc-gen-es] generates valid TypeScript for current versions of the TypeScript compiler with standard settings.
+If you use compiler settings that yield an error for generated code, setting this option generates an annotation at
+the top of each file to skip type checks: `// @ts-nocheck`.
+
+### `json_types=true`
 
-> **TODO** does this overview provide value? we basically go through all parts of it (and more) in the following sections
+Generates JSON types for every Protobuf message and enumeration. Calling `toJson()` automatically returns the JSON type 
+if available.
```
timostamm, line 190: Should we add a link to the section explaining the feature?

```
Learn more about the feature [here](#json-types).
```

MANUAL.md
```diff
@@ -1600,158 +1692,748 @@ if (field.fieldKind == "map") {
...
+The contract between the Protobuf compiler and a code generator plugin is defined in [plugin.proto]. Plugins are simple
+executables (typically on your `$PATH`) named `protoc-gen-x`, where `x` is the name of the language or feature that the
+plugin provides. The Protobuf compiler parses the Protobuf files and invokes the plugin, sending a
+`CodeGeneratorRequest` on stdin, and expecting a `CodeGeneratorResponse` on stdout. The request contains a set of
+descriptors—an abstract version of the parsed Protobuf files (see [descriptor.proto]). The response contains a list of
+files, each with a name and text content.
 
-> **TODO** packages/protobuf-example contains a runnable example that uses the same example.User definition we've been using in this doc. We should link to it.
+For more information on how plugins work, check out [the Buf image documentation][buf-images].
+TODO: Is this still the best place to link to? It's pretty Buf-specific in this more general context.
```
timostamm, line 1707: The page has a section "How Protobuf plugins work" that fits, but I don't think it's worth still linking to it.

```suggestion
```

MANUAL.md
```diff
@@ -1082,40 +1086,42 @@ const b = clone(UserSchema, a);
...
 
-As a general guide to decide between the binary format and JSON: The JSON format is great for debugging, but the binary
-format is more resilient to changes. For example, you can rename a field, and still parse binary data serialized with
+As a general guide when deciding between the binary format and JSON, the JSON format is great for debugging, but the binary
+format is more resilient to changes. For example, you can rename a field and still parse binary data serialized with
 the previous version. In general, the binary format is also more performant than JSON.
 
 ### Binary serialization options
 
+TODO: Examples to show these in context?
```
timostamm, line 1095: The options change how binary data is serialized. To give an example, we have to show binary data. I'm afraid we would need to do an excursion to explain the binary format first to effectively show that. Or we would need a complex setup with multiple example files and many steps. It would probably be about one page of code per option.

MANUAL.md
```diff
@@ -1600,158 +1692,748 @@ if (field.fieldKind == "map") {
...
+
+Additionally, check out [protoc-gen-es](https://github.com/bufbuild/protobuf-es/tree/main/packages/protoc-gen-es), which is the official code generator for Protobuf-ES.
+
+## Examples
+
+> **TODO** packages/protobuf-example contains a runnable example that uses the same example.User definition we've been using in this doc. We should link to it.
+
+## Migrating to Protobuf-ES
+
+TODO: Need to update migration.md and move to this doc
```
timostamm, line 2165: migrating.md needs extensive updates and should also have a section about migrating from Protobuf-ES v1 to v2.

Can we remove this section and TODO for now and do that separately?

MANUAL.md
```diff
@@ -1600,158 +1692,748 @@ if (field.fieldKind == "map") {
...
+}
+```
+
+Custom options can be set on any Protobuf element. They can be simple singular
+string fields as the one above, but also repeated fields, message fields, etc.
+
+Take a look at our [plugin example](./packages/protoplugin-example/)
+to see the custom option above in action and run the code yourself.
+
+## Testing
```
timostamm, line 2147:
```suggestion
### Testing plugins
```

MANUAL.md
```diff
@@ -1600,158 +1692,748 @@ if (field.fieldKind == "map") {
...
+
+## Examples
+
+For a small example of generating a Twirp client based on a simple service definition, take a look at [protoplugin-example](https://github.com/bufbuild/protobuf-es/tree/main/packages/protoplugin-example).
+
+Additionally, check out [protoc-gen-es](https://github.com/bufbuild/protobuf-es/tree/main/packages/protoc-gen-es), which is the official code generator for Protobuf-ES.
+
+## Examples
+
+> **TODO** packages/protobuf-example contains a runnable example that uses the same example.User definition we've been using in this doc. We should link to it.
```
timostamm, line 2161: We merge these:

```suggestion
## Examples

For a runnable example that uses Protocol Buffers to manage a list of users, see
[packages/protobuf-example](packages/protobuf-example).

For a custom plugin, see [packages/protoplugin-example](packages/protoplugin-example). It generates Twirp clients for
your services, and also uses [custom options](#custom-options).
```

timostamm (commented):
Left a couple of comments and fixed links in https://github.com/bufbuild/protobuf-es/pull/930/commits/48c95c134cbc0f00c20762107ff3d8059966c94f and https://github.com/bufbuild/protobuf-es/pull/930/commits/6fd7e22d8c0c1d1132444676920e6afd100488f8.

I'd like to condense "Writing plugins" a bit. But I think we should merge this and it's parent https://github.com/bufbuild/protobuf-es/pull/926 first.

PR #935, opened 2024-07-19
V2: Add GeneratedFile.array

timostamm:
`GeneratedFile` already has the method `string()` to escape string literals. We already have an internal function for array literals. This promotes it to be a method on `GeneratedFile` as well.

```ts
import { GeneratedFile } from "@bufbuild/protoplugin";

declare f: GeneratedFile;

f.print("const arr = ", f.array([1, 2, 3]), ";");
// const arr = [1, 2, 3];
```

### DIFF
--- a/packages/protoplugin/src/safe-identifier.ts
+++ b/packages/protoplugin/src/safe-identifier.ts
@@ -83,9 +83,11 @@ const reservedIdentifiers = new Set([
 ]);
 
 /**
- * Escapes names that are reserved identifiers in ECMAScript, TypeScript.
+ * Escapes reserved words in ECMAScript and TypeScript identifiers, by appending
+ * a dollar sign.
  *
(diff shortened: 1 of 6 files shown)


packages/protoplugin/src/safe-identifier.ts
```diff
@@ -83,7 +83,11 @@ const reservedIdentifiers = new Set([
 ]);
 
 /**
- * Escapes names that are reserved identifiers in ECMAScript, TypeScript.
+ * Escapes reserved words in ECMAScript and TypeScript identifiers, by appending
+ * a dollar sign.
+ *
+ * This function is intended for use with identifiers from Protobuf. The passed
+ * string must be a valid identifier (e.g. not start with a digit).
```
timostamm, line 90: Not directly related, but I noticed the docs for the function `safeIdentifier` need clarification.

smaye81 (approved):

PR #958, opened 2024-08-12
Update buf CLI and make use of the `clean` option

timostamm:
[v1.36.0](https://github.com/bufbuild/buf/releases/tag/v1.36.0) of the buf CLI adds the top-level option `clean` to `buf.gen.yaml`. We can replace our npm scripts with this option.

### DIFF
--- a/packages/protobuf-example/package.json
+++ b/packages/protobuf-example/package.json
@@ -6,7 +6,6 @@
     "build": "../../node_modules/typescript/bin/tsc --noEmit",
     "add": "tsx src/add.ts",
     "list": "tsx src/list.ts",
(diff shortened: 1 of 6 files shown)


packages/protobuf-example/package.json
```diff
@@ -6,7 +6,6 @@
     "build": "../../node_modules/typescript/bin/tsc --noEmit",
     "add": "tsx src/add.ts",
     "list": "tsx src/list.ts",
-    "pregenerate": "rm -rf src/gen",
```
timostamm, line 9: One less shell process to run, and this npm script is not portable, so I'm very happy that we can remove it.

smaye81 (approved):

PR #964, opened 2024-08-23
(WIP) Improve Binary Encode Performance

Ekrekr:
Work in progress - I haven't yet come up with a way to efficiently fork a new message directly within the contigious memory space, for length delimited records.

Efficiency optimisations through avoiding copying of array buffers, by writing directly to an expanding array buffer with dynamic contiguous memory space.

TODO: add performance tests and flamegraphs once working.

### DIFF
--- a/packages/protobuf/src/to-binary.ts
+++ b/packages/protobuf/src/to-binary.ts
@@ -125,11 +125,8 @@ function writeScalar(
   fieldNo: number,
   value: unknown
 ) {
-  writeScalarValue(
-    writer.tag(fieldNo, writeTypeOfScalar(scalarType)),
-    scalarType,
-    value as ScalarValue
-  );
+  writer.tag(fieldNo, writeTypeOfScalar(scalarType));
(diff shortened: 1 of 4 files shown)


packages/protobuf/src/to-binary.ts
```diff
@@ -80,7 +80,7 @@ function writeFields(
   }
   if (opts.writeUnknownFields) {
     for (const { no, wireType, data } of msg.getUnknown() ?? []) {
-      writer.tag(no, wireType).raw(data);
+      writer.tag(no, wireType).bytes(data);
```
timostamm, line 83: The `bytes` method prefixes the length. The change will corrupt data, and needs to be reverted.

timostamm (commented):
Thanks for the PR, Elias!

Unfortunately, resizable ArrayBuffers are relatively new. We'll have to wait until they are more widely available before we can use them.

I left a couple of informative comments.

ajeetdsouza:
Closing in favor of https://github.com/bufbuild/protobuf-es/pull/1108.

PR #985, opened 2024-10-04
Add typed `DescMethod` variants

srikrsna-buf:
Add typed `DescMethod` variants.

### DIFF
--- a/packages/protobuf/src/types.ts
+++ b/packages/protobuf/src/types.ts
@@ -13,7 +13,12 @@
 // limitations under the License.
 
 import type { GenEnum, GenExtension, GenMessage } from "./codegenv1/types.js";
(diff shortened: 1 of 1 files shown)


packages/protobuf/src/types.ts
```diff
@@ -89,6 +94,65 @@ export type UnknownField = {
...
+/**
+ * Describes a unary RPC declaration.
+ */
+export type DescMethodUnary<
+  I extends DescMessage = DescMessage,
+  O extends DescMessage = DescMessage,
+> = DescMethodCommon & {
+  methodKind: "unary";
+  input: I;
+  output: O;
```
timostamm, line 117: The corresponding `DescMethod` properties are readonly. I think it would be good to make the replacements readonly as well.

PR #1001, opened 2024-10-29
Solution for the issue of `snake_case` fields being lost after calling the `create` method

locene:
### Description

Given the following data:

```
const data = {
    "sample_content": "some things"
}
```

And the corresponding `proto` definition:

```
message Data {
    string sample_content = 1;
}
```

When using this setup in the `create` method:

```
create(DataSchema, {
    data: data
})
```

The `"sample_content"` field will be ignored and not converted correctly.

### Analysis

In the `initMessage` method of `create.ts`, the value is accessed using `member.localName`:

```
let value = (init as Record<string, unknown>)[member.localName]
```

At this point, the value of `member.localName` has already been converted to `camelCase`, resulting in `"sampleContent"`, which does not match the original `snake_case` format of `"sample_content"`, causing this field to be ignored.

### Solution

To address this issue, modify the code as follows:

```
let value = (init as Record<string, unknown>)[member.localName] ?? (init as Record<string, unknown>)[member.name] ?? null;
```

This change will resolve the issue without impacting the existing functionality.

### DIFF
--- a/packages/protobuf/src/create.ts
+++ b/packages/protobuf/src/create.ts
@@ -66,7 +66,7 @@ function initMessage<Desc extends DescMessage>(
   init: MessageInitShape<Desc>,
 ): MessageShape<Desc> | FieldError {
   for (const member of messageDesc.members) {
-    let value = (init as Record<string, unknown>)[member.localName];
+    let value = (init as Record<string, unknown>)[member.localName] ?? (init as Record<string, unknown>)[member.name] ?? null;
     if (value == null) {
       // intentionally ignore undefined and null
       continue;


timostamm:
From the [manual](https://github.com/bufbuild/protobuf-es/blob/v2.2.1/MANUAL.md#field-names):

> Property names are always `lowerCamelCase`, even if the corresponding Protobuf field uses `snake_case`. Though there's no official style for ECMAScript, most style guides ([AirBnB][js-style-guide-airbnb], [MDN][js-style-guide-mdn], [Google][js-style-guide-google]) as well as [Node.js APIs][js-style-node] and [browser APIs][js-style-fetch] use `lowerCamelCase`, and so do we.

We want to keep Protobuf-ES simple and focused, and while this change may not impact existing functionality, it does not help with these goals.

The better solution would be to write a function `createFrom` for your project, which accepts properties with the original Protobuf field name. The reflection API should make this feasible.


[js-style-fetch]: https://fetch.spec.whatwg.org/#request-class
[js-style-guide-airbnb]: https://github.com/airbnb/javascript#naming--camelCase
[js-style-guide-google]: https://google.github.io/styleguide/jsguide.html#naming-non-constant-field-names
[js-style-guide-mdn]: https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Code_guidelines/JavaScript#variable_naming
[js-style-node]: https://nodejs.org/dist/latest-v16.x/docs/api/child_process.html#child_processforkmodulepath-args-options

locene:
OK, no problem. Could you please revoke my CLA on cla-assistant.io? I signed a bit hastily, and if I had known you weren't planning to merge it, I wouldn't have done so. Thanks for your understanding.

locene:
OK, no problem.

Could you please revoke my CLA on cla-assistant.io? I signed a bit hastily, and there are some issues regarding my ability to sign the CLA on behalf of my employer. Had I known you weren't planning to merge it, I wouldn't have done so.

Thank you for your understanding.

timostamm:
I cannot revoke your CLA. But you can, on https://cla-assistant.io/my-cla

locene:
No, I can't. The list is empty when I open it, which is why I'm reaching out to you. My understanding is that it requires a maintainer to delete it.

cla-assistant/cla-assistant#907
cla-assistant/cla-assistant#380
cla-assistant/cla-assistant#406

timostamm:
Have you tried signing in first on https://cla-assistant.io/ ?

When I do that, I see CLAs I've signed on https://cla-assistant.io/my-cla, and I have a button to revoke:

Screenshot 2024-10-30 at 15 17 48

timostamm:
We have looked into it, but we don't see a way to revoke it.

For some context, I don't believe that the state of this PR (or any past ones) will change by revoking a CLA, only future ones. If it would be possible to retroactively revoke a CLA for a contribution, OSS could not safely accept any contributions. But you haven't contributed, and as far as I understand, you have revoked your CLA for future contributions. IANAL, but it looks to me like this is void.

PR #1003, opened 2024-10-30
Print diagnostics when a protoplugin fails to emit files

will-wow:
When trying to publish a custom plugin to the BSR, I had some trouble figuring out why it was failing to generate an SDK.

I ended up debugging the problem by implementing a custom transpiler that prints extra diagnostic information when results are skipped, since these errors show up in the npm install log when it fails to generate.

So, this seems like it could be useful for others trying to debug SDK generation failures in the future.

This generates an error like:

```
protoc-gen-es-required: A problem occurred during transpilation and files were not generated.

Generating Files:

buf/validate/validate_required.ts
factory_execution/cycle_time/v1/cycle_time_required.ts

Diagnostics:

Property 'in' of exported interface has or is using private name 'Uint8Array'.
Property 'notIn' of exported interface has or is using private name 'Uint8Array'.
Property 'example' of exported interface has or is using private name 'Uint8Array'.
```

### DIFF
--- a/packages/protoplugin-test/src/transpile.test.ts
+++ b/packages/protoplugin-test/src/transpile.test.ts
@@ -97,4 +97,66 @@ describe("built-in transpile", () => {
       ]);
     });
   });
+
+  describe("failing to emit", () => {
+    test("raises error with helpful message", async () => {
+      await expect(async () =>
+        testTranspileToDts([
+          `export interface Foo {`,
+          `  p: {`,
+          `    [K in keyof P]: string;`,
+          `  },`,
+          `}`,
+        ]),
+      ).rejects.toThrow(
+        /^A problem occurred during transpilation and files were not generated\. {2}Contact the plugin author for support\.\n/,
+      );
+    });
+    test("raises error with diagnostics", async () => {
+      await expect(async () =>
+        testTranspileToDts([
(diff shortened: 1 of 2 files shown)


timostamm (approved):
In general, I recommend the following approach if transpilation fails:
- Generate the same input with the plugin option `target=ts` to local files
- Compile the local files with TypeScript v5.4.5 (the version that transpilation uses)
- Inspect errors in an IDE

This will give you all error details, and let's you iterate quickly. Of course it's still helpful to print diagnostics, so thanks for the PR!

I've pushed up https://github.com/bufbuild/protobuf-es/pull/1003/commits/7e1786052e3c40abca431d23ca7f0b498d7c78bf to use the TypeScript compiler to format diagnostics (it handles locations and error chains):

```
A problem occurred during transpilation and files were not generated.  Contact the plugin author for support.

test.ts(3,17): error TS4033: Property 'p' of exported interface has or is using private name 'P'.
test.ts(8,17): error TS4033: Property 'p' of exported interface has or is using private name 'P'.
test.ts(13,17): error TS4033: Property 'p' of exported interface has or is using private name 'P'.
2 more diagnostics elided
```

It also prints at most three diagnostics because the error message can potentially grow to megabytes of text.

PR #1004, opened 2024-10-31
Fix JSON types for fields with uncommon json_name

nicole0707:
### Context

This PR enhances the existing logic for handling JSON property names by addressing edge cases to ensure compatibility with JavaScript naming conventions. The update focuses on identifying and properly handling property names that do not conform to valid JavaScript identifiers, such as names starting with digits, containing special characters, or including non-ASCII characters.

### Changes
* Enhanced Edge Case Handling
  * Updated logic to wrap JSON property names in quotes when they:
    * Start with a digit.
    * Contain special characters not allowed in JavaScript identifiers.
    * Include non-ASCII characters.

* Additional Test Cases
  * Added test cases to verify that the updated logic correctly handles:
    * Property names with special characters (e.g., foo@, foo-bar).
    * Property names starting with digits (e.g., 123foo).
    * Non-ASCII characters in property names (e.g., 你好).
    * Property names containing spaces or escape sequences (e.g., foo bar, foo\nbar).


packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
```diff
@@ -477,9 +477,13 @@ function generateMessageJsonShape(f: GeneratedFile, message: DescMessage, target
...
               let jsonName: Printable = field.jsonName;
+              const startWithNumber = /^[0-9]/;
+              const containsSpecialChar = /[^a-zA-Z0-9_$]/;
+              const isAscii = /^[\x00-\x7F]*$/;
               if (jsonName === ""
-                || /^[0-9]/.test(jsonName)
-                || jsonName.indexOf("@") > -1) {
+                || startWithNumber.test(jsonName)
+                || containsSpecialChar.test(jsonName)
+                || isAscii.test(jsonName)) {
```
timostamm: Because the first regex ensures a subset of ASCII, we don't need the ASCII check. Removed it in https://github.com/bufbuild/protobuf-es/pull/1004/commits/f095a7d42140570cab7e3c749ffa3b29c3946bd1

timostamm (approved):
Thanks for the fix!

PR #1013, opened 2024-11-05
Update to TypeScript 5.6.3

timostamm:


packages/protobuf/src/reflect/reflect.ts
```diff
@@ -370,12 +370,12 @@ class ReflectMapImpl<K, V> implements ReflectMap<K, V> {
   has(key: K) {
     return Object.prototype.hasOwnProperty.call(this.obj, mapKeyToLocal(key));
   }
-  *keys() {
+  *keys(): MapIterator<K> {
```
timostamm, line 373: The inferred generator type no longer matches the built-in `ReadonlyMap`. We have to add an explicit return type. Same as a couple of other places. This is unlikely to cause issues, and should be covered by the typescript-compat tests.

srikrsna-buf (approved):

PR #1022, opened 2024-11-28
Doc: clarify which packages are dev dependencies

davidfiala:
I've proposed splitting the runtime vs dev dependencies into separate lines as aligned with best practices. Additionally, I've added a default-collapsed section for yarn users.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -58,11 +58,11 @@ The quickstart below shows a simple example of code generation for a `.proto` fi
     npx tsc --init
     ```
 
-2.  Install the compiler [@bufbuild/buf], plugin, and runtime library:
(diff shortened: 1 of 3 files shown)


timostamm (commented):
Thanks for the PR, David!

We want to keep the quickstart short, but one extra line to show which dependencies are built-time only is worth it 👍 

I love the use of collapsed sections, but I'm afraid we'd rather keep it minimal, and only use `npm`. It's not that we don't like Yarn, but there's also `pnpm`, `bun`, and `jsr`.

timostamm (approved):
Pushed up https://github.com/bufbuild/protobuf-es/pull/1022/commits/0e7ae1b5d4e392a25aa049e74ac5a06cdece907d to make the same change in two other markdown files.

LGTM!

PR #1029, opened 2024-12-02
Add experimental options to equals() to include extensions and others

timostamm:
The function `equals` from `@bufbuild/protobuf` compares two messages of the same type. 

This PR adds an options argument to enable comparison of extensions, unknown fields, and semantic comparison of `google.protobuf.Any`:

```ts
interface EqualsOptions {
  /**
   * A registry to look up extensions, and messages packed in Any.
   *
   * @private Experimental API, does not follow semantic versioning.
   */
  registry: Registry;
  /**
   * Unpack google.protobuf.Any before comparing.
   * If a type is not in the registry, comparison falls back to comparing the
   * fields of Any.
   *
   * @private Experimental API, does not follow semantic versioning.
   */
  unpackAny?: boolean;
  /**
   * Consider extensions when comparing.
   *
   * @private Experimental API, does not follow semantic versioning.
   */
  extensions?: boolean;
  /**
   * Consider unknown fields when comparing.
   * The registry is used to distinguish between extensions, and unknown fields
   * caused by schema changes.
   *
   * @private Experimental API, does not follow semantic versioning.
   */
  unknown?: boolean;
}
```

For now, the feature is an experimental API, and it's marked `@private`. This means it's possible to use the API, but at your own risk: It is possible that we change the API, or remove it again.

### DIFF
--- a/packages/protobuf/src/equals.ts
+++ b/packages/protobuf/src/equals.ts
@@ -255,7 +255,7 @@ function extensionsEquals(
   ): DescExtension[] {
     return (msg.getUnknown() ?? [])
       .map((uf) => registry.getExtensionFor(msg.desc, uf.no))
-      .filter((e) => !!e)
+      .filter((e) => e != undefined)
       .filter((e, index, arr) => arr.indexOf(e) === index);
   }
   const extensionsA = getSetExtensions(a, opts.registry);
--- a/packages/protobuf/src/equals.ts
+++ b/packages/protobuf/src/equals.ts
@@ -15,8 +15,47 @@
 import type { MessageShape } from "./types.js";
 import { scalarEquals, type ScalarValue } from "./reflect/scalar.js";
 import { reflect } from "./reflect/reflect.js";
-import type { DescField, DescMessage } from "./descriptors.js";
+import {
+  type DescExtension,
+  type DescField,
+  type DescMessage,
+  ScalarType,
+} from "./descriptors.js";
 import type { ReflectMessage } from "./reflect/index.js";
+import type { Registry } from "./registry.js";
+import { type Any, anyUnpack } from "./wkt/index.js";
+import { createExtensionContainer, getExtension } from "./extensions.js";
+
+export interface EqualsOptions {
+  /**
+   * A registry to look up extensions, and messages packed in Any.
+   *
+   * @private Experimental API, does not follow semantic versioning.
+   */
+  registry: Registry;
+  /**
+   * Unpack google.protobuf.Any before comparing.
+   * If a type is not in the registry, comparison falls back to comparing the
+   * fields of Any.
+   *
+   * @private Experimental API, does not follow semantic versioning.
+   */
+  unpackAny?: boolean;
+  /**
+   * Consider extensions when comparing.
+   *
+   * @private Experimental API, does not follow semantic versioning.
+   */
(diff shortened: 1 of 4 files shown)


packages/protobuf/src/equals.ts
```diff
@@ -135,3 +193,91 @@ function fieldEquals(
...
+  opts: EqualsOptions,
+) {
+  function getSetExtensions(
+    msg: ReflectMessage,
+    registry: Registry,
+  ): DescExtension[] {
+    return (msg.getUnknown() ?? [])
+      .map((uf) => registry.getExtensionFor(msg.desc, uf.no))
+      .filter((e) => !!e)
+      .filter((e, index, arr) => arr.indexOf(e) === index);
```
jhump, line 259: This expression seems like a tautology. What is this trying to filter out? Is this just removing duplicates (where `arr.indexOf(e)` might return an earlier index)?

packages/protobuf/src/equals.ts
```diff
@@ -135,3 +193,91 @@ function fieldEquals(
...
+  b: ReflectMessage,
+  opts: EqualsOptions,
+) {
+  function getSetExtensions(
+    msg: ReflectMessage,
+    registry: Registry,
+  ): DescExtension[] {
+    return (msg.getUnknown() ?? [])
+      .map((uf) => registry.getExtensionFor(msg.desc, uf.no))
+      .filter((e) => !!e)
```
jhump, line 258: Is this just removing unknown values? If so, a little surprising that this is idiomatic instead of something more intuitive/readable 🤷.

packages/protobuf/src/equals.ts
```diff
@@ -64,33 +118,35 @@ function fieldEquals(
...
-        keysA.push(k);
+        keys.push(k);
       }
-      for (const k of mb.keys()) {
-        if (!ma.has(k)) {
+      for (const k of mapB.keys()) {
+        if (!mapA.has(k)) {
           return false;
         }
       }
```
jhump, line 136: You have a more efficient/concise way to test this below, for extensions: instead of scanning through both sets of keys, check that they both have the same _number_ of keys and then just scan `a`, verifying that each key is also present in `b`.

jhump (approved):
Minor remarks/questions. LGTM.

packages/protobuf/src/equals.ts
timostamm: Yes, it's removing duplicates. This expression is widely used, for better or worse.

packages/protobuf/src/equals.ts
timostamm: Yes, it's removing unknown values. We can make this a bit more readable - with https://github.com/bufbuild/protobuf-es/pull/1029/commits/410c9393a220f7be6e1dd6c9194fb1d486f61f1d, it's `e != undefined` now.

It's actually a new feature that the TypeScript compiler can narrow down the expression type from `filter` calls. I was confused that it worked for `!!e`, but didn't appear to work for `e != undefined`, but it actually does.

packages/protobuf/src/equals.ts
timostamm: This is a facade to access an object like a map type. I think the logic is trying to avoid some expensive operations. I've added a comment.

PR #1039, opened 2024-12-18
Update to google protobuf v29.2

timostamm:


packages/protoplugin-test/src/plugin-edition-support.test.ts
```diff
@@ -70,7 +70,7 @@ describe("editions support in plugins", () => {
       generateAny() {},
     });
     expect(res.minimumEdition).toBe(minimumEdition);
-    expect(res.minimumEdition).toBe(minimumEdition);
+    expect(res.maximumEdition).toBe(maximumEdition);
```
timostamm, line 73: This test is supposed to verify min and max edition, but it was testing min twice. This fixes the test.

smaye81 (approved):

Issue #414, opened 2023-03-02
proto2 `required` support


zeroslope:
The required syntax of proto2 is now optional in the generated type, is there any way to make it required?

``` protobuf
syntax = "proto2";

message Foo {
  required string a = 1;
  required string b = 2;
}
```

``` ts
// @generated by protoc-gen-es v1.0.0 with parameter "target=ts"
// @generated from file foo.proto (syntax proto2)
/* eslint-disable */
// @ts-nocheck

import type { BinaryReadOptions, FieldList, JsonReadOptions, JsonValue, PartialMessage, PlainMessage } from "@bufbuild/protobuf";
import { Message, proto2 } from "@bufbuild/protobuf";

/**
 * @generated from message Foo
 */
export class Foo extends Message<Foo> {
  /**
   * @generated from field: required string a = 1;
   */
  a?: string;

  /**
   * @generated from field: required string b = 2;
   */
  b?: string;

  constructor(data?: PartialMessage<Foo>) {
    super();
    proto2.util.initPartial(data, this);
  }

  static readonly runtime = proto2;
  static readonly typeName = "Foo";
  static readonly fields: FieldList = proto2.util.newFieldList(() => [
    { no: 1, name: "a", kind: "scalar", T: 9 /* ScalarType.STRING */ },
    { no: 2, name: "b", kind: "scalar", T: 9 /* ScalarType.STRING */ },
  ]);

  static fromBinary(bytes: Uint8Array, options?: Partial<BinaryReadOptions>): Foo {
    return new Foo().fromBinary(bytes, options);
  }

  static fromJson(jsonValue: JsonValue, options?: Partial<JsonReadOptions>): Foo {
    return new Foo().fromJson(jsonValue, options);
  }

  static fromJsonString(jsonString: string, options?: Partial<JsonReadOptions>): Foo {
    return new Foo().fromJsonString(jsonString, options);
  }

  static equals(a: Foo | PlainMessage<Foo> | undefined, b: Foo | PlainMessage<Foo> | undefined): boolean {
    return proto2.util.equals(Foo, a, b);
  }
}


```

timostamm:
We agree that this would be nice for type-safety, but it means we have to make properties mandatory in the constructor, something we would prefer to avoid. We haven't found the best way forward yet for this conundrum, so for the time being, proto2 required properties are typed as optional, but will raise an error at runtime when serializing, following the pattern of other implementations.

[...]

PR #1043, opened 2024-12-30
Add strict option to v1

lukasIO:
addresses https://github.com/bufbuild/protobuf-es/issues/414

This PR adds a `strict` option to protoc-gen-es v1 to allow for `required` labels in proto2 syntax to be reflected as actually required fields. 

All current tests pass with `strict=0` (which is of course the default), so there should be no breaking changes. 

In order to facilitate type checking for strict mode there's a new `PartialStrictMessage` type that enforces required fields to be part of the initial message constructor.

Would highly appreciate if you considered adding this option. I tried to make it non breaking and as useful as possible for others.

### DIFF
--- a/packages/protobuf/src/codegen-info.ts
+++ b/packages/protobuf/src/codegen-info.ts
@@ -47,11 +47,11 @@ interface CodegenInfo {
       | DescOneof
       | DescField
       | DescService
-      | DescMethod
+      | DescMethod,
   ) => string;
   readonly symbols: Record<RuntimeSymbolName, RuntimeSymbolInfo>;
   readonly getUnwrappedFieldType: (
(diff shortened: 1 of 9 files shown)


nicksnyder:
Going to close this PR because we have no plans to merge it for reasons discussed in the linked issue. In short:
* Using `required` in Protobuf files is not recommend and a legacy option.
* The correct place to enforce a required semantic is in code. Our recommended solution to this is going to be to use protovalidate. We're actively working on a typescript implementation here: https://github.com/bufbuild/protovalidate/issues/67
* Once we have that implemented, we can start generating `*Valid` types as mentioned here: https://github.com/bufbuild/protobuf-es/issues/414#issuecomment-2416472463

PR #1060, opened 2025-01-25
Name the plain type which actually matches the protobuf definition

### LINKED ISSUE #1058: Create raw data types as well
```protobuf
syntax = "proto3";
package example;
message User {
  string id = 1;
}
```

will produce something like

```ts
export type User = Message<"example.User"> & {
  /**
   * @generated from field: string id = 1;
   */
  id: string;
}
```

However, the `User` type is hard to use when working with the message body as data throughout the code based because of the $typeName and $unknown fileds that `Message<"example.User">` adds. What if instead it generated something like:

```ts
export type UserRaw = {
  /**
   * @generated from field: string id = 1;
   */
  id: string;
};

export type User = Message<"example.User"> & UserRaw
```

So that `UserRaw` can be used to hold data in the program logic, but one can also easily do:

```ts
const user: UserRaw = { id: "123" };
create(UserSchema, user)
```

This would remove the need for using a helper like `OmitTypeName` all over the place.

abalmos:
The type was already generated, but not exported (except as a union export). Having access to this plain type is almost essential for application code that utilizes the generated message types for purposes beyond this library's own API.

fixes #1058

### DIFF
--- a/packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
+++ b/packages/protoc-gen-es/src/protoc-gen-es-plugin.ts
@@ -399,7 +399,7 @@ function generateMessageShape(f: GeneratedFile, message: DescMessage, target: Ex
   const { Message } = f.runtime;
   const declaration = target == "ts" ? "type" : "declare type";
   f.print(f.jsDoc(message));
-  f.print(f.export(declaration, f.importShape(message).name), " = ", Message, "<", f.string(message.typeName), "> & {");
+  f.print(f.export(declaration, `${f.importShape(message).name}Plain`), " = {");
   for (const member of message.members) {
     switch (member.kind) {
       case "oneof":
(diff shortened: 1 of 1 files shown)


anthonysessa:
I am aiming to get something very similar here: https://github.com/bufbuild/protobuf-es/issues/1062.  My suggestion is to add an option to make the `json_types` respect optional vs. non-optional fields like the `main` proto types do.  

OR

Should `json_types` respect optional vs non-optional proto definitions out of the box?

Either way same end result.

anthonysessa:
I think the issue here is that `oneOf` messages create a nested structure that utilizes a `case`: `value` approach.  So the end result in this case is not representative of the basic proto typings.

I am aiming to get something very similar here: https://github.com/bufbuild/protobuf-es/issues/1062.  My suggestion is to add an option to make the `json_types` respect optional vs. non-optional fields like the `main` proto types do.  

OR

Should `json_types` respect optional vs non-optional proto definitions out of the box?

timostamm:
As explained in https://github.com/bufbuild/protobuf-es/issues/1058#issuecomment-2772497493, exporting this type is simple, but using it correctly is not. Without test coverage for edge cases with field presence and documentation explaining the tradeoffs, we're setting users up to run into strange bugs.

Thank you for the initiative, @abalmos, I hope we can discuss more alternatives in the issue.

PR #1061, opened 2025-01-27
Allow timestamp with both fractions and offset

liskaant:
Regex parsing for JSON timestamps does not allow both fractional seconds and also express timezone in an offset format. For example the following timestamp does not match: `2025-01-27T11:42:15.689800+00:00`.

This change simply moves the match group for fractions before the offset capture and makes it optional. The ordering and indexes of `matches` are kept the same.

It also allows fractions shorter than 3 digits, since RFC 3339 does not place such restriction and the fraction parsing later in this function handles it just fine.


nicksnyder:
Thanks for the contribution @liskaant! Would you be able to add a test case?

timostamm (approved):
The [Protobuf conformance test suite](https://github.com/bufbuild/protobuf-es/tree/main/packages/protobuf-conformance) includes several dozen tests for parsing Timestamp from JSON. That's why you didn't find explicit tests. 

The suite does not cover fractions with less than 3 digits and fractions with a TZ offset, but both should be valid. Thanks for the fix and the tests, @liskaant.

PR #1083, opened 2025-03-11
Clarify enum behavior FAQ entry for TypeScript 5.x

### LINKED ISSUE #1082: Update FAQ entry on TypeScript enums vs. string unions
[This FAQ entry](https://github.com/bufbuild/protobuf-es/blob/main/MANUAL.md#why-not-use-string-unions-for-protobuf-enumerations-instead-of-typescript-enum) mentions a reliance on TypeScript enum behavior that [has not existed](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#enum-overhaul) since 5.0.

It would be nice for the FAQ to be updated. I would do it myself if I had answers to my questions, but I do not 🌞:
- Is protobuf-es and/or connect-es relying on enum behavior that no longer exists in TypeScript? Is this just unsound from a type checking perspective?
- Is there a path toward a (breaking?) change that allows string unions instead of enums, given that the apparently main reasoning for choosing enums no longer applies?
- Are there other benefits of enums that are not explicitly mentioned there?

Thanks!

isker:
Closes #1082.


timostamm (approved):
Thanks, Ian.

Pushed up an admonition to the main section on enumerations in https://github.com/bufbuild/protobuf-es/pull/1083/commits/6a560ba31e5260eb119ddb7adae2af79ea71c1d2.

PR #1085, opened 2025-03-14
Update to google protobuf v30.0

timostamm:
This release includes new conformance tests that we fail. They are fixed in https://github.com/bufbuild/protobuf-es/pull/1084.


packages/protobuf/scripts/bootstrap-inject.mjs
```diff
@@ -302,15 +309,41 @@ function editionNumbersBetween(
 }
 
 /**
- * @param {FeatureSetDefaults} featureSetDefaults
+ * @param {import("@bufbuild/protobuf/reflect").ReflectMessage} featureSetDefaults - google.protobuf.FeatureSetDefaults
```
timostamm, line 312: Because we took some shortcuts in this script, it didn't handle the addition of the feature `enforce_naming_style`. Updated the script to use reflection.

packages/protobuf/src/registry.ts
```diff
@@ -427,7 +427,7 @@ const OPEN: FeatureSet_EnumType.OPEN = 1;
 
 // prettier-ignore
 // bootstrap-inject defaults: EDITION_PROTO2 to EDITION_2023: export const minimumEdition: SupportedEdition = $minimumEdition, maximumEdition: SupportedEdition = $maximumEdition;
-// generated from protoc v29.2
+// generated from protoc v30.0
```
timostamm, line 430: We're adding `enforce_naming_style`, but don't advertise support for edition 2024 yet, because it hasn't been released.

smaye81 (approved):
Is it worth updating to v30.1 since that was released last night?

timostamm:
> Is it worth updating to v30.1 since that was released last night?

I don't know, feel free to take a look.

PR #1104, opened 2025-04-15
Replace prettier and eslint with biome

timostamm:
`eslint` is a fantastic tool, but maintaining it has become a bit of a burden. This PR replaces `eslint` with [biome](https://github.com/biomejs/biome). Since `biome` also comes with a formatter that's very similar to `prettier`, we're replacing `prettier` as well.


biome.base.json
paul-sachs, line 1: Given that biome is more particular about comments, should this not be a biome.base.jsonc file instead? That will make sure we don't need to worry about converting it if we ever want to add comments

biome.base.json
```diff
@@ -0,0 +1,40 @@
...
+  "vcs": {
+    "enabled": false,
+    "clientKind": "git",
+    "useIgnoreFile": true
+  },
+  "files": {
+    "ignoreUnknown": false,
+    "ignore": [
+      "tsconfig.json",
+      "tsconfig.*.json",
```
paul-sachs, line 12: Digging into [docs](https://biomejs.dev/internals/language-support/#jsonc-support) looks like they have specific rules to handle tsconfig.json files, I think the problem we are running into is that it doesn't recognize `tsconfig.*.json` files as having the same exception. We could also solve this by adding our own overrides, like so:

```json
 "overrides": [
    {
      "include": ["tsconfig.*.json"],
      "json": {
        "parser": {
          "allowComments": true,
          "allowTrailingCommas": true
        },
        "formatter": {
          "trailingCommas": "none"
        }
      }
    }
  ]
``` 

This should allow us to match the expected behaviour on our custom tsconfig.*.json files without having to ignore them completely. WDT?

biome.root.json
paul-sachs, line 1: Out of curiosity, why is this named `biome.root.json` and not just `biome.json`? I guess what I'm thinking through is what config file the editor will use for top level files (files not in packages dir). By default at least, vscode seems to want to use `biome.json` instead.

packages/protobuf-test/src/reflect/reflect.test.ts
```diff
@@ -403,8 +403,8 @@ describe("ReflectMessage", () => {
     });
     test("returns error setting float for int", () => {
       const f = desc.field.singularInt32Field;
-      const err = catchFieldError(() => r.set(f, 3.142));
-      expect(err?.message).toMatch(/^expected number \(int32\), got 3.142$/);
+      const err = catchFieldError(() => r.set(f, 3.14));
```
paul-sachs, line 406: Another curious change... The things you learn. I guess we could also use `Math.PI` 🤷

packages/protobuf/src/reflect/scalar.ts
```diff
@@ -113,6 +108,10 @@ export function scalarZeroValue<
       >;
     case ScalarType.BYTES:
       return new Uint8Array(0) as ScalarValue<T, LongAsString>;
+    default:
```
paul-sachs, line 111: ooo, i like this lint rule, that's nice.

packages/bundle-size/biome.json
paul-sachs, line 1: One annoying aspect of the vscode extension for biome is that is currently seems to only look at the root level biome.json file. I think we can workaround that with workspaces but i'll investigate to see what that means exactly.

paul-sachs: Digging into this (and why the vscode extension is the way it is), it seems they haven't officially completed monorepo support (See https://github.com/biomejs/biome/issues/2228). We can work around this via the suggested method by having a single biome.jsonc file with a bunch of excludes per package. Until 2.x is released and monorepo support is official, I think that might be the best approach. WDT?

packages/protobuf-test/src/reflect/reflect.test.ts
timostamm: This was a fun one :) In this case, we don't want to use Math.PI, but it's a good lint rule.

biome.root.json
timostamm: The JetBrains plugin also only considers biome.json. Renamed in https://github.com/bufbuild/protobuf-es/pull/1104/commits/344462b5dd2593b69ad77c65b3fae4b6f460543b.

packages/bundle-size/biome.json
timostamm: I agree with the first comment on this issue: "run biome in each package and run it in parallel with like a pnpm (or whatever your'e using for workspaces) script?"

This is what we're doing here with turborepo. And since turbo caches per package, crossing packages in npm scripts seems like a bad idea. I wish there were fewer, better tools 😞

biome.base.json
paul-sachs: Yeah probably not. It's good to know, if not just so we know how to handle it either way. I'm good with ignoring the tsconfig.*.json

packages/bundle-size/biome.json
paul-sachs: Yeah i guess that's fair, the problem is that IDEs don't run it like that currently, so with that approach, you have your IDE doing one thing and the command-line doing something different. Not my favourite place to be for sure 😢

paul-sachs (approved):

PR #1114, opened 2025-04-25
Add function qualifiedName

timostamm:
This adds the export `qualifiedName` to @bufbuild/protobuf/reflect. 

The function takes a descriptor as an argument, and returns the [fully-qualified name](https://protobuf.com/docs/language-spec#fully-qualified-names). 

It's just a few lines of code, but it's useful for debugging, creating a key to cache an expensive operation, and similar purposes.


packages/protobuf-test/src/reflect/names.test.ts
```diff
@@ -12,9 +12,87 @@
...
+  qualifiedName,
+} from "@bufbuild/protobuf/reflect";
+import type { DescFile } from "@bufbuild/protobuf";
+import { compileField, compileFile } from "../helpers.js";
+
+describe("qualifiedName", () => {
+  let testFile: DescFile;
+  beforeAll(async () => {
+    testFile = await compileFile(`
+syntax = "proto3";                            // Fully-qualified name
```
jhump, line 28: Ooh, I recognize this 😂

packages/protobuf/src/reflect/names.ts
```diff
@@ -12,6 +12,36 @@
...
+      return desc.parent.typeName + "." + desc.name;
+    case "enum_value": {
+      const p = desc.parent.parent
+        ? desc.parent.parent.typeName
+        : desc.parent.file.proto.package;
+      return p + (p.length > 0 ? "." : "") + desc.name;
+    }
+    case "service":
+    case "message":
+    case "enum":
```
jhump, line 37: I guess these two must already incorporate parent names in their `typeName` property (for nested messages and enums)?

packages/protobuf/src/reflect/names.ts
```diff
@@ -12,6 +12,36 @@
...
+    case "extension":
+      return desc.typeName;
+    case "file":
+      return desc.proto.name;
```
jhump, line 41: This one is iffy.

In protobuf-go, I think `FullName()` for a file returns the package name (since a file path is not really a qualified name). But in the descriptor protos, the `name` attribute is indeed the path.

🤷

jhump (approved):
LGTM

Issue #941, opened 2024-07-23
V2: export type MessageInit


ngbrown:
In V2, there is a type, `MessageInit` that excludes `$typeName` (and `$unknown`) and when passing objects to functions that will call `create()` it is useful to use a derived type like that.

Currently in V2, I have to do something like this: `MessageInitShape<typeof MyTypeSchema>`, but it would be shorter to do this: `MessageInit<MyType>`.

So this request is to export the type `MessageInit` so it can be imported.

Also, having simpler types available might speed up the TypeScript language service, since it seems to have gotten fairly slow while I was trying out the V2 beta.

timostamm:
Hey Nathan, 

we don't export `MessageInit` for forward compatibility. Imagine that we want to accept field representations in `create()` that cannot be inferred from the generated type. `MessageInitShape` allows us to attach a separate type for init to the descriptor, and use it for `create()`.

We expect that you would typically want to write a generic function that works with arbitrary messages. We have an example in the v2 manual:

```ts
import { getOption, type DescMessage, type MessageShape } from "@bufbuild/protobuf";
import { reflect } from "@bufbuild/protobuf/reflect";
import { sensitive } from "./gen/example-option_pb";

export function redact<Desc extends DescMessage>(
  schema: Desc,
  message: MessageShape<Desc>,
) {
  const r = reflect(schema, message);
  for (const field of r.fields) {
    if (getOption(field, sensitive)) {
      // This field has the option (example.options.sensitive) = true
      r.clear(field);
    }
  }
}
```

> [!TIP]
>
> - `EnumShape` extracts the enum type from an enum descriptor.
> - `MessageShape` extracts the type from a message descriptor.
> - `MessageInitShape` extracts the init type from a message descriptor - the initializer object for `create()`.

The function could also take an init object with `MessageInitShape<Desc>`. For this pattern, the types are pretty concise - `MessageInit<MessageShape<Desc>>` would actually be more verbose.

But your function is a different case, it's specific to one message. I'd love to better understand the use case. What kind of function are you using it for?

ngbrown:
Hi Timo,

Thanks for explanation. I'm not sure I totally understand your reasoning though. The message descriptor is really doing a lot because it also has the JSON type reference (when the `json_types=true` option is turned on). I also get errors because the `MessageInitShape` is doing too much. See my examples below.

[...]

PR #1115, opened 2025-04-27
Export MessageInit type

polRk:


timostamm:
https://github.com/bufbuild/protobuf-es/issues/941

Issue #1056, opened 2025-01-17
Provide a strong type name in the generated message schema


timostamm:
From https://github.com/bufbuild/protobuf-es/issues/994#issuecomment-2448975214:

It would be very convenient to include a strongly typed type name in the schema for a generated message. For example:

```proto
syntax = "proto3";
package example;
message User {
  string first_name = 1;
}
```

This definition generates a `type User` and a `const UserSchema`. The type's `$typeName` property is strongly typed, but the schema's `typeName` property is not:

```ts
const user: User = create(UserSchema);
user.$typeName; // "example.User"

UserSchema.typeName; // string
```

This means the schema's property cannot be used to construct an instance of the type, and it cannot be used to narrow down a message by using the property as a descriminator.

If it's possible without breaking changes, we should change the generated schema to support this.

PR #1131, opened 2025-05-14
Provide a strong type name in the generated message schema

timostamm:
This makes the `typeName` property of generated message schemas a literal type.

For example, given this proto:

```proto
syntax = "proto3";
package example;
message User {
  string first_name = 1;
}
```

```ts
import { UserSchema } from "./gen/example_pb";

UserSchema.typeName; // "example.User"
```

Previously, `typeName` was just a `string`.

Closes https://github.com/bufbuild/protobuf-es/issues/1056.

### DIFF
--- a/packages/protobuf-test/src/types.test.ts
+++ b/packages/protobuf-test/src/types.test.ts
@@ -51,10 +51,9 @@ describe("type Message", () => {
           throw new Error();
       }
     });
-    test("cannot switch on Message.$typeName against embedded desc's typeName", () => {
+    test("cam switch on Message.$typeName against embedded desc's typeName", () => {
       switch (msg.$typeName) {
         case UserSchema.typeName:
-          // @ts-expect-error TS2339
           expect(msg.firstName).toBeDefined();
           break;
         default:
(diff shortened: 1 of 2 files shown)


packages/protobuf-test/src/types.test.ts
```diff
@@ -51,10 +51,9 @@ describe("type Message", () => {
           throw new Error();
       }
     });
-    test("cannot switch on Message.$typeName against embedded desc's typeName", () => {
+    test("cam switch on Message.$typeName against embedded desc's typeName", () => {
```
smaye81, line 54:
```suggestion
    test("can switch on Message.$typeName against embedded desc's typeName", () => {
```

smaye81 (approved):

packages/protobuf-test/src/types.test.ts
timostamm: TY, https://github.com/bufbuild/protobuf-es/pull/1131/commits/2d380ceaf1e354c8bb324896e16d185baf312c75

Issue #1038, opened 2024-12-16
FR: Restrict expected case type (camel vs snake) when parsing JSON


davidfiala:
I didn't see any obvious way to restrict the expected type of proto field naming when parsing from JSON in javascript/typescript buf.

[JsonWriteOptions](https://github.com/bufbuild/protobuf-es/blob/2b56e80671c28c1f7b2c6946b4a1e7c13bad0aad/packages/protobuf/src/to-json.ts#L65) allows you to specify an output type for fields as snake_case or camelCase.

However the corresponding [JsonReadOptions](https://github.com/bufbuild/protobuf-es/blob/2b56e80671c28c1f7b2c6946b4a1e7c13bad0aad/packages/protobuf/src/from-json.ts#L62) lacks the ability to restrict to a specific casing format such as snake or camel.

This creates an ambiguity. And arguably a security risk. For instance, a code reviewer may not realize that a sensitive property was overridden further down in a JSON config.

In practice, it appears that whichever field is set last is the winner. Example code below.

From https://protobuf.dev/programming-guides/json/

> Use proto field name instead of lowerCamelCase name: By default the protobuf JSON printer should convert the field name to lowerCamelCase and use that as the JSON name. An implementation may provide an option to use proto field name as the JSON name instead. Protobuf JSON parsers are required to accept both the converted lowerCamelCase name and the proto field name.

In a sense, protobuf-es is perfectly compliant with the behavior quoted above, if we accept the ambiguity. But I believe it would be safer to provide a restricted option, which wouldn't necessarily need to break existing users. Such an option remains compliant with the ambiguous quote above if the default remains to accept whichever field comes last during parsing.

```typescript
// Strawman proposal:
export interface JsonReadOptions {
  // ...
  mandatoryProtoFieldName?: 'camel' | 'snake' | 'accept-last-parsed'; // default to 'accept-last-parsed' to make the behavior explicit
}
```

I'd love to get the team's thoughts. Thank you.

```typescript
it('spaces', () => {
    const str = JSON.stringify({
      field_name_with_spaces: 'hello world',
    });

    const msg = fromJsonString(cfg.UnitTestConfigSchema, str);
    expect(msg.fieldNameWithSpaces).toBe('hello world');
  });

  it('camel', () => {
    const str = JSON.stringify({
      fieldNameWithSpaces: 'hello world',
    });

    const msg = fromJsonString(cfg.UnitTestConfigSchema, str);
[...]

PR #1132, opened 2025-05-17
Add tests for field casing, duplicate field ordering, and unknown field errors when performing JSON deserialization

hudlow:


packages/protobuf-test/src/json.test.ts
```diff
@@ -833,6 +857,56 @@ describe("JsonWriteOptions", () => {
...
+        '{ "singularStringField": "b" }',
+      );
+
+      expect(a.singularStringField).toBe("a");
+      expect(b.singularStringField).toBe("b");
+    });
+  });
+  describe("duplicate fields", () => {
+    // This depends on the ECMA-262-defined JSON.parse() behavior, which is itself
+    // specified to choose the last field value for a duplicate field.
```
timostamm, line 878: This behavior has been discussed in https://github.com/bufbuild/protobuf-es/issues/1038. I'm not sure these tests are particularly useful, because they basically test JSON.parse 🤔 

I just noticed that the [ProtoJSON documentation](https://protobuf.dev/programming-guides/json/) was recently [updated](https://github.com/protocolbuffers/protocolbuffers.github.io/commit/c18c8a72de22b99eaf536eb1f961371510159714#diff-98c94e5ab37db71fd002eff0d9cb2cfae9b0058ea00b412b466312ad819029d7) with the following:

> Multiple values for
singular fields (using duplicate or equivalent JSON keys) are accepted and the
last value is retained, as with binary format parsing. Note that not all
protobuf JSON parser implementations are conformant, and some nonconformant
implementations may reject duplicate keys instead.

With JSON.parse, the last property also wins for non-singular fields:

```ts
    const a = fromJsonString(
      proto3_ts.Proto3MessageSchema,
      '{ "repeatedStringField": ["b"], "repeatedStringField": ["a"] }',
    );
    const b = fromJsonString(
      proto3_ts.Proto3MessageSchema,
      '{ "repeatedStringField": ["a"], "repeatedStringField": ["b"] }',
    );

    expect(a.repeatedStringField).toStrictEqual(["a"]);
    expect(b.repeatedStringField).toStrictEqual(["b"]);
```

Quite different to the binary format, where repeated values accumulate...

hudlow: > This behavior has been discussed in https://github.com/bufbuild/protobuf-es/issues/1038.

I commented there.

> I'm not sure these tests are particularly useful, because they basically test JSON.parse 🤔

That's right, and in that sense they are certainly not _unit_ tests. I think they are worthwhile for completeness, but I'm certainly open to being convinced they're not necessary in this testing context.

> I just noticed that the [ProtoJSON documentation](https://protobuf.dev/programming-guides/json/) was recently [updated](https://github.com/protocolbuffers/protocolbuffers.github.io/commit/c18c8a72de22b99eaf536eb1f961371510159714#diff-98c94e5ab37db71fd002eff0d9cb2cfae9b0058ea00b412b466312ad819029d7)

Yeah, I read that and it's part of what prompted me to define the as-is behavior as the "correct" baseline. I was actually doing that _in preparation_ for filing what I now know would have been a dupe of #1038. 🤪

> Quite different to the binary format, where repeated values accumulate...

I think you mean to say that ProtoJSON arrays are different from the binary format where array elements can be serialized non-contiguously. 😉

packages/protobuf-test/src/json.test.ts
```diff
@@ -833,6 +857,56 @@ describe("JsonWriteOptions", () => {
   });
 });
 
+describe("parsing funny JSON", () => {
+  describe("field casing", () => {
+    test("is case agnostic", () => {
```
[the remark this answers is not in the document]
hudlow: Removed.

packages/protobuf-test/src/json.test.ts
hudlow: > Arrays are non-contiguous in the binary format - well said. It just occurred to me that behavior is also different with singular message fields. They can occur multiple times in the binary format, which will [merge](https://protobuf.com/docs/language-spec#merging-protobuf-messages) them, unlike ProtoJSON.

Interesting; I don't think I comprehend the benefit of this for a wire format. (And, as with ignoring duplicated JSON fields, I _do_ see the drawbacks.)

There are also some interesting things around the edges of the merge algorithm, if I'm reading it right:

- Map values for the same key don't get recursively merged even when they are a mergeable type (a repeated field, a message, or another map). This means (I think) that you couldn't replicate the merge logic in two ProtoJSON representations without considering the underlying types. (Probably not a huge deal, but a small blow to the self-describing benefits of JSON serialization.)
- Like JSON, Protobuf is missing a set type, so there's no way to signal that a merged repeated field should be deduplicated.

hudlow: > I don't think I comprehend the benefit of this for a wire format.

Is the benefit that you can stream a Protobuf message and update a field along the way?

hudlow:
Currently blocked by https://github.com/protocolbuffers/protobuf/pull/21873 which is ongoing.

hudlow:
Should be up-to-date and ready for another look.

PR #1194, opened 2025-08-04
Prepare for Jest v30

timostamm:
Jest 30 [removes aliases for matcher functions](https://jestjs.io/docs/upgrading-to-jest30#removal-of-alias-matcher-functions). This PR updates the code to use the replacement.


packages/protobuf-test/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "include": ["src/**/*.ts"],
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    // we require this setting for the npm package "long"
-    "allowSyntheticDefaultImports": true
+    // We require this setting for the npm package "long"
+    "allowSyntheticDefaultImports": true,
+    // We use Node.js modules in tests
+    "types": ["node"]
```
timostamm, line 8: Since we actually use Node.js modules in tests, we need the types. 

`@jest/types` 29.6.3 has this line in build/index.d.ts:

```
/// <reference types="node" />
```

So it actually still pulled in Node.js types, despite us explicitly configuring `"types": []`.

v30 does not (which is great).

PR #1195, opened 2025-08-04
Add a test case for broken map field parsing

timostamm:
This adds a dedicated test case for the map field parsing issue fixed by https://github.com/bufbuild/protobuf-es/pull/1183.


packages/protobuf-test/src/binary.test.ts
```diff
@@ -116,6 +119,20 @@ describe(`binary serialization`, () => {
...
+    const msg = create(MapsMessageBug1183Schema, {
+      map: {
+        [str128bytes]: 1,
+      },
+    });
+    const bytes = toBinary(MapsMessageBug1183Schema, msg);
+    const msg2 = fromBinary(MapsMessageBug1183Schema, bytes);
+    expect(msg2.map).toStrictEqual({
+      [str128bytes]: 1,
+    });
```
timostamm, line 133: This expectation currently fails with:

```
      Object {
    -   "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx": 1,
    +   "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx": 0,
      }
```

The map entry message is more than 128 bytes long, which is a 2-byte varint. This causes the while-loop in `readMapEntry` to exit early, and never read the value field, falling back to the implicit default `0`.

packages/protobuf-test/src/binary.test.ts
```diff
@@ -116,6 +119,20 @@ describe(`binary serialization`, () => {
...
+    expect(msg2.$unknown).toBeUndefined();
```
timostamm, line 134: This expectation fails with:

```
expect(received).toBeUndefined()
Received: [{"data": [1], "no": 2, "wireType": 0}]
```

This is the value field of the map entry. Because `readMapEntry` exited too early, this tag is treated as an unknown field of the message containing the map field.

If the message contains a field with number 2, `fromBinary` will try to parse this map value as field 2, possibly leading to a runtime error or more data corruption.

srikrsna-buf (approved):

PR #1240, opened 2025-09-24
Cache list of sorted fields in WeakMap

### LINKED ISSUE #1238: Sorting fields when serializing causes major performance penalty
Related to #333, but this deals with a specific issue.

When serializing a message, `ReflectMessageImpl` lazily computes a sorted list of the fields from the schema: https://github.com/bufbuild/protobuf-es/blob/d7cc04b62888a645d31c8da5de5e6755950ee64e/packages/protobuf/src/reflect/reflect.ts#L77-L85

However, this sorted list is recreated on **every single serialization** of a given message type - there is no persistent caching.

We have a particular message which has a large number of fields, and our app creates and serializes this message at a high rate. We applied a patch to cache this sorted list of fields, and saw significantly improved performance.

To demonstrate, I made [this](https://github.com/cptpcrd/proto-perf-test) minimal example, which includes a [patch](https://github.com/cptpcrd/proto-perf-test/blob/master/patches/%40bufbuild%2Bprotobuf%2B2.9.0.patch) that causes the sorted list to be cached on the schema instead of the `ReflectMessageImpl`. The speed improvement from the patch is very clear and consistent:

```bash
# Unpatched
$ node perf-test.mjs
Serialization: 5.468s

# Patched
$ node perf-test.mjs
Serialization: 3.023s
````

The patch is obviously a bit hacky, and other strategies are possible (e.g. precomputing the sorted list when loading the schema, or storing it in a `WeakMap<DescMessage, DescField[]>`, or perhaps simply sorting `DescMessage.fields` up-front), but anything along these lines would be a major improvement. I'd be happy to put together a PR implementing one of those strategies if there's interest.

cptpcrd:
Avoids re-sorting the list every time a message is serialized, which adds a sizable overhead for messages with many fields.

Closes #1238.

### DIFF
--- a/packages/bundle-size/README.md
+++ b/packages/bundle-size/README.md
@@ -16,11 +16,11 @@ usually do. We repeat this for an increasing number of files.
 
 | code generator      | files | bundle size |  minified | compressed |
 | ------------------- | ----: | ----------: | --------: | ---------: |
-| Protobuf-ES         |     1 |   132,764 b |  68,685 b |   15,747 b |
-| Protobuf-ES         |     4 |   134,953 b |  70,192 b |   16,465 b |
-| Protobuf-ES         |     8 |   137,715 b |  71,963 b |   16,986 b |
-| Protobuf-ES         |    16 |   148,165 b |  79,944 b |   19,314 b |
-| Protobuf-ES         |    32 |   175,956 b | 101,962 b |   24,767 b |
+| Protobuf-ES         |     1 |   132,844 b |  68,698 b |   15,784 b |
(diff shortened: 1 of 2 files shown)


timostamm:
ci / bundle-size can be fixed by running `npx turbo run bundle-size` and committing the changes.

timostamm:
Thanks! You'll have to sign the CLA before we can merge this. See https://github.com/bufbuild/protobuf-es/pull/1240#issuecomment-3329462296

timostamm (approved):

PR #1262, opened 2025-10-08
Fix nanos sign in durationFromMs()

timostamm:
The well-known type `Duration` has the [following comment](https://github.com/protocolbuffers/protobuf/blob/v32.1/src/google/protobuf/duration.proto#L108):

> For durations of one second or more, a non-zero value for the `nanos` field must be  of the same sign as the `seconds` field.

This PR fixes the behavior of `durationFromMs()` from `@bufbuild/protobuf/wkt` (added in #1244, unreleased) accordingly.


tdl76942-gif:
Mucho más gracias una pregunta cuál para un diagrama para ver teléfono a
larga distancia ver el diaposito y desbloquear

Issue #1313, opened 2025-12-11
fromJson: null values in maps throw an error


jagobagascon:
# Problem
Hi! I'm using the `google.protobuf.Value` field type to store arbitrary data into some fields. This type allows null values and correctly serialises and deserialises into JSON.

But I found that **deserialising a JSON with null values fails when the `google.protobuf.Value` is part of a map field**.

This message for example:
```proto
message ExampleMessage {
  google.protobuf.Value field1 = 1;
  map<string, google.protobuf.Value> field2 = 2;
}
```

## Allows `null` values in `field1`:

Works fine when parsing a `null` value as part of the `field1`:
```ts
fromJson(ExampleMessageSchema, {
    field1: null,
});
// Generates:
// {
//   '$typeName': 'example.ExampleMessage',
//   field2: {},
//   field1: {
//     '$typeName': 'google.protobuf.Value',
//     kind: { case: 'nullValue', value: 0 }
//   }
// }
```

## Fails when `null` is used inside the map:

But it fails if the `null` is part of the `field2`:
```ts
fromJson(ExampleMessageSchema, {
    field2: {
        value: null,
    },
});
// throws
// cannot decode field example.ExampleMessage.fields from JSON: map value must not be null
```

## Serialisation and deserialisation are inconsistent

Interestingly, if we convert our protobuf message with `null` values inside the map into JSON

```ts
const message = create(ExampleMessageSchema, {
    field2: {
        value: {
            kind: { case: "nullValue", value: NullValue.NULL_VALUE },
        },
    },
});
const jsonMessage = toJson(ExampleMessageSchema, message);
```

It works as expected, and generates the following JSON:
```ts
{ field2: { value: null } }
```

But if we try to convert that same JSON back to the message, then it fails:
```ts
fromJson(ExampleMessageSchema, jsonMessage)
// throws an exception: cannot decode field example.ExampleMessage.fields from JSON: map value must not be null
```

## Expected result

Serialising a message to JSON and then deserialising it should not fail.

## Example repo
I've created a repository with some examples to reproduce the issue:

https://github.com/jagobagascon/protobuf-es-maybe-bug

timostamm:
You found a bug! 🎉 

Thanks for the report, we'll fix it soon.

PR #1314, opened 2025-12-11
Permit google.protobuf.Value.null_value in map values in ProtoJSON

timostamm:
When parsing ProtoJSON, `fromJson` incorrectly rejects `null` in map values and repeated fields using `google.protobuf.NullValue` and `google.protobuf.Value` (with field `null_value` selected).

To illustrate, consider this Protobuf message:

```proto
syntax="proto3";
import "google/protobuf/struct.proto";
message M {
  map<string, google.protobuf.Value> value_map = 1;
  map<string, google.protobuf.NullValue> null_value_map = 2;
  repeated google.protobuf.Value value_list = 3;
  repeated google.protobuf.NullValue null_value_list = 4;
}
```

The following JSON should parse without error, but does not.

```json
{
    "valueMap": { "val1": null },
    "nullValueMap": { "val1": null },
    "valueList": [null],
    "nullValueList": [null],
}
```

This PR rectifies `fromJson`, but still rejects `null` for other repeated fields and map values.

Fixes https://github.com/bufbuild/protobuf-es/issues/1313

### DIFF
--- a/packages/protobuf-test/src/issue1313.test.ts
+++ b/packages/protobuf-test/src/issue1313.test.ts
@@ -1,0 +1,57 @@
+// Copyright 2021-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


hudlow:
Looks good, but my review revealed dead branches — perhaps these are indicative of a different oversight, but if not I think we should clean them up. https://github.com/bufbuild/protobuf-es/pull/1315

PR #1315, opened 2025-12-11
Remove dead branches for null handling in fromJson

hudlow:


packages/protobuf/src/from-json.ts
```diff
@@ -477,17 +445,8 @@ function scalarFromJson(
...
-): ScalarValue | typeof tokenNull | JsonValue;
-function scalarFromJson(
-  field: DescField & { scalar: ScalarType },
-  json: JsonValue,
-  nullAsZeroValue: boolean,
 ): ScalarValue | typeof tokenNull | JsonValue {
   if (json === null) {
-    if (nullAsZeroValue) {
-      return scalarZeroValue(field.scalar, false);
-    }
```
hudlow: All calls that passed `nullAsZeroValue` as `true` also guaranteed that `json` was not `null`.

packages/protobuf/src/from-json.ts
```diff
@@ -451,45 +442,18 @@ function readEnum(
...
-const tokenNull = Symbol();
-
 /**
  * Try to parse a JSON value to a scalar value for the reflect API.
  *
  * Returns the input if the JSON value cannot be converted. Raises a FieldError
  * if conversion would be ambiguous.
  *
- * JSON null returns the scalar zero value.
+ * JSON null returns the scalar type's default value.
```
timostamm: The function no longer accepts `null`, so this comment line should be removed.

PR #1322, opened 2025-12-25
Fix optional chaining in Deno check for Int64 support in proto-int64.ts

ltsg123:
in < Chrome 71，browser not support globalthis，from:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis


timostamm:
Thanks for the fix. This isn't the only reference to `globalThis` though, and exhaustively replacing all relevant sites is much more involved. We will end up implementing a polyfill for `globalThis`, which isn't trivial. 

Chrome 71 is seven years old by now. I recommend to add a polyfill to your application if you still need to support Chrome 71. 
Adding a comment to tsconfig.json in https://github.com/bufbuild/protobuf-es/pull/1331.

PR #1323, opened 2025-12-29
Fix validation for FieldMask (de-)serialization to/from JSON

### LINKED ISSUE #1321: FieldMask shouldn't convert names with trailing underscore
When converting a field mask to JSON, path values with a trailing underscore are converted instead of being rejected.

```js
import { create, toJson, fromJson } from "@bufbuild/protobuf";
import { FieldMaskSchema } from "@bufbuild/protobuf/wkt";

var msg = create(FieldMaskSchema, { paths: ["abc_"] });
console.log(msg);

var json = toJson(FieldMaskSchema, msg);  // Should fail but works
console.log(json);

var msg2 = fromJson(FieldMaskSchema, json);  // The underscore is lost
console.log(msg2);
```

In the same situation, the official Python implementation fails:
```python
from google.protobuf.field_mask_pb2 import FieldMask
from google.protobuf.json_format import MessageToJson

msg = FieldMask(paths=["abc_"])
MessageToJson(msg)  # ValueError: Fail to print FieldMask to Json string: Trailing "_" in path name abc_.
```

hudlow:
Fixes #1321, adds tests to match Google's protobuf-go behavior.

### DIFF
--- a/packages/protobuf/src/reflect/names.ts
+++ b/packages/protobuf/src/reflect/names.ts
@@ -87,7 +87,7 @@ export function protoCamelCase(snakeCase: string): string {
 /**
  * Converts protoCamelCase to snake_case.
  *
- * This function is the reverse of function protoCamelCase. Not that some names
+ * This function is the reverse of function protoCamelCase. Note that some names
  * are not reversible - for example, "foo__bar" -> "fooBar" -> "foo_bar".
  */
 export function protoSnakeCase(lowerCamelCase: string): string {


packages/protobuf/src/from-json.ts
```diff
@@ -794,13 +796,12 @@ function fieldMaskFromJson(fieldMask: FieldMask, json: JsonValue) {
     return;
   }
   function camelToSnake(str: string) {
-    if (str.includes("_")) {
+    if (!camelCasePath.test(str)) {
       throw new Error(
-        `cannot decode message ${fieldMask.$typeName} from JSON: path names must be lowerCamelCase`,
+        `cannot decode message ${fieldMask.$typeName} from JSON: path names must be camel case`,
```
hudlow: Changed the language because an underscore prefix can roundtrip to UpperCamelCase and back.

packages/protobuf/src/to-json.ts
```diff
@@ -485,14 +485,22 @@ function durationToJson(val: Duration) {
   return text + "s";
 }
 
+const validPath = /^[a-zA-Z_][a-zA-Z0-9_]*(\.[a-zA-Z_][a-zA-Z0-9_]*)*$/;
+const reversiblePath =
+  /^_?[a-z][a-z0-9]*(_[a-z][a-z0-9]*)*(\._?[a-z][a-z0-9]*(_[a-z][a-z0-9]*)*)*$/;
```
hudlow: I don't really trust regular expressions written to capture all illegal behavior. Seems much more correct to me to define legal syntax.

These are split into two because, well, the official Go implementation distinguishes between the error cases. If we do not care about distinguishing the errors, only the latter regular expression is needed.

packages/protobuf/src/from-json.ts
```diff
@@ -794,13 +796,12 @@ function fieldMaskFromJson(fieldMask: FieldMask, json: JsonValue) {
...
       );
     }
-    const sc = str.replace(/[A-Z]/g, (letter) => "_" + letter.toLowerCase());
-    return sc[0] === "_" ? sc.substring(1) : sc;
```
hudlow: Scary to delete this, but it seems... incorrect?

AdrienVannson (approved):

s40758976-source:
- Adrien comentó que la lógica de reemplazo en el código parecía ser
   incorrecta.
   - Adrien aprobó la solicitud de extracción que soluciona la validación
   de FieldMask.

packages/protobuf/src/from-json.ts
timostamm: We should not change the language. 

"lowerCamelCase" is used in the Protobuf and protobuf-es documentation, and refers to the conversion defined in code by `protoc`. While this logic can produce UpperCamelCase in rare cases (e.g. "_foo" -> "Foo"), the updated error message is misleading: path names cannot be UpperCamelCase or lowerCamelCase, they have to be "lowerCamelCase" as defined by protoc. 

Updating the docs with related clarification here: https://github.com/bufbuild/protobuf-es/pull/1332

Maybe it would be helpful to stop referring to `lowerCamelCase`, and use `protoCamelCase` instead (like the function `protoCamelCase` from @bufbuild/protobuf/reflect). But we should decide on that separately.

packages/protobuf/src/to-json.ts
timostamm: Hm, feels like we're digging in the wrong direction with more complex expression 🤔 

With the fix to the function `camelToSnake`, the function is the reverse of `protoCamelCase` from packages/protobuf/src/reflect/names.ts, right? We could use that to check reversibility and remove the second regular expression. It might be a bit slower (but only for extremely large field masks), but it should be much more concise.

We need to decide whether we want to follow the Go implementation. We can discuss in person.

packages/protobuf-test/src/field-mask-json.test.ts
```diff
@@ -0,0 +1,316 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { suite, test } from "node:test";
+import { create, equals, toJson, fromJson } from "@bufbuild/protobuf";
+import { FieldMaskSchema } from "@bufbuild/protobuf/wkt";
+import assert from "node:assert";
+
+void suite("FieldMask JSON compatibility", () => {
```
timostamm: We need to decide how to proceed with such additional tests. We can discuss in person.

packages/protobuf/src/to-json.ts
timostamm: Decided to drop field name validation for now in https://github.com/bufbuild/protobuf-es/pull/1323/commits/ae47be4ceec9fc54801084b0aa02beaa0e1cd2ae. Filed https://github.com/bufbuild/protobuf-es/issues/1335.

hudlow:
> Dan Hudlow seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

Please note that my commit in this branch was made while I was a Buf employee and the copyright is inherently owned by Buf.

packages/protobuf/src/reflect/names.ts
```diff
@@ -82,6 +84,19 @@ export function protoCamelCase(snakeCase: string): string {
   return b.join("");
 }
 
+/**
+ * Converts protoCamelCase to snake_case.
+ *
+ * This function is the reverse of function protoCamelCase. Not that some names
```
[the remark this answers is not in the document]
timostamm: TY!

srikrsna-buf (approved):

PR #1324, opened 2025-12-30
Enforce top-level regular expressions for performance

hudlow:
I noticed the lack of enforcement for top-level regular expressions when working on #1323. Biome has a rule to enforce this (excluding regular expressions with the `/g` flag, which are stateful), so I turned this on, and made the fixes.

There's obviously some noise in any given benchmark run, but I did see what _looked_ to me like a statistically-significant improvement.

image


timostamm (commented):
As far as I see, the changes don't affect code paths taken in the benchmarks. I think the diff is all noise. Needs a closer look.

Issue #1294, opened 2025-11-19
feature request: add option to strip protoc-gen-es version from code gen preamble


cody-dot-js:
love `protoc-gen-es`, it's **_so good_**, but...

it is simultaneously highly useful AND annoying that `protoc-gen-es` encodes the `protoc-gen-es` version into the preamble comment

would y'all be open to adding an option to strip the version from the preamble comment to prevent massive diffs like this?

```diff
-// @generated by protoc-gen-es v2.9.0 with parameter "import_extension=none,js_import_style=module,keep_empty_files=false,target=ts,ts_nocheck=false,json_types=true" 

+// @generated by protoc-gen-es v2.10.0 with parameter "import_extension=none,js_import_style=module,keep_empty_files=false,target=ts,ts_nocheck=false,json_types=true"
```

PR #1336, opened 2026-01-09
Add plugin option `elide_plugin_version=true`

timostamm:
This PR adds a new plugin option to[ @bufbuild/protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es) (and all plugins based on [@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin)):

> ### `elide_plugin_version=true`
> 
> By default, `protoc-gen-es` inserts its version number at the top of each generated file. To remove the version number, set `elide_plugin_version=true`. Be aware that we recommend to keep plugin and runtime versions in sync, and that this option can make it more difficult to spot a mismatch.

Closes https://github.com/bufbuild/protobuf-es/issues/1294

### DIFF
--- a/packages/protoc-gen-es/README.md
+++ b/packages/protoc-gen-es/README.md
@@ -114,6 +114,12 @@ time. Unless you use Bazel, you probably don't need this option.
 If you use compiler settings that yield an error for generated code, setting this option generates an annotation at
 the top of each file to skip type checks: `// @ts-nocheck`.
 
+### `elide_plugin_version=true`
+
+By default, [protoc-gen-es] inserts its version number at the top of each generated file. To remove the version number,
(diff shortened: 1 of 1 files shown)


jtszalay:
> Updating the plugin is likely to update the generated files anyway, right?

We don't find that to be true most of the time. These diffs as show in the image cause us to have ~300+ `linguist-generated=true` and bog down reviews in github. 

Screenshot 2026-01-09 at 11 16 02

sbarfurth:
> not commit generated files

We use Bazel for building and not updating files for which content doesn't change will ensure build actions don't need to run again. So even without checking things in this would save build time.

PR #1334, opened 2026-01-02
Cache field maps for fromJson()

hudlow:
@timostamm probably a much more meaningful performance impact than #1315. Here's what I'm seeing:

image

PR #1337, opened 2026-01-09
feat: add json_types_numeric_enums plugin option

realtimetodie:
Adds a new plugin option `json_types_numeric_enums=true` to allow users to generate numeric JSON types for every Protobuf enumeration.

This is especially useful when using the `toJson` conversion function with the `enumAsInteger: true` option.

Example

```protobuf
syntax = "proto3";

enum Format {
  FORMAT_UNSPECIFIED = 0;
  FORMAT_BINARY = 1;
  FORMAT_JSON = 2;
}
```

Instead of 

```ts
/**
 * @generated from enum Format
 */
export type "FORMAT_UNSPECIFIED" | "FORMAT_BINARY" | "FORMAT_JSON";
```

it will generate

```ts
/**
 * @generated from enum Format
 */
export type FormatJson = 1 | 2 | 3;
```

This has drastically improved performance and bundle size for our application.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -194,6 +194,10 @@ the top of each file to skip type checks: `// @ts-nocheck`.
 Generates JSON types for every Protobuf message and enumeration. Calling `toJson()` automatically returns the JSON type
 if available. Learn more about [JSON types](#json-types).
 
+### `json_types_numeric_enums=true`
+
+Generates numeric JSON types for every Protobuf enumeration.
+
 ### `valid_types` (experimental)
 
 Generates a Valid type for every Protobuf message. Possible values:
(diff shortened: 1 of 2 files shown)


timostamm:
Thanks for the PR, @realtimetodie. 

Unfortunately, the plugin option is problematic: 
- Well-known types are imported from `@bufbuild/protobuf/wkt`, and can't honor the option. 
- `toJson()` must be called with the option `enumAsInteger: true` to match the types, but it will fall back to `JsonValue` instead.

An alternative would be to always generate types for all combinations of the [JSON serialization options](https://github.com/bufbuild/protobuf-es/blob/v2.10.2/MANUAL.md#json-serialization-options) and honor them in `toJson`. That's eight permutations however - a lot to properly test, and a lot of code to generate. It would block to road to supporting validation information ([Valid types](https://github.com/bufbuild/protobuf-es/blob/v2.10.2/MANUAL.md#valid-types)) in JSON types. 

I think our initial design is still the best compromise:
> JSON types are specified to match exactly what `toJson()` will emit with standard serialization options.

I'll have to close the PR, but this doesn't mean that we don't see the use case. Please feel free to create a feature request. Long-term, we'd love to provide more flexibility in generated code. Instead of plugin options, we're thinking about providing this flexibility via the plugin framework ([@bufbuild/protoplugin](https://www.npmjs.com/package/@bufbuild/protoplugin)).

> This has drastically improved performance and bundle size for our application.

That's interesting. Is it `fromJson`? It looks up enum values by iterating over all values - this could be improved by applying a cache, similar to https://github.com/bufbuild/protobuf-es/pull/1334.

PR #1357, opened 2026-02-05
Remove support for non-default typescript settings

Zaczero:
Succeeds https://github.com/bufbuild/protobuf-es/pull/1347

I agree with @timostamm and @smaye81 that we should not put effort into supporting non-default TypeScript features. Having said that, I noticed huge potential in removing redundant code that was there to make it "strict"-friendlier. This avoids a huge bug surface and typing misuse which the previous PR was trying to fix. So we get less code, and fewer bugs, and easier maintenance. Win-win-win!

https://www.typescriptlang.org/docs/handbook/compiler-options.html#strict-boolean-default-false

I'll soon do a broader sweep to remove unnecessary code, but this seems like a good first quick win!

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -464,7 +464,7 @@ Protobuf-ES generates the following property:
 result:
   | { case: "number";  value: number }
   | { case: "error";   value: string }
-  | { case: undefined; value?: undefined } = { case: undefined };
+  | { case: "" };
 ```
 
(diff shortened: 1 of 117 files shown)


timostamm:
If this is the default config that `tsc` emits, it has changed behavior recently-ish, and we should re-evaluate the situation.

Can you file an issue with details?

PR #1369, opened 2026-03-04
Fix field type inference when FieldDescriptorProto.type is unset

### LINKED ISSUE #1368: Fields are incorrectly classified as scalar when `type` omitted from FieldDescriptorProto
I came across this bug while writing a Kafka consumer in NodeJS, using Redpanda's schema registry. When deserializing the key and value of the message using a schema registry client lib, scalar fields were deserialized just fine but message types were running into all sorts of error while reading tags and decoding their associated data. I was running into similar problems using both `kafkajs` + `@kafkajs/confluent-schema-registry` as well as `@confluentinc/kafka-javascript` and `@confluentinc/schemaregistry`. Chose to stick with the latter setup and started debugging. I'm here because `@confluentinc/schemaregistry` depends on `@bufbuild/protobuf` for its core protobuf logic.

I found that when `@confluentinc/schemaregistry` fetches the `format=serialized` form of a proto file from the registry server, and [calls `createFileRegistry()`](https://github.com/confluentinc/confluent-kafka-javascript/blob/33c52eed2e81ad167f19d0a9f58ef0e6a7ffd4c4/schemaregistry/serde/protobuf.ts#L430), the field descriptors for `message` fields incorrectly ended up with `fieldKind: 'scalar'`. This led to fields being misinterpreted when the actual Kafka message keys/values were later decoded, [here](https://github.com/bufbuild/protobuf-es/blob/e5d8be67695ece44c31164123f04cf782eae96d6/packages/protobuf/src/from-binary.ts#L150). I traced this back to the serialized proto file from the schema registry omitting `type`, in favor of `type_name`, as is permitted according to https://github.com/protocolbuffers/protobuf/blob/c223b2e1fa96feac05f275853e64a87993d557a2/src/google/protobuf/descriptor.proto#L295-L296,

```protobuf
  // If type_name is set, this need not be set.  If both this and type_name
  // are set, this must be one of TYPE_ENUM, TYPE_MESSAGE or TYPE_GROUP.
  optional Type type = 5;
```

jakewisse:
Some protobuf tooling (e.g. Confluent Schema Registry) omits the `type` field from FieldDescriptorProto when `type_name` is set. According to [`descriptor.proto`](https://github.com/protocolbuffers/protobuf/blob/c223b2e1fa96feac05f275853e64a87993d557a2/src/google/protobuf/descriptor.proto#L295-L296), this is valid : "If type_name is set, this need not be set."

Because descriptor.proto is proto2, protobuf-es places the default value (TYPE_DOUBLE = 1) on the prototype. When `type` is absent from the wire data it is never set as an own property, and `proto.type` inherits the prototype default, causing message and enum fields to be misclassified as scalar.

This resolves the field type from the registry via `type_name` when `type` is not explicitly set. From there, we thread the resolved type through `getFieldPresence()`, `isPackedField()`, and `isDelimitedEncoding()` to avoid the same prototype fallback in those functions.

Fixes #1368.

### DIFF
--- a/packages/protobuf-test/src/registry.test.ts
+++ b/packages/protobuf-test/src/registry.test.ts
@@ -548,6 +548,141 @@ void suite("createFileRegistry()", () => {
       );
     });
   });
+  void suite("FieldDescriptorProto with unset type", () => {
+    // Some protobuf tooling (e.g. Confluent Schema Registry) omits the FieldDescriptorProto `type`
+    // field from the wire data when `type_name` is set. This is valid per descriptor.proto: "If
+    // type_name is set, this need not be set." Since descriptor.proto is proto2, the unset field
+    // inherits the prototype default TYPE_DOUBLE = 1, causing message/enum fields to be
+    // misclassified as scalar.
+    test("infers message type from type_name", async () => {
+      const fds = await compileFileDescriptorSet({
+        "test.proto": `
+          syntax="proto3";
+          import "google/protobuf/timestamp.proto";
+          message Outer {
+            string name = 1;
+            google.protobuf.Timestamp created_at = 2;
+            Inner nested = 3;
+            repeated Inner items = 4;
+          }
+          message Inner {
+            string value = 1;
+          }
+        `,
+      });
+      const testProto = fds.file.find((f) => f.name === "test.proto");
+      assert.ok(testProto);
+      const outerMsg = testProto.messageType.find((m) => m.name === "Outer");
+      assert.ok(outerMsg);
+      // Strip the `type` own property from message-type fields to simulate wire data that omits it
+      for (const field of outerMsg.field) {
+        if (field.typeName) {
+          assert.ok(
+            Object.prototype.hasOwnProperty.call(field, "type"),
+            `expected "type" to be an own property before deletion`,
+          );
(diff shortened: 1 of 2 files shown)


timostamm:
Thanks for the PR, Jake. Can you sign the CLA?

jakewisse:
@timostamm sorry for the delay here, I actually did sign it right away but for some reason CLA assistant would never sync to the PR… Every time I click the link to sign it shows as already signed to me.

jakewisse:
Ah, it was my fault. I created the commit using an email associated with a different Github account. Looks like it has synced properly now.

timostamm:
Thanks for the patience, Jake! I just got a chance to take a look. 

The fix looks good, but I believe we need to take a different approach 🤔 As you can see, the change affects bundle size. It's an insignificant change, but I suspect that there are similar cases in descriptor.proto we'd want to normalize in the future, and then it becomes a problem. I'll comment on #1368.

timostamm:
See https://github.com/bufbuild/protobuf-es/issues/1368#issuecomment-4305542339 for broader thoughts. 

To detangle your PR from those design decisions, would you be open to refactoring this into a free function that operates on a `FileDescriptorProto`? 

We can land it as an internal helper with tests until we settle on the broader design. Happy to help if the rework turns out to be more than a small lift.

PR #1371, opened 2026-03-13
Support `exactOptionalPropertyTypes`

### LINKED ISSUE #1358: Investigate support for exactOptionalPropertyTypes (on by default from TypeScript 5.9)
### context

https://github.com/bufbuild/protobuf-es/pull/1347 was closed incorrectly

follow up at https://github.com/bufbuild/protobuf-es/pull/1357

### conclusion

The library is incompatible with the compiler standard setting exactOptionalPropertyTypes (as seen in `tsc --init`)

### expected result

1347 gets reopened and merged

---

cc @timostamm

haines:
Closes #1358

This PR adds support for the `exactOptionalPropertyTypes` TypeScript compiler option, which is on by default in `tsc --init` from Typescript 5.9 onwards. This necessitates adding `| undefined` to the typing of optional fields so that they may be explicitly set to undefined, not just omitted.

I've enabled the option in the compiler settings for this project to catch any issues, which meant adding a few `| undefined`s to types used internally as well.

I didn't add it to the JSON types because doing so means either adding `| undefined` to the fields of `JsonObject` (which seems wrong when deserializing from JSON can never produce an `undefined` field) or removing the constraint that `MessageJsonType extends JsonValue` (which seems wrong because the JSON types _are_ JSON values).

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -373,7 +373,7 @@ Protobuf-ES generates the following property:
 /**
  * @generated from field: example.User manager = 4;
  */
-manager?: User
+manager?: User | undefined;
 ```
 
 Message fields don't have default values in Protobuf. They are always optional in ECMAScript.
@@ -531,7 +531,7 @@ For this group field, Protobuf-ES generates the following property and the messa
 /**
  * @generated from field: optional example.User.MyGroup mygroup = 1;
  */
-mygroup?: User_MyGroup;
+mygroup?: User_MyGroup | undefined;
 ```
 
 > [!CAUTION]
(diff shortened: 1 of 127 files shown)


timostamm:
Thanks for the PR, Andrew! I'm a bit distracted by other things right now, but I'll check it out soon.

timostamm (approved):
Thanks for the PR, Andrew 🙂 

I was afraid this would entail more changes in our runtime packages, but this all looks reasonable to me. I did push up a small change to `toJson` and `fromJson`, and updated the code snippets in the manual to match the new generated code. 

> I didn't add it to the JSON types

Agreed.

Zaczero:
I don't want to sound salty. But... my original PR #1347 was closed incorrectly, then you made me open an issue to reopen that PR, and instead you just resolved the issue yourself without giving a chance to contribute. And not even co-authored-by.

PR #1375, opened 2026-03-27
Change README title from Deno to Bun

gabszs:
Updated the title of the README to reflect the use of Bun instead of Deno.


timostamm:
Thanks! Can you sign the CLA?

PR #1385, opened 2026-04-17
Update to protocolbuffers/protobuf v34.0

emcfarlane:
This updates to protocolbuffers/protobuf [v34.0](https://github.com/protocolbuffers/protobuf/releases/tag/v34.0). Conformance testing will be done as a follow up.

Continuing on from https://github.com/bufbuild/protobuf-es/pull/1308 the `EDITION_UNSTABLE` enum is now special cased in the runtime and plugin edition-range checks, rather than treating it as a hard error. An unstable edition will fall back to the `maximumEdition` (currently `EDITION_2024`) feature to allow for testing.

Regenerated protobuf-test fixtures against v34.0.

Closes https://github.com/bufbuild/protobuf-es/pull/1361


packages/protoplugin/src/schema.ts
```diff
@@ -250,15 +250,19 @@ function getFilesToGenerate(
...
-          `${file.name}: unsupported edition ${editionToString(edition)} - the earliest supported edition is ${editionToString(minimumEdition)}`,
-        );
-      }
-      if (edition > maximumEdition) {
-        throw new Error(
-          `${file.name}: unsupported edition ${editionToString(edition)} - the latest supported edition is ${editionToString(maximumEdition)}`,
-        );
+      // EDITION_UNSTABLE is a sandbox for in-development features; accept
+      // it regardless of the plugin's min/max bounds.
+      if (edition !== Edition.EDITION_UNSTABLE) {
```
srikrsna-buf: `continue` on the inverse will reduce nesting.

srikrsna-buf (approved):

packages/protoplugin/src/schema.ts
emcfarlane: Fixed in [52d8fa4](https://github.com/bufbuild/protobuf-es/pull/1385/commits/52d8fa4a616c2198ad8077151957eea80522e18b)

timostamm (commented):
I don't see an update to the protobuf-conformance dep. Intentional? See https://github.com/bufbuild/protobuf-es/pull/1363

timostamm:
> Conformance testing will be done as a follow up.

Sorry, missed that. Works for me.

timostamm (approved):

PR #1386, opened 2026-04-21
Fix UTF-8 validation for proto3 and edition feature `utf8_validation`

emcfarlane:
This PR adds support for UTF-8 validation in `fromBinary`, and fixes a bug: proto3 and editions require strings to be valid UTF-8, but this was never checked in `fromBinary`. Invalid data would silently be replaced by the Unicode Replacement Character.

The new behavior is that `fromBinary` throws on invalid strings. If you parse binary data that contains invalid strings and rely on the old behavior, you have to handle the error, or configure text encoding to ignore the new `strict` argument:

```typescript
import { configureTextEncoding, getTextEncoding } from "@bufbuild/protobuf/wire";
const { encodeUtf8, checkUtf8, decodeUtf8 } = getTextEncoding();
configureTextEncoding({
  encodeUtf8,
  checkUtf8,
  // Ignore `strict`: invalid UTF-8 becomes U+FFFD instead of throwing.
  decodeUtf8: (bytes, _strict) => decodeUtf8(bytes),
});
```

With editions, UTF-8 validation can be configured per field or per file. See the [documentation](https://protobuf.dev/editions/features/#utf8_validation) for details.

Note: `BinaryReader` does not change behavior. To opt-in to UTF-8 validation, call `reader.string(true)`.

### DIFF
--- a/packages/bundle-size/README.md
+++ b/packages/bundle-size/README.md
@@ -16,11 +16,11 @@ usually do. We repeat this for an increasing number of files.
 
 | code generator      | files | bundle size |  minified | compressed |
 | ------------------- | ----: | ----------: | --------: | ---------: |
-| Protobuf-ES         |     1 |   133,010 b |  68,782 b |   15,803 b |
-| Protobuf-ES         |     4 |   135,199 b |  70,289 b |   16,433 b |
-| Protobuf-ES         |     8 |   137,961 b |  72,060 b |   16,996 b |
-| Protobuf-ES         |    16 |   148,411 b |  80,041 b |   19,330 b |
-| Protobuf-ES         |    32 |   176,202 b | 102,059 b |   24,783 b |
+| Protobuf-ES         |     1 |   133,886 b |  69,176 b |   15,907 b |
+| Protobuf-ES         |     4 |   136,075 b |  70,683 b |   16,591 b |
+| Protobuf-ES         |     8 |   138,837 b |  72,454 b |   17,109 b |
+| Protobuf-ES         |    16 |   149,287 b |  80,435 b |   19,458 b |
+| Protobuf-ES         |    32 |   177,078 b | 102,453 b |   24,888 b |
 | protobuf-javascript |     1 |   314,120 b | 244,024 b |   35,999 b |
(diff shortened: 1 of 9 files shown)


srikrsna-buf (commented):
This is potentially a breaking change, because some strings that silently passed will no longer do so. I am not saying that it is good thing to pass them just that it can bite people who inadvertently relied on accepting invalid UTF-8 sequences.

Instead of adding a new method to `BinaryReader` and bypassing the global `TextEncoding`, I think we can add an optional bool to both `BinaryReader.string` and `TextEncodin.decodeUtf8`. This will not be a breaking change for `TextEncoding` implementors and can also act as escape hatch for users to bypass the fatal error after the upgrade.

emcfarlane:
Thanks! Updated to the optional bool on the `string` and `decodeUtf8` methods in [eb157c9](https://github.com/bufbuild/protobuf-es/pull/1386/commits/eb157c9381d074df0db77ae21345849e0e8ea84f). 

For the breaking change: failing this check should mean aligning with the spec. Don't think it should be allowed to be overridden.

srikrsna-buf (approved):
`strict` maybe a better name, I know `TextDecoder` uses `fatal` but `strict` is what I have seen in other ES projects like tsc flag strict and the strict directive.

srikrsna-buf (approved):

timostamm (commented):
The implementation looks good to me, but:

This adds support for UTF-8 validation right? We need to make it clear what that means in this PR title and description - it's what drives our changelog in releases.

We're fixing a bug here: proto3 and editions require strings to be valid UTF-8, but this was never checked in `fromBinary`. Invalid data would silently be replaced by the Unicode Replacement Character.

With this change, `fromBinary` validates strings. This fix impacts users: Instead of silent replacement, invalid UTF-8 now causes `fromBinary` to throw. Users relying on the old behavior - for example because their backend produces invalid UTF-8 - will need to handle the error or fix the backend.

Can we add a code snippet with an escape hatch via `configureTextEncoding`?

We should also clarify that `BinaryReader` and `BinaryWriter` do not change behavior.

PR #1387, opened 2026-04-21
Fix binary tag parse validation to reject overlong values

emcfarlane:
This enforces binary tag parsing to reject values greater than uint32 sizes.

`BinaryReader.tag()` validates that the tag is a well-formed uint32 varint. Overlong varints (more than 5 bytes) and 5-byte varints whose value overflows uint32 are rejected with an error. Earlier versions of Protobuf-ES silently decoded these cases by truncating the high bits, which caused malformed payloads to be stored as unknown fields instead of failing to parse. Producers generating tags via `BinaryWriter.tag()` are unaffected, since `BinaryWriter` only emits valid tags.

### DIFF
--- a/packages/protobuf/src/wire/binary-encoding.ts
+++ b/packages/protobuf/src/wire/binary-encoding.ts
@@ -385,16 +385,23 @@ export class BinaryReader {
   }
 
   /**
-   * Reads a tag - field number and wire type.
+   * Reads a tag - field number and wire type. Tags are uint32 varints; values
+   * that do not fit in uint32 are rejected.
    */
   tag(): [number, WireType] {
-    let tag = this.uint32(),
-      fieldNo = tag >>> 3,
-      wireType = tag & 7;
-    if (fieldNo <= 0 || wireType < 0 || wireType > 5)
+    const start = this.pos;
+    const tag = this.uint32();
+    const bytesRead = this.pos - start;
+    if (bytesRead > 5 || (bytesRead == 5 && this.buf[this.pos - 1] > 0x0f)) {
+      throw new Error("illegal tag: varint overflows uint32");
+    }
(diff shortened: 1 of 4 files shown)


srikrsna-buf (approved):
If I understand this correctly, this matters because we end up storing them as unknown fields rather than failing on them, right?

emcfarlane:
This is adding validation to the binary decoding. ~We don't currently validate that the field numbers must be 1 through 2^29−1.~ Edit: wouldn't expect to see this issue, but yes it would truncate potentially making a malformed tag (>2^29-1) be truncated to another tag (most likely an unknown one).

timostamm (commented):
Similar to https://github.com/bufbuild/protobuf-es/pull/1386, we need to clarify what that means for users. `BinaryReader` is used by ts-proto, possibly others, and we want to give good guidance on how to handle this change.

packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -386,16 +386,23 @@ export class BinaryReader {
...
-    let tag = this.uint32(),
-      fieldNo = tag >>> 3,
-      wireType = tag & 7;
-    if (fieldNo <= 0 || wireType < 0 || wireType > 5)
+    const start = this.pos;
+    const tag = this.uint32();
+    const bytesRead = this.pos - start;
+    if (bytesRead > 5 || (bytesRead == 5 && this.buf[this.pos - 1] > 0x0f)) {
+      throw new Error("illegal tag: varint overflows uint32");
+    }
```
timostamm, line 398: It would be really good to add an inline comment here explaining this, because it isn't obvious why uint32 isn't rejecting. It's also a hot path, and we should investigate inlining the logic, cleaning it up and giving us a perf boost at the same time.

PR #1388, opened 2026-04-21
Fix Any JSON encoding for messages without a custom JSON representation

emcfarlane:
Replace `typeName.startsWith("google.protobuf.")` with an explicit allowlist (`hasCustomJsonRepresentation`) in `anyToJson` / `anyFromJson`, so only WKTs with a custom JSON form get the `{"@type","value"}` wrapper.

### DIFF
--- a/packages/protobuf/src/wkt/wrappers.ts
+++ b/packages/protobuf/src/wkt/wrappers.ts
@@ -66,10 +66,13 @@ export function isWrapperDesc(
 }
 
 /**
- * Returns true if the message type has a custom JSON representation per the
- * protobuf JSON spec. When packed inside `google.protobuf.Any`, these types
- * are serialized as `{"@type": ..., "value": <special form>}`; all other
- * messages embed their fields directly.
+ * Returns true if the descriptor is a well-known type with a custom JSON
+ * representation per the protobuf JSON spec. Examples: Timestamp as an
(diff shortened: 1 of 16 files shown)


srikrsna-buf (approved):

packages/protobuf/src/wkt/wrappers.ts
```diff
@@ -65,6 +65,27 @@ export function isWrapperDesc(
   );
 }
 
+/**
+ * Returns true if the message type has a custom JSON representation per the
+ * protobuf JSON spec. When packed inside `google.protobuf.Any`, these types
+ * are serialized as `{"@type": ..., "value": <special form>}`; all other
+ * messages embed their fields directly.
+ */
```
timostamm, line 76:
```suggestion
/**
 * Returns true if the descriptor is a well-known type with a custom JSON
 * representation per the protobuf JSON spec. Examples: Timestamp as an
 * RFC 3339 string, Duration as "5s", wrappers as the unwrapped scalar.
 *
 * When packed inside `google.protobuf.Any`, these messages are serialized
 * as `{"@type": ..., "value": <custom form>}`; all other messages embed
 * their fields directly.
 */
```

timostamm (approved):
Looks good, just a comment nit:

TyG641 (approved):

PR #1389, opened 2026-04-22
Document UTF-8 validation in fromBinary with escape hatch in configureTextEncoding

emcfarlane:
Expand the Text encoding section of MANUAL with a subsection covering the UTF-8 validation added in #1386. Documents what fromBinary now rejects (vs. the prior silent U+FFFD replacement), that BinaryReader/BinaryWriter default behavior is unchanged, and a configureTextEncoding snippet that restores the lax behavior.


emcfarlane:
Moved documentation into PR comment on https://github.com/bufbuild/protobuf-es/pull/1386 so it will be more visible to users looking through PRs in the versions changelog.

Issue #333, opened 2022-12-06
Performance is not ideal


dae:
We currently use protobufjs, and would love to switch over to protobuf-es for the typescript ergonomics and es6/tree shaking that it offers. Parts of our app deal with large messages, so I did a quick performance test, and unfortunately protobuf-es is rather behind at the moment. The test I used was decoding a 6MB binary message, it containing repeated fields with about 150k entries. Protobufjs parses it in about 100ms; protobuf-es takes about 350ms. For reference, the same message as JSON is about 27MB, and JSON.parse() parses it in about 100ms as well. If providing the message/proto would be helpful, please let me know.

NfNitLoop:
Hi @dae  -- I'm not developer for this project, but I'm a user and just saw your issue. Providing a concrete use case is always helpful in bug reports so I just thought I'd chime in and say yeah, that's useful. :)   Especially if you can also provide minimal code to reproduce the performance issue.  Then it becomes a de facto test case so any fixes can be tested with before/after code.

felicio:
Adding Protobuf-ES to proton's benchmarking suite at https://github.com/ipfs/protons/pull/89, where it also performs the slowest.

dae:
I suspect any large message will reproduce the issue, but can produce a sample one if the devs request it.

On a positive note, protobuf-es's code generation seems to be a lot faster than what we currently have with protobufjs, where we have to produce a static version, which is then used to generate the ts, and then a separate json version needs to be produced because the static version's resulting code is huge. protobuf-es appears to be clearly sub-second, when protobufjs was taking around 5 seconds to rebuild each time a proto file was changed.

felicio:
Reference how `protons` dealt with perf issues at https://github.com/ipfs/protons/issues/51.

jcready:
Protons appears to use protobuf.js's UTF-8 decoder/encoder [which has correctness bugs](https://github.com/protobufjs/protobuf.js/issues/1473) and likely the reason protobuf-es uses the native TextEncoder/Decoder. If you're using the NodeJS runtime for your benchmarks you can comment on this bug to ask for a faster TextEncoder/Decoder: https://github.com/nodejs/node/issues/39879 as Bun and Deno appear to have much faster implementations. Chrome, Firefox, and Safari are also generally much faster using the native TextEncoder/Decoder vs. protobuf.js's implementation (especially once you get above 32 characters) https://github.com/timostamm/protobuf-ts/issues/184#issuecomment-1000420219

[...]

PR #1390, opened 2026-04-22
Inline binary tag parsing for performance 

emcfarlane:
This inlines the binary tag parsing for performance. A simple benchmarks on Node v22 show ~15-35% higher throughput on 1-byte tag loop (common case).

Follow up to https://github.com/bufbuild/protobuf-es/pull/1387

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -1252,6 +1252,12 @@ user.firstName; // "Homer"
 user.active; // true
 ```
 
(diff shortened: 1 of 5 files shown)


timostamm:
Thanks, @emcfarlane. Lets look at this along with other changes. See https://github.com/bufbuild/protobuf-es/issues/333#issuecomment-4305232949

emcfarlane:
Closing, small fix, can be picked up later!

PR #1392, opened 2026-04-22
Chore fix turbo license ignores

emcfarlane:
Fix license-header ignore to exclude `bun/conformance` and `deno/conformance` directories (they have their own) and fix cache inputs too. Related to this flake: https://github.com/bufbuild/protobuf-es/actions/runs/24792704111/job/72554213271

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -37,7 +37,12 @@
       "outputLogs": "new-only"
     },
     "//#license-header": {
-      "inputs": ["$TURBO_DEFAULT$", "!packages/**", "!bun/conformance/**", "!deno/conformance/**"],
+      "inputs": [
+        "$TURBO_DEFAULT$",
+        "!packages/**",
+        "!bun/conformance/**",
+        "!deno/conformance/**"
+      ],
       "outputLogs": "new-only"
     },
     "//#lint": {
--- a/turbo.json
+++ b/turbo.json
@@ -37,12 +37,7 @@
       "outputLogs": "new-only"
     },
     "//#license-header": {
-      "dependsOn": [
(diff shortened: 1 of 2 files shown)


timostamm:
I don't think so. 

`npx turbo run license-header` runs all `license-header` tasks, including `//#license-header` and `@bufbuild/bun-conformance#license-header`. The task already has a dependency on `generate` (see [turbo.json](https://github.com/bufbuild/protobuf-es/blob/859013cd4dfb49061da35bd279817dd4fa84c666/turbo.json#L23)).

The problem is that the root `license-header` task tries to modify files out of scope: bun/conformance has its own tasks to generate and add license-headers, and the root task shouldn't try to mess with that. It's already ignoring `packages/**` (see [package.json](https://github.com/bufbuild/protobuf-es/blob/859013cd4dfb49061da35bd279817dd4fa84c666/package.json#L26)). 

It should also ignore `bun/conformance/**` and `deno/conformance/**`. This ignore wasn't updated when deno and bun were added.

turbo.json
```diff
@@ -37,7 +37,7 @@
       "outputLogs": "new-only"
     },
     "//#license-header": {
-      "inputs": ["$TURBO_DEFAULT$", "!packages/**"],
+      "inputs": ["$TURBO_DEFAULT$", "!packages/**", "!bun/**", "!deno/**"],
```
timostamm: Good call.

But we do want this task to run on bun/example and deno/example, though. 

Those two are not part of the npm workspaces (they can't be for ... reasons), and don't have their own tasks, but still need the headers.

timostamm (approved):
I think we got it right 😄

Issue #581, opened 2023-10-04
Improve TypeScript compatibility tests


timostamm:
Currently, we are testing compatibility with older versions of TypeScript by compiling tests. We want to make sure that the generated code, the runtime library, and common use cases compile without issues with older versions of TypeScript. 

Because our tests import from Node.js built-in packages, we need `@types/node`, which requires different versions for different versions of TS. This introduces complexity (e.g. https://github.com/bufbuild/protobuf-es/pull/574#issuecomment-1747345724), but does not give us any extra confidence. 

The best way forward seems to be to add a single file to `packages/protobuf-test/src` with common use cases, which does not import from Node.js built-in modules and does not import from `@jest/globals` either. Then compile it with various versions of TS for confidence.

timostamm:
With TS 5.9+, the TypeScript compatibility tests are no longer sustainable in their current form.

The goal of the tests is to cover the out-of-the-box experience for users with the default compiler settings, and to catch any regressions across supported versions of TypeScript.

Until now, we've got by with a simple trick: We include from `packages/protobuf-test/src/`. The tests include a whole lot of code that creates, serialize and otherwise interacts with messages and other types. There are also tests exercising various aspects of the generated types, including JSON types. 

With TS 5.9, `noUncheckedIndexedAccess` and `exactOptionalPropertyTypes` are enabled by default. We've enabled `exactOptionalPropertyTypes` for all code in this repo in https://github.com/bufbuild/protobuf-es/pull/1371, but we have decided against enabling `noUncheckedIndexedAccess` for the code base because it doesn't improve the code quality. 

TS 5.9 also sets `types: []` by default, which means our test files (which import from Node.js built-in packages) no longer compile with the defaults.

This means that our code base is diverging from the compiler settings we need to test going forward. Seems like the original issue description still holds true:

> The best way forward seems to be to add a single file to packages/protobuf-test/src with common use cases, which does not import from Node.js built-in modules [...]. Then compile it with various versions of TS for confidence.

PR #1397, opened 2026-05-04
Add compat package for TypeScript 5.9.x

ajeetdsouza:
Add a compatibility package to check that `protobuf-es` works with TypeScript 5.9's default `tsconfig.json`.


packages/typescript-compat/v5.9.x/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "ts5.9",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "tsc --outDir dist"
```
timostamm, line 6: We need to trrack why we're calling bare `tsc` here, as opposed to `node_modules/.bin/tsc` in the other compat packages.

Adding a section to the README in https://github.com/bufbuild/protobuf-es/pull/1397/commits/7757710e621615abc4dd96fe46fc5de94d47a957

packages/typescript-compat/v5.9.x/tsconfig.json
```diff
@@ -0,0 +1,44 @@
+{
```
timostamm, line 1: Major gaps here. From the README on tsconfig.json:

> - We have to include the test sources.
> - We explicitly want to check libs.

Updating the file in https://github.com/bufbuild/protobuf-es/pull/1397/commits/8e215c514f035a2b0bae3aacbeee26a9f111079d.

Adding an explanation to the README in https://github.com/bufbuild/protobuf-es/pull/1397/commits/278fe010d73e4aa0fed521f074dbf779dd922684.

timostamm (commented):
Pushed up a few fixes, see comments below:

timostamm (approved):
With TS 5.9+, our current approach for the compat test isn't sustainable. See https://github.com/bufbuild/protobuf-es/issues/581#issuecomment-4497182480

Let's get this merged and address this problem separately.

kog853025-max (approved):

jhaynardresos09-beep (approved):

PR #1398, opened 2026-05-04
Upgrade to TypeScript 6.0

ajeetdsouza:
- Upgrade TypeScript dependency to 6.0.
- Add a compatibility package to check that `protobuf-es` works with TypeScript 6.0's default `tsconfig.json`.


packages/typescript-compat/v6.0.x/tsconfig.json
```diff
@@ -0,0 +1,44 @@
+{
```
timostamm, line 1: This needs the same updates as https://github.com/bufbuild/protobuf-es/pull/1397

packages/bundle-size/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "include": ["src/**/*.ts"],
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "rootDir": "./src",
```
timostamm: Why do we need this option? Same for the other additions.

packages/protobuf-example/tsconfig.json
```diff
@@ -2,9 +2,10 @@
   "files": ["src/add.ts", "src/list.ts"],
   "compilerOptions": {
     "target": "es2017",
-    "moduleResolution": "Node10",
+    "moduleResolution": "Bundler",
```
timostamm: Not wrong per se, but I'm a bit worried that if we switch to bundler now, and later adopt [tshy](https://github.com/isaacs/tshy) (which uses nodenext + extra steps), we're changing this twice.

I think our options are:
1. Adopt `tshy` right away.
2. Stick with node10 via the compat option.

We should cut a release soon because:
- The manual was replaced by a new docs site, and npm package now have dead links to the manual
- We haven't cut a trusted release yet

I believe that a switch to `tshy` would fit in well with this release. 

Let's discuss in person.

packages/bundle-size/tsconfig.json
ajeetdsouza: TypeScript 6.0 has a breaking change that defaults `rootDir` to the location of the `tsconfig.json` file, and requires it to be explicitly set otherwise. In our case, the `rootDir` is `./src` in most places.

packages/protobuf-example/tsconfig.json
ajeetdsouza: Reverted the bundler change, will follow up with tshy integration.

packages/typescript-compat/v6.0.x/tsconfig.json
timostamm: Addressed in https://github.com/bufbuild/protobuf-es/pull/1398/changes/f41537b36b86369cafd2fa3baefba8059819a647

packages/bundle-size/tsconfig.json
timostamm: Details in the release notes: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#rootdir-now-defaults-to-

timostamm (approved):

PR #1417, opened 2026-05-19
Docs: fix buf logo with dark mode

sudorandom:
The Buf logo doesn't look good on `protobufes.com` while using dark mode. This change makes a dark-mode-specific logo which is the same as the other except for the Buf font color being `#fff` instead of `#333`

Before:
Screenshot 2026-05-19 at 09 32 09

After:
Screenshot 2026-05-19 at 09 32 03

ref: https://starlight.astro.build/guides/customization/#light-and-dark-logo-variants


timostamm (approved):
Thanks :)

PR #1435, opened 2026-06-02
Migrate `protobuf` build to `tshy`

ajeetdsouza:
`tshy` eases a lot of the pain of building CJS + ESM modules. This PR migrates the `protobuf` package to build with `tshy`.

Note that `protoplugin` also builds CJS + ESM, but that migration will be done in a followup.

### DIFF
--- a/packages/protobuf/package.json
+++ b/packages/protobuf/package.json
@@ -40,7 +40,7 @@
     }
   },
   "devDependencies": {
-    "tshy": "^4.1.2",
+    "tshy": "^4.1.3",
     "upstream-protobuf": "*"
   },
   "files": [
--- a/packages/protobuf/package.json
+++ b/packages/protobuf/package.json
@@ -3,7 +3,12 @@
   "version": "2.12.0",
   "license": "(Apache-2.0 AND BSD-3-Clause)",
   "description": "Protocol Buffers for ECMAScript. The only JavaScript Protobuf library that is fully-compliant with Protobuf conformance tests.",
-  "keywords": ["protobuf", "schema", "typescript", "ecmascript"],
+  "keywords": [
+    "protobuf",
+    "schema",
+    "typescript",
+    "ecmascript"
+  ],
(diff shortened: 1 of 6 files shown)


ajeetdsouza:
The bundle size seems to have increased - tshy is not propagating `sideEffects` correctly. Filed https://github.com/isaacs/tshy/issues/115 for this.

packages/protobuf/package.json
```diff
@@ -26,44 +28,107 @@
...
+      "./package.json": "./package.json"
+    }
+  },
+  "devDependencies": {
+    "tshy": "^4.1.2",
+    "upstream-protobuf": "*"
+  },
+  "files": [
+    "dist/**"
+  ],
```
timostamm, line 48: Can we simplify to this?


```suggestion
  "files": [ "dist"],
```

packages/protobuf/package.json
```diff
@@ -26,44 +28,107 @@
   },
   "type": "module",
   "sideEffects": false,
-  "main": "./dist/cjs/index.js",
+  "tshy": {
```
timostamm, line 31: Can we remove `packages/protobuf/.npmignore`, and replace what needs to be excluded with the package.json option `tshy.exclude`? E.g.


```suggestion
  "tshy": {
    "exclude": [
      "scripts/*"
    ]
```

timostamm (commented):
Good catch on the sideEffects side effect. Diff looks good to me, just a couple of comments below.

FYI: When referring to an npm package, you need to use the scoped name, e.g. `@bufbuild/protobuf`.

packages/protobuf/package.json
ajeetdsouza: `tshy` rewrites the `package.json` and reformats the file in doing so, so we can't enforce any formatting. This is also why I am now excluding `package.json` from `biome.json`.

Alternate option: formatting with `biome` after every build.

timostamm (approved):

PR #1436, opened 2026-06-03
Add recursion limit to fromBinary and fromJson

timostamm:
Decoding does not currently limit recursion depth, so deeply-nested input can trivially exhaust the call stack.

This adds a `recursionLimit` option to both binary and ProtoJSON parsing:

```ts
  /**
   * The maximum depth of nested messages to parse. If a message nests deeper
   * than this limit, parsing fails with an error instead of exhausting the
   * call stack. Defaults to 100.
   */
  recursionLimit: number;
```

The default of 100 matches C++, Java, and Python's default.
When a message nests deeper than the limit, parsing fails with a catchable error before overflowing the stack:

```ts
try {
  fromBinary(schema, bytes);
} catch (e) {
  e; // cannot decode <type> from binary: maximum recursion depth of 100 reached
}
```

Note that this is a behavior change: input nested deeper than 100 levels used to decode and now throws. Code that relies on deeper nesting can raise the limit explicitly, e.g. `fromBinary(schema, bytes, { recursionLimit: 1000 })`.

The new option is made available on `getExtension` too, which didn't expose binary options so far.

### DIFF
--- a/packages/protobuf-test/src/json.test.ts
+++ b/packages/protobuf-test/src/json.test.ts
@@ -1294,6 +1294,76 @@ void suite("JSON parse errors", () => {
   }
 });
 
+void suite("fromJson recursion limit", () => {
+  // Nested google.protobuf.Any values.
+  function makeNestedAny(depth: number): JsonValue {
+    const root: Record<string, JsonValue> = {
+      "@type": "type.googleapis.com/google.protobuf.Any",
+    };
+    let cursor = root;
+    for (let i = 0; i < depth; i++) {
+      const next: Record<string, JsonValue> = {
+        "@type": "type.googleapis.com/google.protobuf.Any",
+      };
+      cursor.value = next;
+      cursor = next;
+    }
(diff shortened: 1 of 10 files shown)


ajeetdsouza (approved):

packages/protobuf-test/src/json.test.ts
```diff
@@ -1294,6 +1294,76 @@ void suite("JSON parse errors", () => {
...
+    for (let i = 0; i < depth; i++) {
+      message = { recursiveMessage: message };
+    }
+    return message;
+  }
+
+  test("rejects deeply nested Any instead of overflowing the stack", () => {
+    const registry = createRegistry(AnySchema);
+    assert.throws(
+      () => fromJson(AnySchema, makeNestedAny(1000), { registry }),
```
ajeetdsouza, line 1325: I would add a test on the boundary of the recursion limit to check for off-by-one errors.

PR #1441, opened 2026-06-18
Apply recursion limit to BinaryReader.skip

timostamm:
Following up on https://github.com/bufbuild/protobuf-es/pull/1436, this applies the recursion limit to `BinaryReader`'s method `skip`.

### DIFF
--- a/packages/protobuf-test/src/wire/binary-encoding.test.ts
+++ b/packages/protobuf-test/src/wire/binary-encoding.test.ts
@@ -239,6 +239,42 @@ void suite("BinaryReader", () => {
         assert.strictEqual(sr.pos, sr.len);
       }
     });
+    void test("rejects deeply nested groups", () => {
+      // A run of StartGroup tags (0x0b = field 1, WireType.StartGroup) with no
+      // matching EndGroup tags. Without a recursion limit, skipping this
+      // overflows the call stack.
+      const startGroupTags = new Uint8Array(10000).fill(0x0b);
+      const reader = new BinaryReader(startGroupTags);
+      const [, wireType] = reader.tag();
+      assert.strictEqual(wireType, WireType.StartGroup);
+      assert.throws(
+        () => reader.skip(wireType),
+        /maximum recursion depth reached/,
+      );
+    });
+    void test("honors a custom recursionLimit when skipping groups", () => {
+      // `depth` nested groups, each closed by a matching EndGroup.
+      function nestedGroups(depth: number): Uint8Array {
+        const writer = new BinaryWriter();
+        for (let i = 0; i < depth; i++) {
+          writer.tag(1, WireType.StartGroup);
+        }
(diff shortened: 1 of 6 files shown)


packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -431,6 +433,9 @@ export class BinaryReader {
         this.pos += len;
         break;
       case WireType.StartGroup:
+        if (recursionLimit <= 0) {
+          throw new Error("maximum recursion depth reached");
+        }
```
ajeetdsouza, line 438: Unless this is a perf optimization, we should guard `recursionLimit` right at the start of the function for clarity.

packages/protobuf-test/src/wire/binary-encoding.test.ts
```diff
@@ -239,6 +239,42 @@ void suite("BinaryReader", () => {
         assert.strictEqual(sr.pos, sr.len);
       }
     });
+    void test("rejects deeply nested groups", () => {
+      // A run of StartGroup tags (0x0b = field 1, WireType.StartGroup) with no
+      // matching EndGroup tags. Without a recursion limit, skipping this
+      // overflows the call stack.
+      const startGroupTags = new Uint8Array(10000).fill(0x0b);
```
ajeetdsouza, line 246:
```suggestion
      const startGroupTag = (1 << 3) | WireType.StartGroup;
      const startGroupTags = new Uint8Array(10000).fill(startGroupTag);
```

packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -410,9 +410,11 @@ export class BinaryReader {
    * Skip one element and return the skipped data.
    *
    * When skipping StartGroup, provide the tags field number to check for
-   * matching field number in the EndGroup tag.
+   * matching field number in the EndGroup tag. Recursion into nested groups
+   * is guarded by the `recursionLimit` argument: When the limit is reached,
+   * this method throws.
    */
-  skip(wireType: WireType, fieldNo?: number): Uint8Array {
+  skip(wireType: WireType, fieldNo?: number, recursionLimit = 100): Uint8Array {
```
ajeetdsouza, line 417: Should `100` be a constant here? It's used in two places now.

ajeetdsouza (approved):

packages/protobuf/src/wire/binary-encoding.ts
timostamm: No, it's more important that the reader stays self-contained, because it's used by ts-proto.

packages/protobuf-test/src/wire/binary-encoding.test.ts
timostamm: See the comment on line 243

PR #1450, opened 2026-06-23
Add support for the Protobuf Text Format

timostamm:
The [Text Format](https://protobuf.dev/reference/protobuf/textformat-spec/) is a syntax for representing protobuf data in text form. It's mainly used for debugging, tests, and config files. To support those use cases, this PR adds support for the format. You can use it to read and write `.txtpb` files.

Like ProtoJSON and the binary wire format, the format is exposed via free functions, exported from `@bufbuild/protobuf/txtpb`:

- `toText()` - Serialize a message to the protobuf text format.
- `fromText()` - Parse a message from the protobuf text format.
- `mergeFromText()` - Parse into an existing message.

Here's an example using our [example message](https://github.com/bufbuild/protobuf-es/blob/v2.12.1/packages/protobuf-example/proto/example.proto#L18):

```ts
import { toText } from "@bufbuild/protobuf/txtpb";

const user: User = create(UserSchema, {
  firstName: "Alice",
  lastName: "Smith",
  active: true,
  locations: ["NYC", "LDN"],
  projects: { atlas: "infra" },
});

console.log(toText(UserSchema, user));
```

This outputs:

```txtpb
first_name: "Alice"
last_name: "Smith"
active: true
locations: "NYC"
locations: "LDN"
projects: {
  key: "atlas"
  value: "infra"
}
```

The implementation is built to match other Protobuf implementations: output follows the default formatting of [txtpbfmt](https://github.com/protocolbuffers/txtpbfmt), and the format passes the [upstream conformance suite](https://github.com/protocolbuffers/protobuf/tree/main/conformance). Where the [text-format spec](https://protobuf.dev/reference/protobuf/textformat-spec/) leaves details open, behavior was matched against protobuf-go's [prototext](https://pkg.go.dev/google.golang.org/protobuf/encoding/prototext) package.

As a result, text written by `toText()` can be read by the buf CLI, `protoc`, and other implementations, and text they produce can be read by `fromText()`.


packages/protobuf/src/txtpb/is-group-like.ts
```diff
@@ -0,0 +1,49 @@
...
+  // Groups are always delimited-encoded message fields.
+  if (field.message === undefined || !field.delimitedEncoding) {
+    return false;
+  }
+  // Group fields are always named after the lowercase message type name.
+  if (field.message.name.toLowerCase() !== field.name) {
+    return false;
+  }
+  // Groups can only be defined in the file they are used in.
+  if (field.message.file !== field.parent.file) {
```
ajeetdsouza, line 44: Curious: is there a case covered by this that isn't covered by the check below?

```js
return field.message.parent === field.parent;
```

packages/protobuf/src/txtpb/writer.ts
```diff
@@ -0,0 +1,239 @@
...
+ * extra spaces that protobuf-go adds to discourage parsing its output as
+ * canonical.
+ *
+ * Output accumulates into a single buffer of lines, so a deep tree costs no
+ * more than the bytes it prints. To decide between `name: {}` and an indented
+ * block, the caller renders the body, then rolls back with mark()/reset() if it
+ * turned out empty — an O(1) decision that needs no separate child buffer.
+ */
+export class Writer {
+  private readonly chunks: string[] = [];
```
ajeetdsouza, line 42: nit: `lines` might be a better name here.

ajeetdsouza (approved):

PR #1451, opened 2026-06-29
Special-case prototype keys in maps

ajeetdsouza:
Under certain conditions, map accesses can fall back to prototype fields and methods. This PR fixes map access to prevent this from happening.


ajeetdsouza:
This change is unlikely to affect real users, so I don't think it warrants a special case in the code. We can revisit it if that situation ever arises.

Issue #1216, opened 2025-08-25
Improve type representation for open enums


timostamm:
Protobuf has [closed and open enums](https://protobuf.dev/programming-guides/enum). A closed enum (proto2) only accepts declared values, while an open enum (proto3) also accepts other values.  `protoc-gen-es` generates TypeScript enums, which are always [closed since version 5.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#enum-overhaul). 

This means that generated types do not match the semantics for open Protobuf enums. There is no indication that an enum field could contain a value that isn't one of the TypeScript enum values, and additional checks might even be flagged as unnecessary or incorrect by the compiler. 

We should investigate how to improve the generated enum types or fields to properly reflect open enum semantics.

Context:
- [protobuf.dev on Enum behavior](https://protobuf.dev/programming-guides/enum/)
- https://github.com/bufbuild/protobuf-es/issues/1082
- https://github.com/bufbuild/protobuf-es/issues/522
- https://github.com/bufbuild/protobuf-es/issues/1139

PR #1462, opened 2026-07-09
Add TS-compat tests for enum closed/open behavior

timostamm:
This documents the TypeScript behavior across different versions. For context, see https://github.com/bufbuild/protobuf-es/issues/1216


ajeetdsouza (commented):
Hm, this is just testing the language - do we need to check this into the repo?

I know we're doing this for our enums change, but can't we write ephemeral tests to verify behavior on different TS versions and then discard them once we have figured out how it works?

jhaynardresos09-beep (approved):

timostamm:
> Hm, this is just testing the language - do we need to check this into the repo?

Yes, same as with the Uint8Array change. The lever for simplification is dropping old TS versions.

ajeetdsouza (approved):

PR #1463, opened 2026-07-09
Add helper for unknown enum values

timostamm:
This adds the function `isUnkownEnum()`, which narrows an enum value to `UnknownEnum` when it's not in the set of values defined by the enum.

For example, consider:

```protobuf
syntax = "proto3";
enum Color {
  COLOR_UNSPECIFIED = 0;
  COLOR_RED = 1;
  COLOR_GREEN = 2;
}
```

It generates the following enum in TypeScript:

```typescript
export enum Color {
  UNSPECIFIED = 0,
  RED = 1,
  GREEN = 2
}
```

When someone adds `COLOR_BLUE = 3` and sends a message with this value to an older peer, the field will contain `3` - a value that's not in the TypeScript enum.

The helper function helps identifying such values:

```typescript
import { isUnknownEnum } from "@bufbuild/protobuf";

function checkColor(color: Color) {
  if (isUnknownEnum(color)) {
    console.log(`unknown color: ${color}`); // e.g. 3
  }
  console.log(Color[color]); // e.g. "RED"
}
```

### DIFF
--- a/packages/protobuf/src/types.ts
+++ b/packages/protobuf/src/types.ts
@@ -139,6 +139,12 @@ export type UnknownField = {
 };
 
 /**
+ * An unknown enum is a value that's not in the set of values defined by an
+ * enum. Open Protobuf enums may
+ */
+export type UnknownEnum = number & { __unknown_enum: true };
+
+/**
  * Describes a streaming RPC declaration.
  */
 export type DescMethodStreaming<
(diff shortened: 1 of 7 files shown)


packages/protobuf/src/types.ts
```diff
@@ -138,6 +138,12 @@ export type UnknownField = {
   readonly data: Uint8Array;
 };
 
+/**
+ * An unknown enum is a value that's not in the set of values defined by an
+ * enum. Open Protobuf enums may
```
ajeetdsouza, line 143: This sentence looks truncated

ajeetdsouza (approved):

Issue #1139, opened 2025-05-29
Feature Request: `enumAsConst` option


azu:
My codebase has the [erasableSyntaxOnly](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly) option enabled, so `enum` is not available.

Since `protoc-gen-es` generates code using TypeScript `enum`, it generates code that is not available.

For example, GraphQL Codegen has an option to generate code that uses `as const` instead of `enum`, such as `enumAsConst`

- https://the-guild.dev/graphql/codegen/plugins/typescript/typescript#enumsasconst

There are some environments (like [Node.js](https://nodejs.org/api/typescript.html#type-stripping)) where TypeScript's `enum` syntax is not available, so I would like an option to generate code that is interoperable.

## Proto

`messages.proto`:
```proto
syntax = "proto3";

package com.example.enum;

enum Enum {
  A = 0;
  B = 1;
}

message Msg {
  optional Enum enum = 1;
}
```

`buf.gen.yaml`:
```yaml
# buf.gen.yaml
version: v2
plugins:
  - remote: buf.build/bufbuild/es:v2.3.0
    out: web/src/proto_generated 
    opt:
      - target=ts
      - json_types=true
      - ts_nocheck=true
      - import_extension=.ts
```

## Actual

`buf generate` result code
```ts
// @generated by protoc-gen-es v2.3.0 with parameter "target=ts,json_types=true,ts_nocheck=true,import_extension=.ts"
// @generated from file proto/messages.proto (package com.example.enum, syntax proto3)
/* eslint-disable */
// @ts-nocheck

import type { GenEnum, GenFile, GenMessage } from "@bufbuild/protobuf/codegenv1";
import { enumDesc, fileDesc, messageDesc } from "@bufbuild/protobuf/codegenv1";
import type { Message } from "@bufbuild/protobuf";

/**
 * Describes the file proto/messages.proto.
 */
export const file_proto_messages: GenFile = /*@__PURE__*/
  fileDesc("ChRwcm90by9tZXNzYWdlcy5wcm90bxIQY29tLmV4YW1wbGUuZW51bSI5CgNNc2cSKQoEZW51bRgBIAEoDjIWLmNvbS5leGFtcGxlLmVudW0uRW51bUgAiAEBQgcKBV9lbnVtKhQKBEVudW0SBQoBQRAAEgUKAUIQAWIGcHJvdG8z");

/**
 * @generated from message com.example.enum.Msg
 */
[...]

PR #1464, opened 2026-07-09
Support enums with erasable syntax

### LINKED ISSUE #1216: Improve type representation for open enums

timostamm:
This PR adds the plugin option `erasable_syntax=true`, which changes how enums are generated.

Instead of generating a TypeScript enum, `@bufbuild/protoc-gen-es` generates an object with `as const` and a union type for the enum values:

```typescript
export const PhoneType = {
  UNSPECIFIED: 0,
  MOBILE: 1,
  LAND_LINE: 2,
} as const;

export type PhoneType = (typeof PhoneType)[keyof typeof PhoneType] | UnknownEnum;
```

Open enums can contain numeric values that are not in the set of values defined by the enum, so the type for an open enum is a union with `UnknownEnum` (and can be narrowed with [isUnknownEnum](https://github.com/bufbuild/protobuf-es/pull/1463)).

The object does not support mapping from numeric value to string, but it is compatible with the `tsconfig` option [erasableSyntaxOnly](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly), and allows [running TypeScript natively in Node.js](https://nodejs.org/learn/typescript/run-natively).

Note that this option is experimental for now to let us gathered feedback from the community.

Updates #1139, fixes #1216.


JosephusPaye:
Keen for this!

jhaynardresos09-beep (approved):

packages/protoc-gen-es/biome.json
```diff
@@ -3,5 +3,11 @@
   "extends": ["../../biome.base.json"],
   "files": {
     "ignore": ["dist"]
-  }
+  },
+  "overrides": [
+    {
+      "include": ["src/protoc-gen-es-plugin.ts"],
```
ajeetdsouza, line 9: The function-level ignore seems a lot better - there is other stuff in that file that we can still rely on biome to format for us.

ajeetdsouza (approved):

kutsan:
Thanks a lot! When should we expect the release that includes this change?

timostamm:
@kutsan, as soon as https://github.com/bufbuild/protobuf-es/pull/1469 merges, the feature will be available in v2.13.0.

PR #1473, opened 2026-07-20
Format protos

ajeetdsouza:
Ran `buf format -w` over our proto files.


packages/protobuf-test/extra/option-usage.proto
```diff
@@ -17,9 +17,9 @@ package spec;
 
 import "extra/options.proto";
 
-option (spec.file_option_retention_unknown) = "file option retention unknown";
 option (spec.file_option_retention_runtime) = "file option retention runtime";
 option (spec.file_option_retention_source) = "file option retention source";
+option (spec.file_option_retention_unknown) = "file option retention unknown";
```
timostamm: A bit strange that the order changes here, but not for the options below. Are you sure this isn't changing semantics anywhere (repeated extensions)? I'd much prefer to have a consistent order. Can you rename 
- `_option_retention_unknown` -> `_option_retention_default` and 
- `" option retention unknown"` -> `" option retention default"`?

timostamm (approved):

PR #1474, opened 2026-07-23
Add benchmarks

ajeetdsouza:
Adds benchmarks for `fromJson`, `toJson`, `fromBinary`, and `toBinary`. Also covers microbenchmarks for a variety of protobuf-es functions.


packages/protobuf-bench/.gitignore
```diff
@@ -0,0 +1 @@
+.cpuprof
```
timostamm: Move this to the root .gitignore

packages/protobuf-bench/biome.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
+  "extends": ["../../biome.base.json"],
+  "files": {
+    "ignore": ["dist", ".0x", "src/gen"]
```
timostamm: 0x is gone, so the ignore can go as well.

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@bufbuild/protobuf-bench",
+  "version": "0.0.0",
```
timostamm: Just omit the version field.

packages/protobuf-bench/README.md
```diff
@@ -0,0 +1,80 @@
+# @bufbuild/protobuf-bench
```
timostamm:
```suggestion
# Benchmarks for @bufbuild/protobuf
```

packages/protobuf-bench/turbo.json
```diff
@@ -2,18 +2,18 @@
...
     },
     "profile": {
-      "dependsOn": ["^build", "generate"],
+      "dependsOn": ["^build"],
       "cache": false,
       "outputLogs": "new-only"
     },
     "flamegraph": {
-      "dependsOn": ["^build", "generate"],
+      "dependsOn": ["^build"],
```
timostamm, line 16: Flamegraph is gone, right?

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "private": true,
+  "type": "module",
+  "description": "Benchmarks for @bufbuild/protobuf",
+  "scripts": {
+    "build": "tsc --noEmit",
```
timostamm: > WARNING  no output files found for task @bufbuild/protobuf-bench#build. Please check your `outputs` key in `turbo.json`

Please add an override for the task.


Also: Doesn't that need to use the root tsc? See protobuf-test.

packages/protobuf-bench/README.md
```diff
@@ -0,0 +1,80 @@
...
+```sh
+npm run bench                      # the whole corpus
+npm run bench -- '^BinaryWriter/'  # only the BinaryWriter primitives
+npm run bench -- '^toBinary/'      # only toBinary, across all fixtures
+npm run bench -- '/general$'       # all four ops on the general fixture
+npm run bench -- int32             # every case whose name contains "int32"
+```
+
+Arguments are regular expressions matched against case names (unanchored, OR'd
+together; anchor with `^…$` for an exact match). It prints a per-operation
```
timostamm:
```suggestion
It prints a per-operation
```

Let's add `--help` to the CLI instead - bench and profile - see https://github.com/bufbuild/protovalidate-es/blob/47b671fe6d9764753cf343dee4e1af257768e38a/packages/protovalidate-bench/src/bench.ts#L22

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "postgenerate": "license-header src/gen",
+    "bench": "tsx src/bench.ts",
+    "profile": "node --cpu-prof --cpu-prof-dir .cpuprof --import tsx src/profile.ts",
+    "format": "biome format --write",
+    "license-header": "license-header",
+    "lint": "biome lint --error-on-warnings"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@bufbuild/protobuf": "2.12.1",
```
timostamm: We're at 2.13.0 - need to update.

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "tinybench": "^3.1.1"
+  },
+  "devDependencies": {
+    "@bufbuild/buf": "^1.66.1",
+    "@bufbuild/protoc-gen-es": "2.12.1",
```
timostamm: See above

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "license-header": "license-header",
+    "lint": "biome lint --error-on-warnings"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@bufbuild/protobuf": "2.12.1",
+    "tinybench": "^3.1.1"
+  },
+  "devDependencies": {
+    "@bufbuild/buf": "^1.66.1",
```
timostamm: Is this the same version used elsewhere? Same Q for tsx.

packages/protobuf-bench/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@bufbuild/protoc-gen-es": "2.12.1",
+    "@types/node": "^26.0.0",
+    "tsx": "^4.22.4",
+    "typescript": "^6.0.3"
```
timostamm: Drop typescript here, see protobuf-test for example.

packages/protobuf-bench/src/corpus.ts
```diff
@@ -0,0 +1,433 @@
...
+  GeneralSchema,
+  MapMessageSchema,
+  MapScalarSchema,
+  RepeatedMessageSchema,
+  RepeatedScalarSchema,
+  ScalarSchema,
+  UserSchema,
+} from "./gen/bench/v1/bench_pb.js";
+
+export type Run = () => unknown;
```
timostamm: Just inline this in Case.run please

packages/protobuf-bench/src/fixtures.ts
```diff
@@ -0,0 +1,239 @@
...
+  let a = seed >>> 0;
+  const MAX_UINT32 = 2 ** 32; // normalize the uint32 result down to [0, 1)
+  return () => {
+    a = (a + 0x6d2b79f5) | 0;
+    let t = Math.imul(a ^ (a >>> 15), 1 | a);
+    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
+    return ((t ^ (t >>> 14)) >>> 0) / MAX_UINT32;
+  };
+}
+const rand = mulberry32(SEED);
```
timostamm: We need the ability to compare perf improvements against older runs, right? The way you're setting up the PRNG once, any added fixture or change in fixture order will change the payloads, and runs aren't comparable anymore. I believe we need to support that. Can you fix it?

timostamm (changes requested):

timostamm (approved):

PR #1477, opened 2026-07-24
Use native toBase64 / fromBase64 when available

ajeetdsouza:
Leverages [fromBase64](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) and [toBase64](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64) where possible. Improves performance of binary encoding / decoding in JSON.

| case                            | main     | this PR  | Delta time |
|----------------------------------|----------|----------|------------|
| toJson/general                  | 146.0 us | 125.6 us | -13.8%     |
| fromJson/general                | 238.5 us | 221.3 us | -7.3%      |
| toJson/scalar                   | 904 ns   | 842 ns   | -7.2%      |
| fromJson/scalar                 | 2554 ns  | 2505 ns  | -1.8%      |


timostamm (commented):
Is the API is available yet on Node.js? We need tests to cover the fallback.

timostamm:
Node.js supports this in v25 and v26. Let's add v26 to the CI matrix (and drop v20). This gives us coverage for both code paths.

timostamm:
https://github.com/bufbuild/protobuf-es/pull/1477/commits/3fc9d03b2b5a9d346caf123196cde179cd7abe09 is technically half-way there, but this needs to happen in a separate PR: https://github.com/bufbuild/protobuf-es/pull/1488

See https://github.com/bufbuild/protobuf-es/pull/1201 for the previous bump.

timostamm (approved):

PR #1479, opened 2026-07-24
Use native String.prototype.isWellFormed() when available

### LINKED ISSUE #1243: Use String.prototype.isWellFormed() once it's widely available
Protobuf requires strings to be valid UTF-8. When serializing, we check strings [via encodeUriComponent](https://github.com/bufbuild/protobuf-es/blob/v2.9.0/packages/protobuf/src/wire/text-encoding.ts#L62), which is far from ideal for performance. 

[String.prototype.isWellFormed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed) is a suitable alternative. On Node.js, it shows significantly better performance, especially for longer strings:

```bash
$ node --version
v24.5.0
$ node ./t.ts encodeURIComponent 10
node ./t.ts isWellFormed 10
node ./t.ts encodeURIComponent 100
node ./t.ts isWellFormed 100
node ./t.ts encodeURIComponent 1000
node ./t.ts isWellFormed 1000

encodeURIComponent with string length 10: 77.23291699999999 ms
isWellFormed with string length 10: 16.621958 ms
encodeURIComponent with string length 100: 34.113417 ms
isWellFormed with string length 100: 3.5224170000000044 ms
encodeURIComponent with string length 1000: 29.431917 ms
isWellFormed with string length 1000: 0.5034169999999989 ms
```

<details><summary>Benchmark script</summary>

```ts
// t.ts
const type = process.argv[2];
let checkUtf8: (str: string) => boolean;
switch (type) {
  case "encodeURIComponent":
    checkUtf8 = function checkUtf8(str: string) {
      try {
        encodeURIComponent(str);
        return true;
      } catch (_) {
        return false;
      }
    };
    break;
  case "isWellFormed":
    checkUtf8 = function checkUtf8(str: string) {
      // @ts-expect-error
      return str.isWellFormed();
    };
    break;
  default:
    throw new Error("Unknown type: " + type);
}

const strLen = process.argv[3];
let strings: string[];
switch (strLen) {
  case "10":
    strings = new Array(1_000_000).fill("012345678¼");
    break;
  case "100":
    strings = new Array(100_000).fill("012345678¼".repeat(10));
    break;
  case "1000":
    strings = new Array(10_000).fill("012345678¼".repeat(100));
    break;
  default:
    throw new Error("Unknown strLen: " + strLen);
}

const start = performance.now();
for (const str of strings) {
  if (!checkUtf8(str) ) {
    throw new Error(`Unexpected invalid utf-8 ${str}`);
  }
}
const elapsed = performance.now() - start;
console.log(`${type} with string length ${strLen}: ${elapsed} ms`);
```
</details>

`isWellFormed` is not widely available yet, but it will be in April 2026. See the [definition of "widely available" on MDN](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility).

Related to: https://github.com/bufbuild/protobuf-es/issues/333

ajeetdsouza:
Leverages https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed where possible. Improves performance of UTF-8 validation, which runs on every string field, key, and value when parsing JSON.

The previous check called encodeURIComponent() and caught the URIError thrown on lone surrogates. This is correct, but it percent-encodes the whole string and allocates the result just to discard it. isWellFormed() returns the same answer. Falls back to encodeURIComponent() on runtimes without it (pre-ES2024).

| Task                      | before (ns/op) | after (ns/op) |      Δ |
|---------------------------|----------------|---------------|--------|
| fromJson/map-scalar       |          7,220 |         3,626 | −49.8% |
| fromJson/repeated-scalar  |          4,968 |         3,809 | −23.3% |
| fromJson/user-normal      |        1,817.9 |       1,424.5 | −21.6% |
| fromJson/general          |        262,555 |       228,461 | −13.0% |
| fromJson/map-message      |      2,873,944 |     2,738,440 |  −4.7% |
| fromJson/repeated-message |      2,601,457 |     2,493,185 |  −4.2% |
| fromJson/scalar           |        2,584.2 |       2,487.2 |  −3.8% |
| fromJson/user-tiny        |          605.4 |         607.1 |  +0.3% |

The gain scales with how string-heavy the message is (map keys and repeated strings validate the most); user-tiny has no string content, so it's flat within noise. 

Fixes https://github.com/bufbuild/protobuf-es/issues/1243


timostamm (commented):
It would great if we could simply drop the kludge, and rely on isWellFormed alone.
We dropped support for Node.js 18 last year. Is the fallback still necessary for other runtimes we still support?

timostamm:
> It would great if we could simply drop the kludge, and rely on isWellFormed alone. We dropped support for Node.js 18 last year. Is the fallback still necessary for other runtimes we still support?

The API is widely supported in web browsers. But only since a couple of months. Let's keep the fallback, and drop it on a separate occasion.

timostamm (approved):

PR #1482, opened 2026-07-27
Extract ByteBuffer and use it in BinaryWriter

ajeetdsouza:
The growable buffer that BinaryWriter kept inline (buffer, pos,
ensureCapacity) is the same thing sizeDelimitedDecodeStream already had
as a private ByteBuffer class. Move that class into its own module, give
it a reserve() so callers can write directly into the backing buffer,
and have BinaryWriter hold one.

This is a refactor. Encoded output is unchanged, verified byte-identical
over 174 writer cases covering every scalar writer, payloads straddling
the capacity and varint-length boundaries, nested forks, and writer
reuse. finish() still returns a copy, and reserve() keeps the writer's
existing growth policy: start at 128 bytes, double until the required
size fits.

One incidental change worth noting: fixed64() now uses setUint32 rather
than setInt32 for its two halves. Both write identical bytes; setUint32
just matches the field type.

### DIFF
--- a/packages/bundle-size/README.md
+++ b/packages/bundle-size/README.md
@@ -16,11 +16,11 @@ usually do. We repeat this for an increasing number of files.
 
 | code generator      | files | bundle size |  minified | compressed |
 | ------------------- | ----: | ----------: | --------: | ---------: |
-| Protobuf-ES         |     1 |   136,544 b |  70,588 b |   16,261 b |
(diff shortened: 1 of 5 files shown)


timostamm:
I don't believe it's worth DRYing this up, and IIRC correctly, you came to the same conclusion. Closing.

PR #1490, opened 2026-07-30
Reserve bytes for lenPrefix in strings / fork-join

ajeetdsouza:
In both strings and fork-join, we prefix the payload with its length as a varint. However, we don't know the length upfront, so we don't know how much space to reserve for the varint. Instead, we can make a guess:

- Strings: most strings are ASCII, so we can assume a length = number of bytes
- Fork-join: we're going with a length of 1, which works with payloads of size < 128 bytes

For strings, this also means we can use a faster API (`encodeInto` instead of `encode`). This avoids an allocation in all cases, and a copy in the happy path. Benchmarks are happy:

case | main (ops/s) | branch (ops/s) | change
-- | -- | -- | --
BinaryWriter/string-ascii | 6,871,110 | 19,021,976 | +176.8%
BinaryWriter/string-utf8 | 6,061,158 | 13,874,788 | +128.9%
toBinary/map-scalar | 60,388 | 108,101 | +79.0%
toBinary/user-normal | 360,969 | 630,710 | +74.7%
BinaryWriter/fork-join | 33,719,181 | 48,604,550 | +44.1%
toBinary/general | 5,276 | 7,088 | +34.3%
toBinary/repeated-scalar | 158,816 | 205,442 | +29.4%
toBinary/map-message | 759 | 878 | +15.7%
toBinary/repeated-message | 1,009 | 1,115 | +10.5%
toBinary/user-tiny | 2,270,830 | 2,316,987 | +2.0%

### DIFF
--- a/packages/protobuf/src/wire/text-encoding.ts
+++ b/packages/protobuf/src/wire/text-encoding.ts
@@ -52,8 +52,24 @@ interface TextEncoding {
  * Note that the Text Encoding API does not provide a way to validate UTF-8.
  * Our implementation falls back to use encodeURIComponent().
  */
-export function configureTextEncoding(textEncoding: TextEncoding): void {
-  (globalThis as GlobalWithTextEncoding)[symbol] = textEncoding;
+export function configureTextEncoding(
+  textEncoding: Partial<TextEncoding>,
+): void {
+  const current = getTextEncoding();
+  const merged: TextEncoding = {
+    checkUtf8: textEncoding.checkUtf8 ?? current.checkUtf8,
+    decodeUtf8: textEncoding.decodeUtf8 ?? current.decodeUtf8,
+    encodeUtf8: textEncoding.encodeUtf8 ?? current.encodeUtf8,
+  };
+  // encodeUtf8Into must encode identically to encodeUtf8, so only carry over
+  // the previous encodeUtf8Into if encodeUtf8 wasn't also overridden.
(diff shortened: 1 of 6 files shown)


packages/protobuf/src/wire/text-encoding.ts
```diff
@@ -28,6 +28,13 @@ interface TextEncoding {
    * Encode UTF-8 text to binary.
    */
   encodeUtf8: (text: string) => Uint8Array<ArrayBuffer>;
+  /**
+   * Encode UTF-8 text to a Uint8Array.
+   */
+  encodeUtf8Into?: (
```
timostamm: This API is becoming very awkward. I believe it should be possible to clean it up by letting configureTextEncoding() accept a partial, and getTextEncoding() returning the full object.

timostamm (commented):
The approach looks good to me, but I think it's time to give our text encoding facade some attention, see comment below.

timostamm:
Pushed up https://github.com/bufbuild/protobuf-es/pull/1490/commits/c896131fced734512fae2fcba9ca4ac16d7046f0 to fix stale docs and some other minor bits, see the commit message for details.

One thing that's still missing before we can land this: This PR adds `encodeIntoUtf8`, but there are not tests. See text-encoding.test.ts for the existing tests.

timostamm (approved):

PR #1491, opened 2026-07-31
Compile descriptors to closure trees

ajeetdsouza:
The overhead of interpreting descriptors in codecs is substantial. This PR "compiles" descriptors at runtime to a tree of closures, which reduces that overhead to zero. This reduces the minified bundle size by 8%, and comes with substantial gains in performance:

| Case              | fromBinary | toBinary | fromJson | toJson |
|-------------------|------------|----------|----------|--------|
| general           | +37%       | +44%     | +15%     | +76%   |
| scalar            | +52%       | +43%     | +36%     | +54%   |
| repeated-scalar   | +39%       | +54%     | +15%     | +209%  |
| map-scalar        | +19%       | +13%     | +19%     | +153%  |
| repeated-message  | +56%       | +63%     | +34%     | +53%   |
| map-message       | +49%       | +58%     | +38%     | +64%   |
| user-tiny         | +84%       | +36%     | +51%     | +63%   |
| user-normal       | +37%       | +24%     | +56%     | +157%  |

### DIFF
--- a/packages/protobuf/src/from-binary.ts
+++ b/packages/protobuf/src/from-binary.ts
@@ -13,15 +13,14 @@
 // limitations under the License.
 
 import { type DescField, type DescMessage, ScalarType } from "./descriptors.js";
-import type { MessageShape } from "./types.js";
-import type {
-  ReflectList,
-  ReflectMap,
-  ReflectMessage,
-} from "./reflect/index.js";
+import type { Message, MessageShape, UnknownField } from "./types.js";
 import { scalarZeroValue } from "./reflect/scalar.js";
-import type { ScalarValue } from "./reflect/scalar.js";
-import { reflect } from "./reflect/reflect.js";
+import type { ReflectMessage } from "./reflect/index.js";
+import { FieldError } from "./reflect/error.js";
+import { unsafeLocal } from "./reflect/unsafe.js";
+import { localMessageMapper } from "./reflect/message.js";
+import { create } from "./create.js";
+import { protoInt64 } from "./proto-int64.js";
 import { BinaryReader, WireType } from "./wire/binary-encoding.js";
 import { varint32write } from "./wire/varint.js";
 
@@ -67,15 +66,14 @@ export function fromBinary<Desc extends DescMessage>(
   bytes: Uint8Array,
   options?: Partial<BinaryReadOptions>,
 ): MessageShape<Desc> {
-  const msg = reflect(schema, undefined, false);
-  readMessage(
-    msg,
+  const message = create(schema);
+  compiledReader(schema).read(
+    message as Record<string, unknown>,
     new BinaryReader(bytes),
     makeReadContext(options),
-    false,
     bytes.byteLength,
   );
-  return msg.message as MessageShape<Desc>;
+  return message;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/protobuf/src/wkt/json.ts
```diff
@@ -0,0 +1,33 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Minimum google.protobuf.Timestamp in JSON, in milliseconds (inclusive).
+ */
+export const timestampMsMin = /*@__PURE__*/ Date.parse("0001-01-01T00:00:00Z");
```
timostamm, line 21: Good call to DRY those up. IMO worth being very clear on where they are enforced, see prose suggestions. Please add a `@private` JSDoc tag (so that it's obvious it's intentionally not part of the public API surface). 

```suggestion
/**
 * Minimum google.protobuf.Timestamp in milliseconds (inclusive).
 * Only enforced in ProtoJSON.
 *
 * @private
 */
export const timestampMsMin = /*@__PURE__*/ Date.parse("0001-01-01T00:00:00Z");
```

timostamm (commented):
That was a fair bit of work, good job. Haven't reviewed completely, but want to leave one minor comment, see below.

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +275,360 @@ export function readField(
...
-      return reader.int32();
+      return (reader) => reader.int32();
     case ScalarType.INT64:
-      return reader.int64();
+      if (longAsString) {
+        return (reader) => String(reader.int64());
+      }
+      return protoInt64.supported
+        ? (reader) => reader.int64()
+        : (reader) => protoInt64.parse(reader.int64());
```
timostamm: I don't think the alternative branch is needed (same for the 4 other 64-bit integer types):

```suggestion
      if (longAsString) {
        return (reader) => String(reader.int64());
      }
      return (reader) => reader.int64();
```

I expect BinaryReader's int64 to call into protoInt64.dec, which will already return bigint if supported, string otherwise.

packages/protobuf/src/from-json.ts
```diff
@@ -212,405 +244,777 @@ export function isEnumJson<Desc extends DescEnum>(
...
+/**
+ * Compile the conversion of a validated scalar value to its local
+ * representation: 64-bit integers become bigint, or string with the
+ * longAsString option.
+ */
+function compileScalarToLocal(
+  field: DescField & { scalar: ScalarType },
+): (value: unknown) => unknown {
+  const longAsString =
+    (field as { longAsString?: boolean }).longAsString === true;
```
timostamm: Avoid the cast:

```suggestion
  const longAsString = field.fieldKind !== "map" && field.longAsString;
```

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +275,360 @@ export function readField(
...
+      readChild(child, reader, ctx);
+      (message[localName] as unknown[]).push(toLocal(child));
+    };
   }
+  const scalarType = field.listKind == "enum" ? ScalarType.INT32 : field.scalar;
+  const readScalar = compileScalarReader(
+    scalarType,
+    field.utf8Validation,
+    (field as { longAsString?: boolean }).longAsString ?? false,
+  );
```
timostamm, line 433: Avoid the cast:

```suggestion
  const scalarType = field.listKind == "enum" ? ScalarType.INT32 : field.scalar;
  const longAsString = field.listKind == "scalar" ? field.longAsString : false;
  const readScalar = compileScalarReader(
    scalarType,
    field.utf8Validation,
    longAsString,
  );
```

The better fix would be to let `descFieldMapScalar` declare `longAsString: false`. Out of scope here.

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +275,360 @@ export function readField(
...
+  switch (field.mapKind) {
+    case "scalar": {
+      const scalar = field.scalar;
+      const longAsString =
+        (field as { longAsString?: boolean }).longAsString === true;
+      const readScalar = compileScalarReader(
+        scalar,
+        field.utf8Validation,
+        longAsString,
+      );
```
timostamm:
```suggestion
      const readScalar = compileScalarReader(
        scalar,
        field.utf8Validation,
        false,
      );
```

The `jstype` option cannot be set on map fields. The branch below is also unreachable for any real descriptor, and we only support compiled descriptors.

packages/protobuf/src/from-json.ts
```diff
@@ -212,405 +244,777 @@ export function isEnumJson<Desc extends DescEnum>(
...
+      if (checkValue !== undefined) {
+        const check = checkValue(value);
+        if (check !== true) {
+          throw new FieldError(
+            field,
+            `map entry ${formatVal(key)}: ${reasonSingular(field, value, check)}`,
+          );
+        }
+      }
+      record[mapKeyToLocal(key)] = toLocalValue(value);
```
timostamm:
```suggestion
      // Coerced to string, same as a .toString() call for the possible key types
      record[key as string] = toLocalValue(value);
```

This means compileMapKeyToLocal can be removed as well.

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +275,360 @@ export function readField(
...
+          break;
+      }
+    }
+    if (key === undefined) {
+      key = keyZero;
+    }
+    if (val === undefined) {
+      val = valueDefault();
+    }
+    record[keyToLocal(key) as string] = val;
```
timostamm:
```suggestion
    // Coerced to string, same as a .toString() call for the possible key types
    record[key as string] = val;
```

Same as the from-json.ts counterpart.

timostamm (commented):
Reviewed the rest. Found a bit of dead code and some avoidable casts, see below.

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +274,320 @@ export function readField(
...
   );
-  return message;
+  const keyZero = scalarZeroValue(field.mapKey, false);
+  let readValue: (reader: BinaryReader, ctx: BinaryReadContext) => unknown;
+  let valueDefault: () => unknown;
+  switch (field.mapKind) {
+    case "scalar": {
+      const scalar = field.scalar;
+      const longAsString =
+        (field as { longAsString?: boolean }).longAsString === true;
```
timostamm: This should go, see my suggestion in https://github.com/bufbuild/protobuf-es/pull/1491#discussion_r3721923420

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +274,320 @@ export function readField(
...
+          break;
+      }
+    }
+    if (key === undefined) {
+      key = keyZero;
+    }
+    if (val === undefined) {
+      val = valueDefault();
+    }
+    record[key as string] = val;
```
timostamm, line 527: It is surprising that this works for different key types, and readers will wonder about what's going on, and whether this is a bug.

Please add an explanatory comment as suggested. Same for the from-json.js equivalent.

packages/protobuf/src/from-binary.ts
```diff
@@ -164,183 +274,320 @@ export function readField(
...
 
-function readScalar(
-  reader: BinaryReader,
+/**
+ * Returns a reader for a scalar value, including the conversion to the
+ * local representation for 64-bit integers: the reader returns them as
+ * strings when BigInt is unavailable, and they are run through protoInt64,
+ * mirroring the reflect layer. Which case applies is fixed at load time,
+ * see protoInt64.supported.
+ */
```
timostamm, line 535: The doc is now stale. 64-bit integers no longer run through protoInt64. And with that change, it also no longer mirrors the reflect layer. But that's okay because the reflect layer handles user-provided values, and needs to reject `"garbarge string"`, which is impossible in this code path, where the string comes from BinaryReader's methods.

timostamm (approved):
LGTM!

PR #1498, opened 2026-08-02
Avoid allocations in varint64

ajeetdsouza:
The `varint64` method currently returns a `[number, number]`, which causes an allocation on a fairly hot function. This PR modifies it to return nothing, and use `this.varint64Lo` and `this.varint64Hi` as scratch variables. It also contains a fast path for bools.

Putting this up for general review, but note that this PR does contain a breaking change on a `protected` method.

| Case                         | main (ns/op)  | this PR (ns/op) | Delta  |
|-------------------------------|---------------|------------------|--------|
| BinaryReader/bool             | 3.83          | 1.00             | -73.9% |
| BinaryReader/int64             | 12.32         | 6.25             | -49.2% |
| BinaryReader/sint64              | 10.62         | 6.43             | -39.5% |
| fromBinary/repeated-message       | 1,519,371.85  | 1,452,020.93     | -4.4%  |
| fromBinary/scalar                    | 1,505.08      | 1,444.06         | -4.1%  |
| fromBinary/repeated-scalar              | 3,844.59      | 3,725.59         | -3.1%  |
| fromBinary/general                        | 165,200.04    | 162,520.18       | -1.6%  |
| fromBinary/map-message                       | 1,713,279.28  | 1,713,810.44     | 0.0%   |
| fromBinary/user-normal                          | 2,245.20      | 2,250.30         | 0.2%   |
| fromBinary/user-tiny                               | 730.64        | 746.78           | 2.2%   |

### DIFF
--- a/packages/protobuf/src/wire/binary-encoding.ts
+++ b/packages/protobuf/src/wire/binary-encoding.ts
@@ -477,7 +477,7 @@ export class BinaryReader {
    */
   readonly len: number;
 
-  protected readonly buf: Uint8Array;
+  private readonly buf: Uint8Array;
   private readonly view: DataView;
 
   constructor(
@@ -562,9 +562,9 @@ export class BinaryReader {
     return this.buf.subarray(start, this.pos);
   }
 
-  protected varint64Lo = 0;
-  protected varint64Hi = 0;
-  protected varint64 = varint64read as () => void; // dirty cast for `this`
+  private varint64Lo = 0;
+  private varint64Hi = 0;
+  private varint64 = varint64read as () => void; // dirty cast for `this`
 
   /**
    * Throws error if position in byte array is out of range.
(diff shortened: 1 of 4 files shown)


packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -562,7 +562,9 @@ export class BinaryReader {
     return this.buf.subarray(start, this.pos);
   }
 
-  protected varint64 = varint64read as () => [number, number]; // dirty cast for `this`
+  protected varint64Lo = 0;
+  protected varint64Hi = 0;
+  protected varint64 = varint64read as () => void; // dirty cast for `this`
```
timostamm: Better make all three private. https://github.com/bufbuild/protobuf-es/pull/1108 already dropped `protected buf`.

timostamm (approved):

PR #1499, opened 2026-08-05
Compile and cache create() function

ajeetdsouza:
Follow up to https://github.com/bufbuild/protobuf-es/pull/1491. We spend a lot of time interpreting descriptors when constructing new messages. This PR precomputes a list of init operations for each message descriptor.

| fixture           | create | fromJson | fromBinary |
|---------------------|--------|----------|------------|
| repeated-message       | -55.4% | -9.5%    | -19.3%     |
| map-message               | -53.9% | -9.6%    | -16.0%     |
| scalar                       | -52.2% | -8.0%    | -25.1%     |
| user-normal                     | -39.7% | -5.3%    | -7.1%      |
| repeated-scalar                    | -34.3% | 0.0%     | -3.2%      |
| map-scalar                            | -34.1% | -0.7%    | -4.2%      |
| user-tiny                                | -29.5% | -12.3%   | -13.2%     |
| general                                     | -22.4% | -2.3%    | -1.8%      |

### DIFF
--- a/packages/protobuf/src/create.ts
+++ b/packages/protobuf/src/create.ts
@@ -21,9 +21,7 @@ import {
 } from "./descriptors.js";
 import type { Message, MessageInitShape, MessageShape } from "./types.js";
 import { scalarZeroValue } from "./reflect/scalar.js";
-import type { FieldError } from "./reflect/error.js";
-import { isObject, type OneofADT } from "./reflect/guard.js";
-import { unsafeGet, unsafeOneofCase, unsafeSet } from "./reflect/unsafe.js";
+import { isObject } from "./reflect/guard.js";
 import { isWrapperDesc } from "./wkt/wrappers.js";
 
 // bootstrap-inject google.protobuf.Edition.EDITION_PROTO3: const $name = $number;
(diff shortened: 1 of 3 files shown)


timostamm:
Is this ready for review? A PR description, branch update and CI fix would be nice.

packages/protobuf/src/create.ts
```diff
@@ -46,216 +44,365 @@ export function create<Desc extends DescMessage>(
...
-function initScalar(
-  field: DescField & { fieldKind: "scalar" },
-  value: unknown,
-): unknown {
-  if (field.scalar == ScalarType.BYTES) {
-    return toU8Arr(value);
+  // For smaller messages, a megamorphic clone has a high fixed cost, so we're
+  // better off setting fields to their zero values ourselves.
+  let template: Record<string, unknown> | undefined;
+  const isSmallMessage = desc.members.length + 1 <= smallMessageMaxProperties; // + 1 for $typeName
```
timostamm: The clone applies to a subset of `desc.members`, right? If you've already looked into this and came to the conclusion that it's still a good threshold, that's fine, but it's worth a comment.

packages/protobuf/src/create.ts
```diff
@@ -46,216 +44,365 @@ export function create<Desc extends DescMessage>(
...
+  let convertItem: Converter | undefined;
+  if (field.listKind == "message") {
+    convertItem = compileConvertMessage(field);
+  } else if (field.scalar == ScalarType.BYTES) {
+    convertItem = toU8Arr;
+  }
+  if (convertItem === undefined) {
+    return (message, init) => {
+      const value = init[name];
+      message[name] = value ?? [];
```
timostamm: There's a very subtle behavior change here. For list and map of google.protobuf.Struct, we used to create a copy of the container, now we're storing the container by reference.

I don't think behavior was consistent across field types before, but this adds another split. I think it would be better to keep the existing behavior.

packages/protobuf/src/create.ts
```diff
@@ -46,216 +44,365 @@ export function create<Desc extends DescMessage>(
...
+          Object.create(prototype) as Record<string, unknown>,
+          template,
+        );
+      } else {
+        message = { ...template };
+      }
+    }
+    if (init === undefined) {
+      for (let i = 0; i < lists.length; i++) {
+        message[lists[i]] = [];
```
timostamm: This means that property order differs between `create(S)` and `create(S, {})`, which is not ideal. Usually, field order in Protobuf is meaningful, we're careful to preserve it in generated code, and it would be great if we could also preserve it in memory.

timostamm (commented):
Took me a minute to go through this. As far as I see, this only really makes sense along with https://github.com/bufbuild/protobuf-es/pull/1502, right?

timostamm (approved):

PR #1502, opened 2026-08-06
Upgrade TypeScript target to ES2022

ajeetdsouza:
In line with our compat policy, this PR upgrades our target to ES2022. Along with reducing bundle size, this also speeds up some of our benchmarks (see https://github.com/bufbuild/protobuf-es/pull/1499).


timostamm:
I'm more comfortable bumping the target in a major version, along with dropping old TS versions from the TS-compat tests.

PR #1503, opened 2026-08-06
Track seen values with bitset

ajeetdsouza:
Rather than tracking seen values with a `Set`, we now use a bitset `number | number[]`. This speeds up checks, and also avoids allocation for most messages:

| case                            | main ns/op | ajeet/bitset ns/op | delta  |
|-----------------------------------|------------|--------------------|--------|
| fromJson/user-normal                | 648        | 532                | -18.0% |
| fromJson/scalar                        | 1,624      | 1,369              | -15.7% |
| fromJson/repeated-message                 | 1,700,680  | 1,453,488          | -14.5% |
| fromJson/map-message                         | 1,828,154  | 1,587,302          | -13.2% |
| fromJson/user-tiny                              | 278        | 242                | -13.1% |
| fromJson/repeated-scalar                           | 2,659      | 2,574              | -3.2%  |
| fromJson/general                                      | 165,755    | 160,643            | -3.1%  |
| fromJson/map-scalar                                      | 2,834      | 2,810              | -0.9%  |

A similar change could probably be explored for `seenOneofs`, but for now, I've made a quickfix to avoid the allocation in messages with zero oneof fields.

### DIFF
--- a/packages/protobuf/src/from-json.ts
+++ b/packages/protobuf/src/from-json.ts
@@ -274,6 +274,8 @@ interface CompiledFieldEntry {
   // A oneof member that is a scalar field skips JSON null, see conformance
   // test Required.Proto3.JsonInput.OneofFieldNull{First,Second}.
   oneofScalarNullSkip: boolean;
+  // The field's index in the descriptor (different from the field number).
+  fieldIdx: number;
 }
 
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
Closing this - the next release of Protobuf does not track seen values in JSON, so we can remove this check entirely.

PR #1504, opened 2026-08-07
Fix byte size estimate in base64 fallback

ajeetdsouza:
Correctness fix - for certain poorly formed inputs, the handwritten impl of base64 may underallocate the buffer, missing a byte in the end.

### DIFF
--- a/packages/protobuf-test/src/wire/base64-encoding.test.ts
+++ b/packages/protobuf-test/src/wire/base64-encoding.test.ts
@@ -211,8 +211,9 @@ void suite("base64Decode()", () => {
     "c3VyZQ==\tc3VyZQ==",
     "c3VyZQ==\rc3VyZQ==",
     "c3VyZQ== c3VyZQ==",
+    "c3VyZXN1cmU==",
   ]) {
-    void test(`allows inner padding in ${b64}`, () => {
+    void test(`allows padding in ${b64}`, () => {
       assert.deepStrictEqual(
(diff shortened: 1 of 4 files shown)


packages/protobuf-test/src/wire/base64-encoding.test.ts
```diff
@@ -211,8 +211,9 @@ void suite("base64Decode()", () => {
     "c3VyZQ==\tc3VyZQ==",
     "c3VyZQ==\rc3VyZQ==",
     "c3VyZQ== c3VyZQ==",
+    "c3VyZXN1cmU==",
   ]) {
-    void test(`allows inner padding in ${b64}`, () => {
+    void test(`allows padding in ${b64}`, () => {
```
timostamm: The test names correspond to the JSDoc bullets. The new test doesn't fit here. Revert and add:

```ts
  for (const [b64, expected] of [
    ["c3Vy=", "sur"],
    ["c3Vy==", "sur"],
    ["c3VyZS4==", "sure."],
    ["c3VyZXN1cmU==", "suresure"],
  ]) {
    void test(`allows excess padding in ${b64}`, () => {
      assert.deepStrictEqual(
        base64Decode(b64),
        new TextEncoder().encode(expected),
      );
    });
  }
```

timostamm (commented):
Good catch, just one thing to fix up in the tests:

timostamm (approved):

PR #1506, opened 2026-08-08
Reorder hasOwnProperty for explicit-presence fields

ajeetdsouza:
This doesn't move any numbers on our current benchmarks. But let's name the two checks:

- H = `Object.prototype.hasOwnProperty.call(message, localName)`
- M = `message[localName] != undefined` 

Then:

- For absent message fields, any syntax: `M` gets replaced with `H` (no effect)
- For absent scalars on proto3: `M` gets replaced with `H` (no effect)
- For absent scalars on proto2 / editions: `M+H` gets replaced with `H` (big win)
- For own properties that are explicitly marked as `undefined`: `M` becomes `M+H` (big loss)

Not sure if this is a clean win, but I think absent scalars are more common that explicit `undefined` properties.

### DIFF
--- a/packages/bundle-size/README.md
+++ b/packages/bundle-size/README.md
@@ -16,11 +16,11 @@ usually do. We repeat this for an increasing number of files.
 
 | code generator      | files | bundle size |  minified | compressed |
 | ------------------- | ----: | ----------: | --------: | ---------: |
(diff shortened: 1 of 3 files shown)


timostamm:
What if we instead use `H+M` for fields that use the prototype chain, and `M` for fields that don't? That would be optimal for all cases.

PR #1507, opened 2026-08-10
Define descriptor toString instead of assigning it

ltrouton:
We use Protobuf-ES as part of the license key system in the TinyMCE rich text editor. A customer embedded the editor in an environment that freezes the built-in JavaScript prototypes. In this frozen environment, Protobuf-ES throws an exception when attempting to create a file descriptor. The exception occurs at module load time before any of our code runs, so it cannot be caught. 

In Chrome, the exception `TypeError: Cannot assign to read only property 'toString' of object '#<Object>` is thrown. Firefox reports `"toString" is read-only`. Safari reports `Attempted to assign to readonly property`.

I found two locations in registry.ts that assign `toString`. The other seven descriptors define it in the object literal, so are safe. A frozen `Object.prototype.toString` is non-writable, so it throws in strict mode. `Object.defineProperty` creates the own property directly and is unaffected. The descriptor matches what assignment produces: `writable: true, enumerable: true, configurable: true`. `Object.keys` and spread behave as before.

The added tests cover string output and enumerability for non-frozen and frozen environments. The frozen-prototype test fails on main without the `defineProperty` fix.

Bundle size grows 236 b raw and 150 b minified. Compressed is unchanged within noise. The regenerated report is included.

Let me know if there is anything else I need to do for this PR.

### DIFF
--- a/packages/protobuf-test/src/registry.test.ts
+++ b/packages/protobuf-test/src/registry.test.ts
@@ -2182,26 +2182,6 @@ void suite("descriptor toString()", () => {
     extend Msg { optional int32 ext = 100; }
   `;
 
-  // Object.freeze(Object.prototype) is irreversible and would leak into every
-  // later test in this process.
-  function withNonWritableToString<T>(fn: () => T): T {
-    const original = Object.getOwnPropertyDescriptor(
-      Object.prototype,
-      "toString",
-    );
-    assert.ok(original);
-    Object.defineProperty(Object.prototype, "toString", {
-      ...original,
-      writable: false,
-      configurable: true,
-    });
-    try {
-      return fn();
-    } finally {
-      Object.defineProperty(Object.prototype, "toString", original);
-    }
-  }
-
   function describeAll(reg: FileRegistry): string[] {
     const msg = reg.getMessage("Msg");
     assert.ok(msg);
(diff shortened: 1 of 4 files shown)


packages/protobuf/src/registry.ts
```diff
@@ -860,7 +860,14 @@ function newField(
     field.oneof = undefined;
     field.typeName = typeName;
     field.jsonName = `[${typeName}]`; // option json_name is not allowed on extension fields
-    field.toString = () => `extension ${typeName}`;
+    // A plain assignment throws where built-in prototypes are frozen. The
+    // attributes below match what an assignment produces.
+    Object.defineProperty(field, "toString", {
```
timostamm: Instead of calling defineProperty twice, declare `let toStr: () => string;` upfront, assign in the two branches, and call defineProperty once.

packages/protobuf-test/src/registry.test.ts
```diff
@@ -2171,3 +2171,75 @@ void suite("DescMethod", () => {
...
+    message Msg {
+      optional int32 fld = 1;
+      extensions 100 to 200;
+    }
+    extend Msg { optional int32 ext = 100; }
+  `;
+
+  // Object.freeze(Object.prototype) is irreversible and would leak into every
+  // later test in this process.
+  function withNonWritableToString<T>(fn: () => T): T {
```
timostamm: Instead of this wrapper, use [after](https://nodejs.org/api/test.html#afterfn-options) to restore. See [here](https://github.com/bufbuild/protobuf-es/blob/94662118aefbdb53efcef0aafc4c4757945c3b0c/packages/protoplugin-example/test/generated.ts#L43) for an example.

timostamm (commented):
Freezing the built-in prototypes is a good defense against prototype pollution. Thanks for the contribution. Can you merge in the latest changes and address the comments below? Happy to merge the PR after that.

packages/protobuf/src/registry.ts
ltrouton: Thanks for the suggestion. I have made the change so `defineProperty` only has to be called once.

packages/protobuf-test/src/registry.test.ts
ltrouton: That's a good way to do it. Thanks for the suggestion. Applied

ltrouton:
Thanks for the review @timostamm. I have addressed your suggestions and rebased with the `main` branch, and regenerated the bundle-size numbers.

timostamm (approved):
Thanks for the PR!

PR #1509, opened 2026-08-11
Add ASCII fast paths

ajeetdsouza:
Calling TextEncoder functions incurs a fixed cost, below which it is much cheaper to execute code in JS itself. This PR bets on most small strings being pure ASCII:

| case                        | delta  |
|-------------------------------|--------|
| BinaryWriter/string-ascii       | -69.6% |
| BinaryReader/string-ascii         | -43.9% |
| toBinary/user-normal                 | -42.0% |
| fromBinary/user-normal                  | -29.2% |
| toBinary/general                           | -17.3% |
| fromBinary/general                            | -5.4%  |
| BinaryWriter/string-utf8                         | +6.5%  |
| BinaryReader/string-utf8                            | +6.1%  |

### DIFF
--- a/packages/protobuf/src/wire/binary-encoding.ts
+++ b/packages/protobuf/src/wire/binary-encoding.ts
@@ -299,9 +299,28 @@ export class BinaryWriter {
     if (typeof value !== "string") {
       value = String(value);
     }
+    const len = value.length;
+
+    // Fast path for ASCII.
+    if (len <= ASCII_MAX_LENGTH) {
+      this.ensureCapacity(len + 1);
+      const ascii = this.buffer;
+      let pos = this.pos;
+      ascii[pos++] = len;
+      let i = 0;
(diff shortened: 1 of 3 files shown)


packages/protobuf/src/wire/binary-encoding.ts
```diff
@@ -755,7 +780,23 @@ export class BinaryReader {
    * `strict` is true, throw on invalid UTF-8 instead of substituting U+FFFD.
    */
   string(strict?: boolean): string {
-    return this.decodeUtf8(this.bytes(), strict);
+    const bytes = this.bytes();
```
timostamm, line 783: It would be possible to avoid the allocation and run the fast path on this.buf.
I'm not sure it's worth the complication. Maybe something worth looking into in a follow-up.

timostamm (approved):

PR #1510, opened 2026-08-11
Render background for bundle-size

ajeetdsouza:
The bundle-size chart is not readable on GitHub with the dark theme enabled. This PR adds a white background for visibility.


pkwarren (approved):
Thanks!

timostamm (approved):
I was hoping for a second this was rendering the bundle size report in the background, which would be really great :)

Issue #1203, opened 2025-08-16
Document `rewrite_imports` option in `MANUAL.md` + Explain how to do gradual migration


tenzinhl:
We ended up doing a very large migration internally within our company that was all at once when migrating from protobuf-es v1 to protobuf-es v2 (to get access to connect-es v2 features).

However, it would've been very helpful for us if we had been aware of the `rewrite_imports` plugin option to instead do the migration in a piece-wise fashion as demonstrated by @timostamm 's lovely example repo [HERE](https://github.com/timostamm/example-protobuf-es-v1-to-v2). (Would've also been huge for relocating sources within our bazel output tree)

I'm not sure I fully understand the plugin option myself yet since we didn't use it, but a mention in `MANUAL.md` would be awesome. It would also be lovely if the "How to migrate" section of the `MANUAL.md` could highlight that it's possible to do the migration in a piecewise fashion with a link to https://github.com/timostamm/example-protobuf-es-v1-to-v2.

## Related Info

Issue comment where @timostamm discusses the `rewrite_imports` plugin.
- https://github.com/bufbuild/protobuf-es/issues/947#issuecomment-2257889255

Thanks a bunch for the project team.

timostamm:
The reason we've been hesitant to document the option is that it does many different things at once, and that makes it a bit difficult to use. 

In  [v2.15.0](https://github.com/bufbuild/protobuf-es/releases/tag/v2.15.0) we've added the new [plugin option map_imports](https://protobufes.com/reference/plugin-options/#map_imports) for a subset of the functionality of `rewrite_imports`. It rewrites imports for generated Protobuf files, so that generated code for a shared Protobuf can be shared as well, for example in separate packages in a monorepo or published on npmjs.com.

It doesn't replace the migration use case. For that, `rewrite_imports` is still the right choice. But my guess is that anybody who wants to migrate from protobuf-es v1 to v2 has already migrated, so I'll close the issue. If there's enough demand for such a feature, we'll consider a separate plugin option that's scoped to rewriting imports for the runtime library.

PR #1515, opened 2026-08-24
Apply import_extension only to relative import paths

timostamm:
The plugin option `import_extension` controls the file extension of generated imports. But it must only control the file extensions for generated code, not for code that is imported from published npm packages.

If a plugin generates an import from an npm package `import { Foo } from "foo/foo.js"`, the plugin option `import_extension=none` or `import_extension=ts` must not modify that import.

### DIFF
--- a/packages/protoplugin-test/src/import_extension.test.ts
+++ b/packages/protoplugin-test/src/import_extension.test.ts
@@ -94,6 +94,30 @@ void suite("import_extension", () => {
       "json",
     ]);
   });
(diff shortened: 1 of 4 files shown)


timostamm:
This brings us one step closer to https://github.com/bufbuild/protobuf-es/issues/1203, which will help with https://github.com/bufbuild/protovalidate-es/issues/151.

ajeetdsouza (approved):

PR #1516, opened 2026-08-24
Backfill `encodeUtf8Into` for backwards compat with `protobuf-es v2.13`

alexandercampbell-wk:
This change is intended to fix a bug. Repro steps:

1. Launch a JS isolate (e.g., browser tab)
2. Load protobuf-es v2.13
3. Load protobuf-es v2.14
4. Attempt to use the v2.14 `BinaryWriter.string()`
5. Observe that `TypeError` was thrown

Explanation: protobuf-es v2.13 installs some functions onto `globalThis[Symbol.for("@bufbuild/protobuf/text-encoding")]`, but it does not install `encodeUtf8Into`. protobuf-es v2.14 assumes that if the singleton exists, it has `encodeUtf8Into`. This causes v2.14 to fail writes by throwing `TypeError`.

To fix: when grabbing the global singleton, check if the method is missing. If it's missing, backfill it.

### DIFF
--- a/docs/src/content/docs/guides/serialization.md
+++ b/docs/src/content/docs/guides/serialization.md
@@ -172,3 +172,5 @@ This format is compatible with the delimited message support in the C++, Java, G
 Protobuf-ES uses the WHATWG [Text Encoding API](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API) to convert UTF-8 to and from binary.
 
 If your environment does not provide the API, call `configureTextEncoding()` from `@bufbuild/protobuf/wire` early during initialization and supply your own implementation.
+
+Multiple copies of this library in the same JavaScript isolate share one codec via a singleton on `globalThis`. If an older copy (v2.13 and earlier) initialized that slot without `encodeUtf8Into`, `getTextEncoding()` backfills it.
--- a/packages/protobuf-test/src/wire/text-encoding.test.ts
+++ b/packages/protobuf-test/src/wire/text-encoding.test.ts
@@ -17,7 +17,11 @@ import * as assert from "node:assert";
 import {
   getTextEncoding,
   configureTextEncoding,
+  BinaryWriter,
+  BinaryReader,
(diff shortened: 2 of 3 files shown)


e2digital-eurostar:
Ran into this bug, keen to see this merged.

timostamm:
Thanks for the report. Can you share some context on the situation where you're running into the issue? I assume you're not using `configureTextEncoding()`?

alexandercampbell-wk:
@timostamm sorry, I'm not sure what you're asking for. Is there a part of the PR description that is unclear? I could prepare a repro demo if needed.

I can confirm we are not calling `configureTextEncoding()` directly.

alexandercampbell-wk:
Ah, it occurs to me you may be wondering why anyone would load two different versions of `protobuf-es` into the same browser tab.

We are using a microfrontend architecture, where several independently compiled and versioned JS bundles are shipped and loaded together. One bundle imported 2.13, the other imported 2.14. This caused the exception at runtime.

e2digital-eurostar:
I ran into the problem when using the https://github.com/elastic/elastic-otel-node image for auto-instrumenting otel. It pulls in an older version of protobuf-es.

timostamm:
Thanks for the context! The diagnosis is correct and the fix addresses the type error.

But there's also an older instance of the same bug: v2.12.0 added the `strict` argument to `decodeUtf8` to enforce UTF-8 validation on binary input. If you've loaded v2.11 or earlier before, the new argument is never honored. The difference to the bug you're addressing is that it's silent. This leads me to think that we ought to rip off the bandaid, and stop sharing the text encoding implementation in global context altogether: https://github.com/bufbuild/protobuf-es/pull/1517.

timostamm:
The fix is released in [v2.14.1](https://github.com/bufbuild/protobuf-es/releases/tag/v2.14.1). Let us know if you still encounter any issues.

alexandercampbell-wk:
Thank you!

PR #1517, opened 2026-08-31
Stop sharing the text encoding implementation via globalThis

timostamm:
V2 of `@bufbuild/protobuf` introduced a facade for the Text Encoding API. The purpose was to make it easier to bring your own implementation in environments that don't have it.

The implementation is kept in a singleton on `globalThis`, shared by every copy of the library. Unfortunately, that made the shape of the implementation an undeclared contract between versions, and we have accidentally broken it twice:

- v2.12 added the `strict` argument to `decodeUtf8`
- v2.14 added `encodeUtf8Into`

When more than one copy of `@bufbuild/protobuf` is loaded, whichever copy initializes first wins the slot on the singleton, binding the implementation for every copy that comes after. A newer copy then runs on an implementation built against an older interface, ignoring `strict` or throwing a TypeError on a missing `encodeUtf8Into`.

To fix this issue, we're changing the facade to stop keeping state on `globalThis`. Every copy of `@bufbuild/protobuf` gets its own implementation now, which guarantees that it always works as intended.

We're also deprecating the function `configureTextEncoding`. While it's still possible to bring your own Text Encoding implementation via the function, it will only install your implementation for the copy it is imported from. To provide the Text Encoding API globally, we recommend installing `TextEncoder` (with the methods `encode` and optionally `encodeInto`) and `TextDecoder` (with the `fatal: true` constructor argument and the `decode` method) on `globalThis` instead.

Thanks to @alexandercampbell-wk and @e2digital-eurostar for the report in #1516.

### DIFF
--- a/packages/protobuf/src/wire/text-encoding.ts
+++ b/packages/protobuf/src/wire/text-encoding.ts
@@ -71,7 +71,7 @@ export function configureTextEncoding(textEncoding: TextEncodingConfig): void {
 export function getTextEncoding(): TextEncoding {
   if (!te) {
     const globals = globalThis as unknown as GlobalWithTextEncoderDecoder;
-    if (!globals.TextEncoder || !globals.TextDecoder) {
+    if (!globals.TextEncoder || !globals.TextEncoder) {
       throw new Error(
         "Encoding API missing - install TextEncoder and TextDecoder on globalThis",
       );
--- a/packages/protobuf/src/wire/text-encoding.ts
+++ b/packages/protobuf/src/wire/text-encoding.ts
@@ -71,9 +71,9 @@ export function configureTextEncoding(textEncoding: TextEncodingConfig): void {
 export function getTextEncoding(): TextEncoding {
   if (!te) {
     const globals = globalThis as unknown as GlobalWithTextEncoderDecoder;
-    if (!globals.TextEncoder || !globals.TextEncoder) {
+    if (!globals.TextEncoder || !globals.TextDecoder) {
       throw new Error(
-        "Encoding API missing - install TextEncoder and TextDecoder on globalThis",
+        "encoding API missing: install TextEncoder and TextDecoder on globalThis",
       );
     }
     const textEncoder = new globals.TextEncoder();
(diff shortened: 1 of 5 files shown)


packages/protobuf/src/wire/text-encoding.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const symbol = Symbol.for("@bufbuild/protobuf/text-encoding");
+let te: TextEncoding | undefined;
```
ajeetdsouza, line 15: Since this is a global, I'd suggest a more descriptive name (`textEncoding`).

packages/protobuf/src/wire/text-encoding.ts
```diff
@@ -61,9 +69,13 @@ export function configureTextEncoding(textEncoding: TextEncodingConfig): void {
...
 
 export function getTextEncoding(): TextEncoding {
-  const globals = globalThis as unknown as GlobalWithTextEncoding &
-    GlobalWithTextEncoderDecoder;
-  if (!globals[symbol]) {
+  if (!te) {
+    const globals = globalThis as unknown as GlobalWithTextEncoderDecoder;
+    if (!globals.TextEncoder || !globals.TextDecoder) {
+      throw new Error(
+        "Encoding API missing - install TextEncoder and TextDecoder on globalThis",
```
ajeetdsouza: Formatting nit:


```suggestion
        "encoding API missing: install TextEncoder and TextDecoder on globalThis",
```

ajeetdsouza (approved):

packages/protobuf/src/wire/text-encoding.ts
timostamm: Agree that a more descriptive name wouldn't hurt, but it conflicts with the argument to `configureTextEncoding`.

packages/protobuf/src/wire/text-encoding.ts
```diff
@@ -71,7 +71,7 @@ export function configureTextEncoding(textEncoding: TextEncodingConfig): void {
 export function getTextEncoding(): TextEncoding {
   if (!te) {
     const globals = globalThis as unknown as GlobalWithTextEncoderDecoder;
-    if (!globals.TextEncoder || !globals.TextDecoder) {
+    if (!globals.TextEncoder || !globals.TextEncoder) {
```
ajeetdsouza: Seems like an erroneous commit - this condition has changed and the error message is still the same.

PR #1528, opened 2026-09-08
Add plugin option `map_imports` to point imports at packages

anuraaga:
This adds `map_imports`, an option to point imported `.proto` files to a different location then the default relative path from the output directory. This notably allows importing from external libraries published to NPM. 

For example, if your proto imports `buf/validate/validate.proto` to add validation annotations, you can use the following configuration that matches on the `buf/validate/` prefix to have generated code import from the `@bufbuild/protovalidate` package, avoiding needing `require_imports` to compile a duplicate copy of the validate.proto gencode.

```yaml
version: v2
plugins:
  - local: protoc-gen-es
    out: src/gen
    opt:
      - target=ts
      - map_imports=buf/validate/:@bufbuild/protovalidate/gen
```

### DIFF
--- a/docs/src/content/docs/reference/plugin-options.md
+++ b/docs/src/content/docs/reference/plugin-options.md
@@ -67,7 +67,7 @@ The option takes the form `map_imports=<pattern>:<target>` and can be repeated. 
 - `**/` matches zero or more directories.
 - A trailing `/` matches every file in the directory and its subdirectories.
 
-`import_extension` only applies to imports of local files. Mapped imports always end in `_pb.js`, the file name of the generated code the package provides. Well-known types are always imported from `@bufbuild/protobuf/wkt`.
+The target can also be a path relative to the output directory, for example to import from the output of another plugin. Relative targets are local files, so they receive the `import_extension` like any other local import. Package targets always end in `_pb.js`. Well-known types are always imported from `@bufbuild/protobuf/wkt`.
 
 ## `js_import_style`
 
--- a/docs/src/content/docs/reference/plugin-options.md
+++ b/docs/src/content/docs/reference/plugin-options.md
@@ -67,7 +67,7 @@ The option takes the form `map_imports=<pattern>:<target>` and can be repeated. 
 - `**/` matches zero or more directories.
 - A trailing `/` matches every file in the directory and its subdirectories.
 
-The target can also be a path relative to the output directory, for example to import from the output of another plugin. Relative targets are local files, so they receive the `import_extension` like any other local import. Package targets always end in `_pb.js`. Well-known types are always imported from `@bufbuild/protobuf/wkt`.
+Package imports always end in `_pb.js`. Well-known types are always imported from `@bufbuild/protobuf/wkt`.
 
 ## `js_import_style`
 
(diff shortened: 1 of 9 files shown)


packages/protoplugin/src/names.ts
```diff
@@ -46,6 +48,10 @@ export function generateFilePath(
   ) {
     return wktFrom;
   }
+  const target = mapImportTarget(file.proto.name, mapImports);
+  if (target !== undefined) {
+    return target + "/" + file.name + "_pb.js";
+  }
```
timostamm, line 54: Nice!

docs/src/content/docs/reference/plugin-options.md
```diff
@@ -42,6 +42,33 @@ Use this option when your environment requires one:
...
+With this option, `buf/validate/validate.proto` is imported from `@bufbuild/protovalidate/gen/buf/validate/validate_pb.js`.
+
+The option takes the form `map_imports=<pattern>:<target>` and can be repeated. The pattern is matched against the path of the Protobuf file, and the first matching pattern wins. The target is prepended to the path of the generated file. Patterns support a subset of glob:
+
+- `*` matches zero or more characters except `/`.
+- `**` matches zero or more characters, including `/`.
+- `**/` matches zero or more directories.
+- A trailing `/` matches every file in the directory and its subdirectories.
+
+The target can also be a path relative to the output directory, for example to import from the output of another plugin. Relative targets are local files, so they receive the `import_extension` like any other local import. Package targets always end in `_pb.js`. Well-known types are always imported from `@bufbuild/protobuf/wkt`.
```
timostamm: I think we need to give an example to make this understandable without users having to read it twice.

> The target can also be a path relative to the output directory, for example to import from the output of another plugin:
> 
> ```yaml
> version: v2
> plugins:
>   # Base types
>   - local: protoc-gen-es
>     out: src/gen/es
>     opt:
>       - target=ts
>   # Twirp clients, importing the base types from the output above
>   - local: protoc-gen-twirp
>     out: src/gen/twirp
>     opt:
>       - target=ts
>       - map_imports=**:../es
> ```

But to be honest, I am not sure that this feature is worth keeping. I can't think of a good use case - do you have one in mind?

docs/src/content/docs/reference/plugin-options.md
```diff
@@ -42,6 +42,33 @@ Use this option when your environment requires one:
 
 Using ECMAScript modules in Node.js typically requires `.js`. Deno typically requires `.ts`.
 
+## `map_imports`
```
timostamm, line 45: Can you add the new option to the enumeration in docs/src/content/docs/reference/index.md?

timostamm (commented):
This looks clean and simple to me. Two comments below:

docs/src/content/docs/reference/plugin-options.md
anuraaga: Yeah I don't think there's a great use case for it - there isn't any special handling to support relative imports, when testing since I found they happen to work as normal, so went ahead and just documented it. I went with simplifying the docs here by not mentioning it to focus on the main use case.

timostamm (approved):
LGTM!

Can you update the PR description to show an example for the option? So that users clicking through from the release notes see what this is right away.

PR #1536, opened 2026-09-17
Round JSON float32 values to float32 precision before range checks

YuEfSaEDU:
Decoding a JSON float field fails with "expected number (float32): out of range" when the value is a round-trippable representation of the largest finite float32, such as the `3.4028235e+38` that Go and Python emit for `math.MaxFloat32`. The decimal literal compares greater than the float64 representation of FLOAT32_MAX, so the range check rejected it even though the value converts to exactly FLOAT32_MAX at float32 precision.

This rounds values to the nearest float32 with `Math.fround()` before the range comparison, both in the scalar check that backs `fromJson()` and the reflect API, and in the binary writer's `assertFloat32()`, matching the behavior of the Go, C++, and Python implementations. Values that genuinely exceed the float32 range, such as `3.5e38`, are still rejected with the existing "out of range" message, and NaN/infinity handling is unchanged.

Reproduction and discussion: connectrpc/connect-es#1716


timostamm (approved):
Thanks for the PR!

I pushed up an optimization that makes it a bit faster and reduces bundle size compared to main.

PR #1538, opened 2026-09-18
Export well-known Protobuf files with their full name

timostamm:
`@bufbuild/protobuf/wkt` exports all well-known types as a single barrel. That's convenient for commonly used types such as `google.protobuf.Timestamp`, but it becomes a problem if the well-known types ever start using the same symbol name twice. That's the case in [protocolbuffers/protobuf v36.0](https://github.com/protocolbuffers/protobuf/releases/tag/v36.0).

Going forward, all well-known files will be exported with their full name, for example `@bufbuild/protobuf/wkt/google/protobuf/c_sharp_features_pb.js` to avoid conflicts.

Commonly used types are still exported from the `@bufbuild/protobuf/wkt` barrel, and generated code still imports from there.

Note that generated code will change for some imports (`cpp_features.proto`, `go_features.proto`, `java_features.proto`, `source_context.proto`, `type.proto`, `api.proto`), but they are still exported from the barrel for backwards compatibiliy.


anuraaga (approved):
Thanks - was considering a very similar change for protobuf-py for similar reasons

